PR #10, opened 2015-01-10
Add safe alternative to query_row

marcusklaas:
`query_row_safe` is functionally identical to `query_row`, but returns an `SqliteError` when it fails instead of panicking. This pull request also contains the changes contained in my other pull request, so that it will compile properly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusqlite"
-version = "0.0.6"
+version = "0.0.7"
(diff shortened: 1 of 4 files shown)


jgallagher:
Thanks! I made a few small changes to the 1.0.0 updates (e.g., switching the return type from `uint` to `c_int` instead of `usize` to match sqlite), but otherwise this is all rolled into #11. Please let me know if anything there is amiss.

PR #9, opened 2015-01-10
update for rust 1.0 alpha

marcusklaas:
several small fixes

PR #11, opened 2015-01-11
Rust 1.0.0 alpha changes, add `query_row_safe`

jgallagher:
This should encompass both #9 and #10

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,5 @@
+rusqlite contributors (sorted alphabetically)
+=============================================
+
+* [John Gallagher](https://github.com/jgallagher)
+* [Marcus Klaas de Vries](https://github.com/marcusklaas)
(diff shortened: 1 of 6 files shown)


marcusklaas:
Looks good, but I'd strongly advise against returning `c_int` instead of a native rust type.

jgallagher:
Honestly I was more uneasy about they way the code was before, casting the `c_int` that sqlite returns into a `uint` (what if sqlite returns -1 for some reason?). `c_int` is just a type alias for the appropriately sized rust native integer ([usually i32](http://doc.rust-lang.org/libc/types/os/arch/c95/type.c_int.html)), so callers can treat it as an `i32` without any problems (unless they go to build on a system where `c_int` isn't 32 bits, at which point they'll get an appropriate error since their rust code and sqlite aren't in agreement).

marcusklaas:
That's a solid argument. But the C code won't ever return a value that cannot be represented by 32 bits, right? The reason I'm asking is because `c_int` exposes internals of your library, which may not be necessary.

PR #12, opened 2015-01-13
Query and collect convenience method

marcusklaas:
Provides a safe and easy way to collect the results from a query

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusqlite"
-version = "0.0.6"
+version = "0.0.7"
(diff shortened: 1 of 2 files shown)


jgallagher:
This would be excellent as an example, but I would rather not add it to the library. I think it might be a little advanced for the README (maybe not, curious about what you think about that). What would you think about an `examples/` directory?

marcusklaas:
I understand. It doesn't add any new functionality. The reason the method appealed to me is that it's not possible to access stale rows, as opposed to working with an iterator.

It's possible to add it as an example, but by no means necessary. The examples provided in the documentation/ readme are already very insightful in my mind.

I'll close this PR for now. Thanks for your considerate as always :+1:

PR #14, opened 2015-01-19
import crate for bitflags macro

marcusklaas:
makes it compile on rust nightly again


jgallagher:
Thanks! The current rust nightly is out of date, so I can't publish to crates.io yet; I'll do that ASAP.

PR #15, opened 2015-01-24
Rustup

marcusklaas:
`unwrap_err` had to be replaced by explicit matches because the unit type `()` does not implement `Display`, which is required in order to use `unwrap_err`. This is not very natural (requiring `Debug` makes more sense) and likely to be reverted soon, or so I've been told. Until then, these changes will make rusqlite compile on nightly rustc again.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusqlite"
-version = "0.0.7"
+version = "0.0.8"
 authors = ["John Gallagher <jgallagher@bignerdranch.com>"]
 description = "Ergonomic wrapper for SQLite"
 homepage = "https://github.com/jgallagher/rusqlite"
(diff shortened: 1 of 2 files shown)


gchp:
+1 on merging this!

jgallagher:
Sorry for the delay. Are you guys still having problems on the latest nightly? I just updated to `rustc 1.0.0-nightly (458a6a2f6 2015-01-25 21:20:37 +0000)` and it appears this fix is no longer required, presumably because of https://github.com/rust-lang/rust/pull/21558. If you guys confirm, I'll close this and fix the deprecations.

gchp:
@jgallagher I'm not able to update to the most recent nightly right now due to some problem verifying the hash. I'll comment again once it installs.

jgallagher:
Looks like everything is okay on the latest master - closing this PR.

PR #16, opened 2015-02-04
Rustup 20150203

pfernie:
Simple changes for recent rustc:
Show -> Debug
String -> Display

some #[feature] attributes for unstable APIs


jgallagher:
Thanks :+1:

PR #17, opened 2015-02-04
add SqliteStatement::query_iter()

pfernie:
In situations where a query is dynamically built with a variable
number of binding parameters, query_iter() allows passing an Iterator
over items that impl ToSql

This is from a rust neophyte, so this may be doable via the existing query() fn? However, this impl was the only way I could find to have a dynamically sized set of params passed. Furthermore, I may have overcomplicated the fn signature re: lifetimes?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -533,7 +533,7 @@ impl<'conn> SqliteStatement<'conn> {
     /// ```rust,no_run
     /// # use rusqlite::{SqliteConnection, SqliteResult};
     /// fn get_names_for_ids(conn: &rusqlite::SqliteConnection, ids: Vec<i64>) -> Result<Vec<String>, rusqlite::SqliteError> {
-    ///     let id_bindings = repeat("?").take(ids.len()).collect::<Vec<_>>().connect(", ");
+    ///     let id_bindings = std::iter::repeat("?").take(ids.len()).collect::<Vec<_>>().connect(", ");
     ///     let mut stmt = try!(conn.prepare(&*format!("SELECT name FROM people WHERE id IN ({})", id_bindings)));
     ///     let mut rows = try!(stmt.query_iter(ids.iter()));
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -525,6 +525,44 @@ impl<'conn> SqliteStatement<'conn> {
         }
     }
 
+    /// Execute the prepared statement, supplying parameters via an iterator and
+    /// returning an iterator over the resulting rows.
+    ///
+    /// ## Example
+    ///
+    /// ```rust,no_run
+    /// # use rusqlite::{SqliteConnection, SqliteResult};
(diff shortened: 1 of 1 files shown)


jgallagher:
This is interesting. I want to play with it a little bit before I merge - may be a few days.

marcusklaas:
Maybe it would be nice to take a `ToIterator` trait instead?

pfernie:
I experimented some more, as well as with the IntoIterator suggestion from @marcusklaas, and I think the change here (or with the IntoIterator) is not the right answer, hence the close on this PR. In particular:
* In my explorations with IntoIterator, the "working" solution had a number of gotchas, and notably ended up breaking the ergonomics of 
```rust
conn.query("SELECT foo FROM bar", &[]); // type inference on &[] fails
```
* For my actual use-case, I actually have some dynamic number of params where some are, say, i32, and other, say, str, and the changes in this PR don't make that any better really. This had been papered over in my test usage as I was just converting everything to str. I think the better solution is to instead use the existing query() like below:
```rust
    #[test]
    fn test_query() {
        let db = checked_memory_handle();
        db.execute_batch("CREATE TABLE person(id INTEGER, name VARCHAR);").unwrap();

        let mut insert_stmt = db.prepare("INSERT INTO person(id, name) VALUES(?, ?)").unwrap();
        assert_eq!(insert_stmt.execute(&[&0i32, &"foo".to_string()]).unwrap(), 1);
        assert_eq!(insert_stmt.execute(&[&1i32, &"bar".to_string()]).unwrap(), 1);
        assert_eq!(insert_stmt.execute(&[&2i32, &"baz".to_string()]).unwrap(), 1);

        {
            let ids = vec![0i32, 2];
            let names = vec!["goo", "gus", "bar"];
            let sql = format!("SELECT name FROM person WHERE id IN ({}) OR NAME IN ({}) ORDER BY id ASC",
                repeat("?").take(ids.len()).collect::<Vec<_>>().connect(", "),
                repeat("?").take(names.len()).collect::<Vec<_>>().connect(", "),
            );
            let mut query = db.prepare(&sql).unwrap();
            let params = ids.iter().map(|p| p as &types::ToSql)
                .chain(names.iter().map(|p| p as &types::ToSql)).collect::<Vec<_>>();
            let rows = query.query(params.as_slice()).unwrap();
            let v: Vec<String> = rows.map(|r| r.unwrap().get(0)).collect();
            assert_eq!(v.as_slice(), ["foo", "bar", "baz"].as_slice());
        }
    }

```
  * If you're at all interested in the IntoIterator experiments, I put up a separate [branch](https://github.com/pfernie/rusqlite/tree/query-into-iterator) with where I got with that. Again, the function signature may be overly complex, which could be why I was tripped up. I like the idea of using the IntoIterator trait for query/execute, but I hit a lot of roadblocks when having non-homogenous sets of params, and ended up having to cast to &ToSql before calling query/execute anyway, meaning the existing implementation would have been sufficient anyway.

Thanks for taking the time to look; I'd be happy to add the use case above to documentation if you're interested, although in retrospect I think it's maybe somewhat obvious and not really germane to the library usage.

PR #18, opened 2015-02-04
Implement standard Error trait for SqliteError

jshs:
The [`Error`](http://doc.rust-lang.org/nightly/std/error/index.html) trait provided by the standard library is useful for building higher-level interfaces.

I also made the `Display` format a little bit more user-friendly, as opposed to the `Debug` representation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,6 +57,7 @@ extern crate libc;
 use std::mem;
 use std::ptr;
 use std::fmt;
+use std::error;
 use std::rc::{Rc};
 use std::cell::{RefCell, Cell};
 use std::ffi::{CString};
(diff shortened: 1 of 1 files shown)


jgallagher:
Nice, I've been meaning to look at this. Is there a reason you're running `ffi::code_to_str` inside the `Error` implementation instead of returning `self.message`?

jshs:
I'm not sure how verbose `description()` is intended to be. I've had a look at a couple of existing `Error` impls, and that method always returns one of a set of fixed strings which roughly classifies the error. A more detailed representation is usually provided via the display format. For example, see [`std::old_io::IoError`](https://github.com/rust-lang/rust/blob/ce5aad2f107e79c5f1baab40aff35b7899322d94/src/libstd/old_io/mod.rs#L360-L376) and [`std::os::MapError`](https://github.com/rust-lang/rust/blob/ce5aad2f107e79c5f1baab40aff35b7899322d94/src/libstd/os.rs#L856-L891).
I tried to mirror this behavior, but I think returning `self.message` is just as good.

PR #19, opened 2015-02-07
`SqliteConnection::open` takes a `std::path`.

### LINKED ISSUE #13: Non-idiomatic constructor
Firstly, please me tell me if I'm intruding on your project too much!

I was thinking the constructor for `SqliteConnection` could be made more idiomatic by passing it a `&Path` instead of a `&str`. An added benefit would be that non-utf8 file paths could be opened as well.

A different constructor would have to be added for an in-memory connection.

What do you think?

jgallagher:
Add new constructors, `open_in_memory` and `open_in_memory_with_flags`,
for opening in-memory databases.

Closes #13.


jgallagher:
Hm, this is going to break opening paths on Windows. Any chance anyone watching this repo is using windows and can comment on the best way to fix this?

marcusklaas:
How so? I have access to a windows machine. I'll try to run some tests.

jgallagher:
I had been using one of the `std::os::unix::...` modules, so I think it wouldn't have even compiled on Windows (no idea about under cygwin?). I switched to using a platform-independent way of converting `Path` to C-style strings, so this should be ok now. It is a breaking change, unfortunately.

PR #26, opened 2015-03-29
Use `std::ptr::Unique` to ensure SqliteConnection is Send.

huonw:
Semantically the C resource is owned by the SqliteConnection, so Unique
is correct. Being Send is safe because sqlite3 can only possibly be
thread unsafe in cases when there is simultaneous access from multiple
threads, but transferring ownership between threads doesn't allow that
to occur.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@
 //!     }
 //! }
 //! ```
-#![feature(unsafe_destructor)]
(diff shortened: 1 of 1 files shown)


marcusklaas:
Nice! I bumped into this problem before, but didn't know it could be solved by using a unique pointer. :+1:

jgallagher:
Thanks!

PR #35, opened 2015-05-04
Change path parameter type to &AsRef<Path>

marcusklaas:
This should close https://github.com/jgallagher/rusqlite/issues/33. The pull request also adds a small test to make sure the new behaviour is as expected.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusqlite"
-version = "0.0.16"
+version = "0.0.17"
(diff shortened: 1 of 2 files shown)


jgallagher:
Thanks! I'm going to wait to merge this until this evening - right now master requires rustc-nightly (due to the use of `ptr::Unique`), and the stable branch is what's actually published to crates.io. I'm going to merge stable into master to avoid having to track both (since the difference is so small), and then I'll merge this too.

marcusklaas:
Cool. Always a pleasure :+1:

PR #38, opened 2015-05-05
Make `query_row` a synonym for `query_row_safe`.

jgallagher:
This is a breaking change for anyone using `query_row`. To update code
that used the old `query_row`, you must now `.unwrap()` the returned
result.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -274,28 +274,27 @@ impl SqliteConnection {
     /// ## Example
     ///
     /// ```rust,no_run
-    /// # use rusqlite::{SqliteConnection};
-    /// fn preferred_locale(conn: &SqliteConnection) -> String {
-    ///     conn.query_row("SELECT value FROM preferences WHERE name='locale'", &[], |row| {
+    /// # use rusqlite::{SqliteResult,SqliteConnection};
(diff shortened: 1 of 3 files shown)


marcusklaas:
Yes, exactly. Minus the lifetime 'a, I believe.

marcusklaas:
The primary drawback I can think of is that this will always do a copy on each row. I can imagine that for some applications, you wouldn't need to do that. When decrypting raw data from the database, for example. You'd only need to see data once.

jgallagher:
Yeah, I think we should have both. Do we:

* Leave `query` as it is, and add this as `query_map` (or similar)?
* Make this the new `query`, and rename the existing one `query_raw` (or similar)?

marcusklaas:
I think `query_map` is a good name. This approach will also preserve backwards compatibility.

jgallagher:
That's the way I'm leaning too, although I'd prefer to shepherd people toward the safer variant. Maybe we can do that in the docs/README.

marcusklaas:
Sounds good.

jgallagher:
There's been a handful of changes (some breaking) today, so I'd like to bump to 0.1 before publishing again. I think query_map should be in that version. Are you actively working on it, or shall I?

marcusklaas:
Sorry for the late response. I'd like to give it a go, if that's alright.

PR #41, opened 2015-05-05
Initial implementation for query_map

marcusklaas:
Here's a coarse implementation of what we discussed yesterday. It needs some refinement still. For example: passing a reference to a `Fn` isn't very nice ergonomically.

It also needs documentation and so on, but I thought I'd run this by you first.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,13 +36,16 @@
 //!                  &[&me.name, &me.time_created, &me.data]).unwrap();
 //!
 //!     let mut stmt = conn.prepare("SELECT id, name, time_created, data FROM person").unwrap();
-//!     for row in stmt.query(&[]).unwrap().map(|row| row.unwrap()) {
-//!         let person = Person {
+//!     let mut rows = stmt.query(&[], |row| {
+//!         Person {
 //!             id: row.get(0),
 //!             name: row.get(1),
 //!             time_created: row.get(2),
 //!             data: row.get(3)
-//!         };
+//!         }
+//!     }).unwrap();
+//!
+//!     for person in rows {
 //!         println!("Found person {:?}", person);
 //!     }
 //! }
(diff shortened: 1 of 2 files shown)


jgallagher:
Thanks! I pushed up a branch with two changes (diff here: https://github.com/jgallagher/rusqlite/compare/d3ab56c...marcusklaas-query-map):

* We can avoid the reference to a `Fn` by storing a boxed `Fn` instead. I don't have a strong opinion about this. Passing `&Fn` would avoid the box allocation, so I'd be happy to revert that change.
* I think MappedRows can hold onto a SqliteRows instead of duplicating what it does. What do you think?

marcusklaas:
Cool. The reason I duplicated most of `SqliteRows` is to avoid having to check for staleness. You are probably right that this too much effort and duplication to avoid one simple runtime check.

I was thinking a bit about it and I believe we can completely replace `query` by `query_map` for all cases when we pass an `FnMut` instead of an `Fn`. That way we can also handle the case where we only need to inspect data and mutate some state without doing a copy. Implementing `FromSql` for `&str` and `&[u8]` will make this much easier.

Also, I found out that we can simple store the passed function in our iterator, even if we do not its length at compile time. That's what the standard library does for `Iterator::map` (http://doc.rust-lang.org/src/core/iter.rs.html#1617). I think this is the way to go.

marcusklaas:
I made the changes I mentioned earlier. Some may be a bit far reaching, so please only what you like, even if that set is empty ;-)

Still not 100% happy with it. The reference to `SqliteStatement` in `MappedRows` should be mutable, I think. Also, `query_row` should probably take an `FnOnce` again.

jgallagher:
Definitely agree on storing the function in the iterator. Good call.

I'm not sure I completely understand the `FnMut` bit. One thing you can do now with `query` is mutate a local (stack) variable; I don't think the borrow checker will let us pass a closure that has a reference to a local var into `query_map`, will it?

marcusklaas:
You are correct, it won't. I was overzealous in thinking that `FnMut` would cover all use cases. We will indeed need `query` to things like conditional field evaluation, i.e.,

    if row.get::<i64>(0) == 1 { row.get(1) } else { 0 }

Apologies for making such a mess!

marcusklaas:
I'll close this pull request and open a new one that tweaks the query-map branch.

PR #42, opened 2015-05-07
Slightly adjust the signature of query_map

marcusklaas:


src/lib.rs
```diff
@@ -636,17 +639,20 @@ impl<'conn> SqliteStatement<'conn> {
         Ok(SqliteRows::new(self))
     }
 
-    pub fn query_map<'a, 'map, T, F>(&'a mut self, params: &[&ToSql], f: F)
-                                     -> SqliteResult<MappedRows<'a, 'map, T>>
+    /// Executes the prepared statement and maps a function over the resulting
+    /// rows. 
+    ///
+    /// Unlike the iterator produced by `query`, the returned iterator will
+    /// never panic.
```
jgallagher: This claim is too strong - `f` might panic itself, for example. Maybe "Unlike the iterator produced by `query`, the returned iterator does not expose the possibility for accessing stale rows." ?

src/lib.rs
```diff
@@ -700,16 +706,19 @@ impl<'conn> Drop for SqliteStatement<'conn> {
     }
 }
 
-pub struct MappedRows<'stmt, 'map, T> {
+/// An iterator over the mapped resulting rows of a query which never panics.
```
jgallagher: Here we could probably just drop "which never panics".

src/lib.rs
marcusklaas: That's much better. Will change.

marcusklaas:
Updated the comments

PR #44, opened 2015-05-07
Implement FromSql for &'a str, &'a [u8]

marcusklaas:


jgallagher:
I don't think this isn't memory-safe, at least the way the lifetimes are currently written. The returned `&str` is built from a pointer under SQLite's control, and its documentation states that the returned pointer is only valid for a limited (but documented) time. This test fails, for example:

```rust
    #[test]
    fn test_str() {
        let db = checked_memory_handle();

        db.execute("INSERT INTO foo(t) VALUES ('hello')", &[]).unwrap();
        db.execute("INSERT INTO foo(t) VALUES ('world')", &[]).unwrap();

        let mut stmt = db.prepare("SELECT t FROM foo ORDER BY t ASC").unwrap();
        let mut rows = stmt.query(&[]).unwrap();

        let hello: &str = rows.next().unwrap().unwrap().get(0);
        let world: &str = rows.next().unwrap().unwrap().get(0);

        assert_eq!(hello, "hello");
        assert_eq!(world, "world");
    }
```

because after the `let world` line, both pointers happen to be pointing to "world". (It could just as easily have crashed, if SQLite had freed "hello" and put "world" somewhere else.)

marcusklaas:
Good catch! Is there a way we could salvage this? This should be okay in theory in a `query_map`, I think..

jgallagher:
Would it be useful in `query_map` since it requires `T: 'static`? If you could get out a reference like this, you'd still have to copy it before returning, right?

marcusklaas:
It could still be useful:

```rust
fn has_name(db: &SqliteDatabase, name: &str) -> bool {
    db.prepare("SELECT name FROM foo")
        .unwrap()
        .query_map(&[], |row| row.get::<&str>(0) == name)
        .any(|b| b)
}
```

jgallagher:
Ah, yeah, I see. Unfortunately I don't think it's possible to make this completely safe. We could fix the staleness problem by tying to the lifetime of an individual row somehow, but SQLite documents that the returned pointer can also be invalidated if a type conversion happens on the column. So something like this:

```rust
let s: &str = row.get(0);
let i: i32 = row.get(0);
```

while silly, would result in `s` holding an invalid pointer.

This seems like kind of a performance optimization edge case. We could maybe come up with an `unsafe` variant of this (and point to the SQLite documentation where it talks about the pointer invalidation).

PR #45, opened 2015-05-11
Implement `query_map`

### LINKED ISSUE #40: Add `query_map` to SqliteStatement.
See #38 for discussion

jgallagher:
Closes #40.

@marcusklaas Want to take one more look over this before we merge? We're at the point now where the closure passed to `query_map` can even refer to things on the stack. I guess the only thing it can't do is create some kind of `T` that isn't static but doesn't refer to the row; I'm not sure it's possible to express that kind of restriction. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,9 +90,10 @@ fn bad_function_will_panic(conn: &SqliteConnection) -> SqliteResult<i64> {
 There are other, less obvious things that may result in a panic as well, such as calling
 `collect()` on a `SqliteRows` and then trying to use the collected rows.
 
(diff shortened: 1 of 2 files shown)


marcusklaas:
I think the `'static` bound really only restricts us from doing stuff that we shouldn't be doing (holding onto references to stale rows). We can still generate values which are non static, we just have to create those in a different part of the chain:

```rust
statement.query_map(param_list, |row| {
    ...
})
.map(f)
```

where `f` maps our static type to a non-static type.

It all looks good to me! :airplane:

PR #53, opened 2015-08-01
Extra check in SqliteStatement.execute

gwenn:
Ensure `SqliteStatement.execute` is not used with a SELECT which returns no row.


jgallagher:
:+1: thanks!

PR #54, opened 2015-08-01
Reset as soon as possible

gwenn:


jgallagher:
I'm not sure if this is the right thing to do from a performance point of view. If a statement is only going to be executed once (which I think is probably common), this will unnecessarily reset it the statement. This is probably super minor, so I'm happy to be convinced - is there another reason to reset ASAP?

gwenn:
I don't find anything on the sqlite website.
But here is suggestion:
https://books.google.fr/books?id=HFIM47wp0X0C&pg=PA130&dq=sqlite+should+be+reset+as+soon+as+you
’re+done+using+it&hl=en&sa=X&ved=0CCIQ6AEwAGoVChMI7_DxuLuIyAIVCwoaCh0DrwF-#v=onepage&q=sqlite%20should%20be%20reset%20as%20soon%20as%20you’re%20done%20using%20it&f=false
"A statement should be reset as soon as you’re done using it. This will
ensure any locks are released."
The books is quite old (2010)...

jgallagher:
Interesting. I don't see anything in the SQLite source code that suggests a statement could hold a lock the way the book is describing, but it seems reasonable to err on the side of caution here. Can you merge with `master` to clean this PR's history?

jgallagher:
Ahh, here is some info on the SQLite site that confirms we should reset ASAP: http://www.sqlite.org/cvstrac/wiki?p=ScrollingCursor. This is also quite dated, but is more authoritative. The locks they're referring to are process-level

PR #55, opened 2015-08-01
Debug db path and stmt sql.

gwenn:


jgallagher:
Thanks! These commits were squashed and merged in #69.

PR #61, opened 2015-08-02
LRU statement cache

gwenn:


src/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /// Search the cache for a prepared-statement object that implements `sql`.
+    // If no such prepared-statement can be found, allocate and prepare a new one.
+    ///
+    /// # Failure
+    ///
+    /// Will return `Err` if no cached statement can be found and the underlying SQLite prepare call fails.
+    pub fn get(&mut self, sql: &str) -> Result<Statement<'conn>> {
```
jgallagher: I'm inclined to think this should return a `Result<CachedStatement<'cache>>`. `CachedStatement` would wrap up `Statement` (and implement `Deref`/`DerefMut` for it too, I think), and would return it to the cache in `drop`. This would remove the need to remember to call `release`. We could actually remove `release` altogether, I think, and add a `release` or `discard` method to `CachedStatement` that would (a) take `self` (to consume it) and (b) set a flag (or something similar) to avoid returning the statement to the cache in `drop`.

What do you think?

src/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+            return stmt.finalize();
+        }
+        stmt.reset_if_needed();
+        // clear bindings ???
+        self.cache.insert(stmt.sql(), stmt).map_or(Ok(()), |stmt| stmt.finalize())
+    }
+
+    /// Flush the prepared statement cache
+    pub fn flush(&mut self) {
+        self.cache.clear();
```
jgallagher: Why change the name from `clear` to `flush`?

src/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// Return (current, max) sizes.
+    pub fn size(&self) -> (usize, usize) {
+        (self.cache.len(), self.cache.capacity())
```
jgallagher: I think we should keep these as two separate methods (so people don't have to look up the docs to figure out what the returned tuple means).

src/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// Set the maximum number of cached statements.
+    pub fn set_size(&mut self, capacity: usize) {
+        self.cache.set_capacity(capacity);
```
jgallagher: Similarly, I think we should keep the `set_capacity` name here too.

src/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+        let db = Connection::open_in_memory().unwrap();
+        let mut cache = StatementCache::new(&db, 10);
+        let sql = "PRAGMA schema_version";
+        let mut stmt = cache.get(sql).unwrap();
+        // println!("NEW {:?}", stmt);
+        cache.release(stmt, false).unwrap();
+        stmt = cache.get(sql).unwrap();
+        // println!("CACHED {:?}", stmt);
+        cache.release(stmt, true).unwrap();
+        cache.flush();
```
jgallagher: This isn't testing much. I think the cache could be exercised more thoroughly:

* Check that the size is 1 after a statement is released
* Ensure that the cached queries work

src/lib.rs
```diff
@@ -935,6 +936,14 @@ impl<'conn> Statement<'conn> {
         }
     }
 
+    fn sql(&self) -> String { // TODO Maybe SQL should by kept as an SqliteStatement field ?
```
jgallagher: Couple comments:

* This may need to be `#[cfg(feature = "cache")]` flagged to avoid a dead code warning if you're compiling without the cache feature.
* Keeping the SQL as a field on `Statement` is tempting, but I'm inclined not to (and keep this as-is). SQLite is already keeping a copy around; no need to keep a second.

src/cache.rs
gwenn: No reason at all...

src/cache.rs
gwenn: Definitely

src/cache.rs
gwenn: Ok.

src/lib.rs
gwenn: You are right.

src/cache.rs
gwenn: Sorry, I've been lazy about testing...

jgallagher: No worries! I know on most of your PRs I've added a bunch of stuff - if you want me to do that again on this PR and you look it over, that's fine with me. I'm working on #58 and a fix for #79 and figured we could get everything in faster if you have time to update this one. If not I can get to it later :)

src/cache.rs
gwenn: Yes, introducing a `CachedStatement` is the right way.

src/cache.rs
gwenn: I have been working on user defined aggregate functions (https://github.com/gwenn/rusqlite/tree/aggregate) but I will try to fix all your remarks.

gwenn:
I am sorry to bother you but I cannot make the CachedStatement work: `FIXME cannot borrow immutable borrowed content `*self.cache` as mutable`.

jgallagher:
Good news / bad news. Good news is you can get around the issues in `drop` by putting the statement inside an `Option` (which we'll guarantee is always `Some()` and is therefore safe to call `.unwrap()` on), so you can `.take()` the value out during destruction:

```rust
pub struct CachedStatement<'conn> {
    stmt: Option<Statement<'conn>>,
    cache: RefCell<StatementCache<'conn>>,
    pub cacheable : bool,
}

impl<'conn> Drop for CachedStatement<'conn> {
    #[allow(unused_must_use)]
    fn drop(&mut self) {
        if self.cacheable {
            self.cache.borrow_mut().release(self.stmt.take().unwrap(), false);
        } else {
            self.stmt.take().unwrap().finalize();
        }
    }
}
```

Bad news is it looks like `lru-cache` doesn't work on stable or beta rust. It [depends on linked-hash-map](https://github.com/contain-rs/lru-cache/blob/master/Cargo.toml#L18) which uses [an unstable feature](https://github.com/contain-rs/linked-hash-map/blob/master/src/lib.rs#L31). I don't think we can merge this in and keep rusqlite usable on stable rust. :(

gwenn:
Thanks.
Maybe this [one](https://crates.io/crates/lru_time_cache) works on stable ?
The travis configuration file seems to say so.

gwenn:
I will need your help: how to create a `RefCell` from a reference ?
```
src/cache.rs:78         stmt.map(|stmt| CachedStatement::new(stmt, RefCell::new(self)))
```

jgallagher:
Mm, that's never going to work - if it did, you'd end up with multiple `&mut StatementCache`s in each statement, which should be impossible.

What if you put the internal cache itself inside a `RefCell`? Then `get` and `release` could take `&self` (they could `borrow_mut()` the underlying storage), and `CachedStatement` could hold a plain `&'a StatementCache`?

gwenn:
The implementation is not perfect but it seems to be working...
I will try to make some benches to show the difference between cache and no cache.

gwenn:
test bench_cache    ... bench:         124 ns/iter (+/- 76)
test bench_no_cache ... bench:       5,538 ns/iter (+/- 4,178)

jgallagher:
Merged as part of #113.

PR #66, opened 2015-08-27
Change invalid idx behavior of SqliteRow::get_checked()

pfernie:
This changes the behavior of get_checked() to first check the validity of the index & staleness of the row, and then check that column type is a valid source for T. Current behavior is (which contradicts the documented behavior of panic!() on invalid idx) to return SQLITE_MISMATCH/"Invalid column type" on an invalid index (return of sqlite3_column_type() is undefined on an invalid index).

This is a simple copy/paste of some logic from get_opt().

Alternatively, this could be implemented as an additional fn get_opt_checked() or somesuch, but seems like API proliferation... I would in fact think it makes sense to remove get_opt() also...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -302,6 +302,37 @@ impl SqliteConnection {
         }
     }
 
+    /// Convenience method to execute a query that is expected to return a single row,
+    /// and execute a mapping via `f` on that returned row with the possibility of failure.
+    /// The `Result` type of `f` must implement `std::convert::From<SqliteError>`.
+    ///
+    /// ## Example
+    ///
+    /// ```rust,no_run
+    /// # use rusqlite::{SqliteResult,SqliteConnection};
+    /// fn preferred_locale(conn: &SqliteConnection) -> SqliteResult<String> {
+    ///     conn.query_row_and_then("SELECT value FROM preferences WHERE name='locale'", &[], |row| {
+    ///         row.get_checked(0)
+    ///     })
+    /// }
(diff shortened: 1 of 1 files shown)


pfernie:
This also implements 2 more convenience methods: SqlStatement::query_and_then() and SqlConnection::query_row_and_then(). query_and_then() is analogous to query_map(), but the provided function is expected to return Result<T, E> where E: From<SqliteError>; similarly query_row() <-> query_row_and_then(). This makes the API slightly more ergonomic in certain cases; I often write builder methods such as
````
// MyError : From<SqliteError> ...
impl Foo {
  fn from_row(row: SqliteRow) -> Result<Foo, MyError> {
    if bad_phase_of_moon {
      Err(MyError::MoonPhase)
    } else {
      Ok(Foo {
        id: try!(id.get_checked(0)),
      })
    }
  }
}
````

Suggestions for better naming welcome; current names a result of using Result::and_then() in the impl.

Alternatively, I could see an argument for simply making this the signature for query_map()/query_row() but this seems like a large API change and introduces burden on clients when they "know" a simple .get(0) isn't going to panic, etc.

jgallagher:
Thanks for the fix in `get_checked` and the new `*_and_then` methods. I want to keep `query_map`/`query_row` - I could maybe see an argument for renaming `query_row` -> `query_row_map` for consistency, but I think I still prefer the former.

Also, you're right that `get_opt` is now superfluous - removed it in #70.

PR #68, opened 2015-09-08
Relax uses of `P: AsRef<...>` from `&P` to `P`.

huonw:
This means that one can pass `AsRef` types directly, without having to
insert a `&`, e.g. `SqliteConnection::open("foo.db")` (new)
vs. `SqliteConnection::open(&"foo.db")` (old).

This should be backwards compatible, since there is an impl in the
standard library:

    impl<'a, T, U> AsRef<U> for &'a T where U: ?Sized, T: AsRef<U> + ?Sized

I.e. the old `&P` satisfies the new bound still. (Taking `P` directly is
what the standard library does with similar functions, like
`File::open`.)


jgallagher:
:+1: thanks!

PR #73, opened 2015-10-14
Avoid the RFC1214 warning on nightly.

huonw:
The compiler was tightened up a little to check things more rigorously,
which caused some new errors to be generated (the errors are being
phased in as warnings at the moment, but it doesn't hurt to fix it
pre-emptively).

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -75,7 +75,8 @@ pub trait ToSql {
 
 /// A trait for types that can be created from a SQLite value.
 pub trait FromSql {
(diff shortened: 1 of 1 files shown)


Dashed:
@huonw I think this is probably better approach: https://github.com/jgallagher/rusqlite/pull/74/files

huonw:
I mildly disagree, but whatever. (It would be better written as `trait FromSql: Sized { ... }` in any case.)

jgallagher:
Addressed in #75. Thanks!

PR #74, opened 2015-10-28
RFC 1214 fixes

steveklabnik:
This will break on Rust 1.5, and this patch fixes it

You may want to release a new version to crates.io as well.


jgallagher:
Addressed in #75. Thanks!

PR #78, opened 2015-11-08
Update bindings with bindgen 0.15.0/sqlite3 3.8.10

gwenn:


Dashed:
@gwenn Do you happen to know what version of sqlite3 that `libsqlite3-sys` bindings on master is generated against?

gwenn:
I should know but I don't.
I can only guess by looking at http://sqlite.org/changes.html: [3.8.5 (sqlite3_rtree_query_callback), 3.8.7 (sqlite3_malloc64)[

jgallagher:
This seems problematic, and I'm not exactly sure how to handle it. 3.8.10 is relatively new; even 3.8.7 is newer than the SQLite available on several platforms:

* OS X 10.10 shipped with 3.8.5; 10.9 shipped with 3.7.13.
* Travis-CI is using Ubuntu LTS versions; even moving to Ubuntu Trusty (which only [became available](https://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta) last month) has SQLite 3.8.2.

I'd rather rusqlite still support those platforms. I wonder if the right solution is to go ahead and update libsqlite3-sys to reflect the latest and greatest, but avoid using any features newer than some reasonable cutoff point in rusqlite itself. Very open to better alternatives.

jgallagher:
Closing in favor of #100.

PR #81, opened 2015-11-28
libsqlite3-sys: build on Mac OS X by linking to system library

astraw:
This change will let libsqlite3-sys (and hence rusqlite) compile on Mac OS X using the system sqlite3 installation.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -1,5 +1,14 @@
 extern crate pkg_config;
 
+use std::env;
+
 fn main() {
-    pkg_config::find_library("sqlite3").unwrap();
+    let target = env::var("TARGET").unwrap();
+
+    if target.contains("darwin") {
+        println!("cargo:rustc-link-lib=sqlite3");
+        println!("cargo:rustc-link-search=/usr/lib");
+    } else {
+        pkg_config::find_library("sqlite3").unwrap();
+    }
 }


jgallagher:
Thanks, this is a good idea. A couple of thoughts:

* I'm inclined to think we should only fall back to this if `pkg-config` fails. E.g., on OS X, if I've installed a newer version of SQLite via Homebrew and configured `pkg-config` to point to it, that should take precedence.
* In the same vein, I wonder if it'd be useful to allow users to specify the search path via an environment variable; e.g, something like

```rust
if let Some(search_path) = option_env!("LIBSQLITE3_SYS_LIB_PATH") {
    println!("cargo:rustc-link-search={}", search_path);
}
```

happening before either pkg-config or defaulting to `/usr/lib`. What do you think?

astraw:
Yes, good points. I would suggest (in this order): 1) use environment variable if present, 2) use pkg-config if valid result found, and 3) system install if present. Finally, one could consider as number 4 potentially shipping sqlite3 and building from source as in [libcurl](https://github.com/carllerche/curl-rust/blob/master/curl-sys/build.rs).

[Cargo features](http://doc.crates.io/manifest.html#the-features-section) could also be used to force a particular behavior, but I'm not sure it's a good fit here.

It seems the rust/cargo community should have some kind of consensus for this issue, as it will be a while before all system libraries are re-written in rust. :)

astraw:
On the topic of consensus within the community: `libgit2-sys` is mentioned in [the cargo documentation](http://doc.crates.io/build-script.html#case-study-linking-to-system-libraries) as solving many of these issues. So, [the libgit2-sys build.rs file](https://github.com/alexcrichton/git2-rs/blob/master/libgit2-sys/build.rs) is worth looking over for ideas.

jgallagher:
@astraw Want to take a look at #88? I think it should do the basic thing.

I also opened #89 to track the desire to more closely follow what libgit2-sys does (building the lib from source if it isn't found). I don't currently have access to a Windows system to work on that, and it looks like that's where a lot of the complexity happens.

jgallagher:
Closing in favor of #88.

PR #84, opened 2015-12-01
Update libc version to ~0.2 and fix mod test

krdln:
I've only changed version in Cargo.tomls, because libc 0.1 had incorrect signedness of char on ARM and the crate didn't compile. I've also removed redundant `extern crate`s from mod test, because `cargo test` failed to compile for me. I hope new libc don't break anything, but I've tested only on ARM, so YMMV.


krdln:
Huh, interesting. Did something changed in nightly Rust in how `use super::*` work? Anyway, I don't really know how to remove a commit from PR. You can close it, I guess.

PR #86, opened 2015-12-01
Add Rust interface to SQLite's logging, tracing, and profiling hooks

jgallagher:
Builds on #56, providing a slightly more Rust-y interface (users can use pure Rust fns instead of `extern "C"` fns).

@gwenn want to look this over?


gwenn:
Perfect !

PR #94, opened 2015-12-11
Add support for named parameters

jgallagher:
Builds on #63.

@gwenn do you want to look this over? I made several small changes; the bigger ones are:

* Added assertions to check that we actually bind all the expected parameters. This is tricky and I'm not totally happy with what I came up with - I'm always `assert!()`-ing that the _number_ of parameters matches, and an extra sanity check is to `debug_assert!()` that all the names are unique.
* Renamed `query_named_row` -> `query_row_named`; I think that's slightly more consistent with things like `iter` / `iter_mut`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Version UPCOMING (TBD)
 
 * Adds `backup` feature that exposes SQLite's online backup API.
+* Adds a variety of `..._named` methods for executing queries using named placeholder parameters.
 
 # Version 0.5.0 (2015-12-08)
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -369,20 +369,12 @@ impl SqliteConnection {
     /// underlying SQLite call fails.
     pub fn query_row<T, F>(&self, sql: &str, params: &[&ToSql], f: F) -> SqliteResult<T>
         where F: FnOnce(SqliteRow) -> T
-        {
-            let mut stmt = try!(self.prepare(sql));
-            let mut rows = try!(stmt.query(params));
+    {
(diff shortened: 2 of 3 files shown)


gwenn:
Why prevent from reusing previous bindings or initial NULL bindings ?
```rust
let mut stmt = db.prepare("INSERT INTO test (a, b, c) VALUES (:a, :b, :c)").unwrap();
stmt.execute_named(&[(":a", &"a1"), (":b", &"b1")]).unwrap(); // c is NULL
stmt.execute_named(&[(":a", &"a2"), (":c", &"c2")]).unwrap(); // b is 'b1'
...
```

jgallagher:
I was actually thinking about that yesterday (not just for named parameters). I'm a little concerned about memory safety - e.g., a `ToSql` value passed to a previous bind being deallocated but then reused. All the `ToSql` implementations that we provide are fine (because they use `SQLITE_TRANSIENT` to tell SQLite to make its own copy of the data), so maybe I shouldn't worry about it. Or should add a note about it to the docs of `ToSql`, maybe, and then allow this?

gwenn:
If a custom destructor is passed/used, it will be called by `sqlite3_clear_bindings` or `sqlite3_finalize`: so no lifetime problem (or at least used after free).
Is there any way to (safely) use `SQLITE_STATIC` ? With static lifetime variables ?

jgallagher:
Ah, yeah, good point. Ok, I'll take the assertions back out and allow null / reuse.

`SQLITE_STATIC` would be safe for anything that is guaranteed to outlive the statement. `'static` lifetime variables would definitely be fine; I'm not sure it'd be possible to effectively use anything else.

PR #95, opened 2015-12-11
Add support for user-defined functions

jgallagher:
Builds on #64.

@gwenn Do you want to look this one over too? I made more substantial changes this PR to allow passing in closures and to make writing the functions themselves less painful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,9 @@ script:
     - cargo test --features backup
     - cargo test --features load_extension
     - cargo test --features trace
-    - cargo test --features "backup load_extension trace"
-    - cargo doc --no-deps --features "backup load_extension trace"
+    - cargo test --features functions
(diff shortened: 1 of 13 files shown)


Dashed:
This looks really slick. I'm using #64 PR on localized rusqlite build; but I'm looking forward to see minimized use of `unsafe` blocks.

gwenn:
Perfect!
I will try to implement aggregate functions...

gwenn:
Oops !
The number of parameter may be negative:
http://sqlite.org/c3ref/create_function.html
> The third parameter (nArg) is the number of arguments that the SQL function or aggregate takes. If this parameter is -1, then the SQL function or aggregate may take any number of arguments between 0 and the limit set by sqlite3_limit(SQLITE_LIMIT_FUNCTION_ARG). If the third parameter is less than -1 or greater than 127 then the behavior is undefined.

PR #98, opened 2015-12-13
Update bindings with bindgen 0.15.0/sqlite3 3.8.10

jgallagher:
Builds on #78. Puts the switch to text64/blob64 behind a Cargo feature since it requires a relatively new SQLite.

@gwenn this seems like the best we can do to keep supporting older SQLites. Are you reasonably happy with this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ load_extension = ["libsqlite3-sys/load_extension"]
 backup = []
 functions = []
 trace = []
+bind_text64 = []
 
 [dependencies]
 time = "~0.1.0"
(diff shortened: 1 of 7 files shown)


gwenn:
I am sorry.
I introduced sqlite3_bind_{text,blob}64 only to simplify the rust code by delegating the size check to SQLite.
So you can close/ignore my pull request.

jgallagher:
Oh, I didn't realize SQLite still only allows up to i32::MAX even with the `*64` functions. Yeah, I'll close this and your PR but keep the first commit that updates the bindings.

jgallagher:
Closing in favor of #100.

PR #99, opened 2015-12-13
Simplify execute_batch implementation.

gwenn:
There is no difference between calling `sqlite3_errmsg` or using the last parameter of `sqlite3_exec`:
```c
      memcpy(*pzErrMsg, sqlite3_errmsg(db), nErrMsg);
```


jgallagher:
:+1:

PR #101, opened 2015-12-14
Add feature for BLOB I/O.

jgallagher:
Builds on #58.

@gwenn Changes from #58:

* Moved read/write/seek out of `impl Blob` and into `impl std::io::{Read, Write, Seek}`s on `Blob`.
* Changed write to return `Ok(0)` instead of failing if asked to write more data than will fit in the blob. I'm not sure about this change, but I think it's more consistent with the way `std::io::Write` is expected to be implemented (e.g., `std::io::BufWriter` will return a `WriteZero` error if the underlying writer returns `Ok(0)` while it still has data buffered up that it wants to write).

I'm really on the fence with the change to `write`. What do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ name = "rusqlite"
 [features]
 load_extension = ["libsqlite3-sys/load_extension"]
 backup = []
+blob = []
 functions = []
 trace = []
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,7 @@
   The old, prefixed names are still exported but are deprecated.
 * Adds a variety of `..._named` methods for executing queries using named placeholder parameters.
 * Adds `backup` feature that exposes SQLite's online backup API.
+* Adds `blob` feature that exposes SQLite's Incremental I/O for BLOB API.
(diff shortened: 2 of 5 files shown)


gwenn:
I don't understand why you don't want to fail fast.
When `buf.len() > remaining BLOB size`, you *can't* write the whole buffer successfully.
Why do you do a partial write and then return Ok(0) ?

jgallagher:
It matches the description of [`Write::write`](https://doc.rust-lang.org/std/io/trait.Write.html), specifically:

> If the return value is Ok(n) then it must be guaranteed that 0 <= n <= buf.len(). A return value of 0 typically means that the underlying object is no longer able to accept bytes and will likely not be able to in the future as well, or that the buffer provided is empty.
>
> ...
>
> It is not considered an error if the entire buffer could not be written to this writer.

jgallagher:
Oh, I mentioned `BufWriter` above, but `Write::write_all` also knows to look for `Ok(0)` and treat it as an error: 3482e1c (source - https://doc.rust-lang.org/src/std/io/mod.rs.html#964-966).

gwenn:
Ok, `write_all` is perfect because users will not have to handle the `Ok(0)` case.

Issue #106, opened 2015-12-19
Need to know the number of columns


gwenn:
It seems that there is no simple way to iterate on the columns of a `Row` from a wildcard query (`select * from foo`).
We can only call `Row.get_checked` incrementally until an error is returned.
Or use `Statement.column_names` to know the number of columns.
Maybe the `column_count ` should be public ?

jgallagher:
Fixed in #109

PR #109, opened 2016-01-02
Make possible to execute dynamic queries.

gwenn:
Queries with dynamic column count/type.


gwenn:
Fix issue #106

gwenn:
Related issue #106 can be fixed.

PR #115, opened 2016-01-07
Add support for user-defined aggregate functions.

jgallagher:
Builds on #108.

@gwenn I changed how step/finalize are called if there are no rows - see b189f6b. I think this is the right thing to do since otherwise you couldn't implement the `my_count` example in the unit tests (I don't believe?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ script:
     - cargo build
     - cargo test
     - cargo test --features backup
+    - cargo test --features blob
     - cargo test --features load_extension
     - cargo test --features trace
     - cargo test --features functions
-    - cargo test --features "backup functions load_extension trace"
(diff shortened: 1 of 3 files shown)


gwenn:
You are right: `final` function should be called when there is no row.
But it seems useless to call `init`when there is no row !
Why not just change the signature of `final` to 
```rust
fn finalize(&self, Option<A>) -> Result<T>;
```
?

jgallagher:
Good idea. Working on that.

jgallagher:
Alright, fixed in ca761d7. I like that a lot, actually, because the type of `Sum`'s `Aggregate` impl makes more sense now: `A` is `i64`, but `T` is `Option<i64>` (since sum returns null if it ran with no rows).

gwenn:
Thanks.

PR #116, opened 2016-01-07
Add cache feature to Travis

jgallagher:


jgallagher:
Oops, this should be part of #113

PR #122, opened 2016-01-31
Introduce ZeroBlob struct.

gwenn:


jgallagher:
:+1: Code looks good. I think I'd be inclined to put this in the `blob` module instead of `types` (so it'd live alongside the `Blob<'a>` struct); what made you decide to put it in `types`?

gwenn:
There's no particular reason to put in `types.rs` file: I just did a copy/paste/change of the `Null` struct...
You are right: keep all blob related stuff in `blob.rs`. I will update the changes related to this PR.

jgallagher:
Thanks!

PR #121, opened 2016-01-28
Appveyor

gwenn:
CI on Windows platform:
https://ci.appveyor.com/project/gwenn/rusqlite

Issue #46, opened 2015-05-19
how to build rusqlite on windows?


majkcramer:
On Linux eco system i had no problems. But on windows I feel a little bit helpless. Could you provide some short hints on how to build rusqlite on windows, please?

Install pkg_config for windows? 
Which files I have to download from sqlite hompage? 
Where to place? 
...

Any hints?

jgallagher:
Sorry, I don't have access to a Windows box for development at the moment. I know Rust's support on Windows is evolving too. Have you tried asking in the #rust IRC channel? There might be folks there who can help.

jgallagher:
Oh! If you do figure something out, please reply; I'd love to document this.

majkcramer:
I have no final solution for now. Only figured out something. Some steps must be wrong.

## Part #a: compile sqlite.dll.a
- Download mingw32 or mingw64 
- Download the sqlite amalgamation source code from https://www.sqlite.org/download.html
- Install mingw
- Decompress amalgamation source
- "Start Terminal" from mingw Startmenu entry (cmd.exe with env path to mingw/bin/ folder). Do the following 2 steps within the cmd Terminal
- cd to you decompresses amalgamation source
- "dlltool --def sqlite3.def --dllname sqlite3.dll --output-lib sqlite3.a"
- now there should be a sqlite3.a within that folder
- optional: check the sqlite.a does not contain debugger symbols: "nm --debug-syms sqlite.a" (For debugger symbols the second column indicates N.)

## Part #b: download and setup pkg-config
Hint: Do yourself a favor und download the gtk+ bundle instead only "pkg-config" and trying to fulfill required dependency
- Download gtk+ bundle from http://www.gtk.org/download/index.php 
- Decompress the gtk+ bundle
- Copy "sqlite3.def" from amalgamtion source to {gtk+ folder}/lib/
- Copy your compileted "sqlite3.a" to {gtk+ folder}/lib/ as "sqlite3.dll.a"
- Within {gtk+ folder}/lib/pkgconfig/ create a "sqlite3.pc" file with following contents:
```
# Package Information for pkg-config

prefix=/srv/win32builder/fixed_364/build/win32
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: sqlite3
Description: SQL database engine lib
Version: 3.7.7.1
Libs: -L${libdir} -lsqlite3
Libs.private: -ldl -lpthread 
Cflags: -I${includedir}
```
[...]

PR #124, opened 2016-02-01
Add CI on Windows platform via Appveyor

jgallagher:
Builds on #121, opening up to all branches/PRs and adding tests on all features.


jgallagher:
Thanks, this is great. I still want to investigate the amalgamation option from #46, but this is a big step up from "Windows? No idea!"

PR #125, opened 2016-02-01
Fix warning with Rust 1.6

gwenn:
warning: lint raw_pointer_derive has been removed: using derive with raw
pointers is ok


jgallagher:
:+1:

PR #127, opened 2016-02-01
Introduce exists and insert convenient methods

gwenn:
I am not sure that they are really useful.
And there is a potential misuse/problem with `insert` for tables without rowid...

### DIFF
--- a/src/convenient.rs
+++ b/src/convenient.rs
@@ -1,0 +1,72 @@
+use super::ffi;
+
+use {Error, Result, Statement};
+use types::ToSql;
+
+impl<'conn> Statement<'conn> {
+    /// Execute an INSERT and return the ROWID.
+    ///
+    /// # Failure
+    /// Will return `Err` if no row is inserted or many rows are inserted.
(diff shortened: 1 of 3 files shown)


jgallagher:
I'm hesitant to include these, for different reasons:

* `insert` has at least a few ways it could be misused. Tables with rowid as you mentioned; you could also pass any query that changed one row (e.g., an `UPDATE ...`) and you'd get back whatever row id happened to last be inserted (potentially unrelated to the insert).
* I'm not sure what the relationship between `fn exists` and a SQL `SELECT ... WHERE EXISTS ...` query is or should be.

I guess the common thing in both of these is that they're Rust functions meant to be used with a particular kind of query, but we don't have any way of knowing for sure that they're being used with the intended query type. I'm not sure how to address that, though.

gwenn:
`exists` is used/declared in [TCL interface](http://sqlite.org/tclsqlite.html#exists).

jgallagher:
Merged as part of #146. Thanks!

PR #130, opened 2016-02-11
Virtual Table

gwenn:
It is not much tested.
But would you mind give it a look.
Currently, only read-only virtual table is supported.
Thanks.


jgallagher:
I think we need to do a lot more work on this interface to make it usable safely from clients. There are some functions that aren't marked `unsafe` that would need to be as written (such as `VTab::create` and `VTabCursor::filter`), but a better solution would be to do more work in rusqlite to avoid making the user deal with raw SQLite pointers. I'd also like to see if we can implement this without a macro to declare the module, and instead use more fleshed out traits.

This is just a rough sketch, but I'd like to see something like:

```rust
trait VTabModule {
    type Aux;
    type Table: VTab;

    fn create(aux: &Aux, args: &[&str], declare: &Fn(&str)) -> Result<Table>;
    fn destroy(vtab: Table) -> Result<()>;

    fn connect(aux: &Aux, args: &[&str], declare: &Fn(&str)) -> Result<Table>;
    // not sure we need disconnect at all - should this be handled by VTab's drop?
    // fn disconnect(vtab: Table) -> Result<()>;
}
// probably a separate EponymousVTabModule that only defines connect (possibly disconnect if needed)?

trait VTab {
    type Cursor: VTabCursor;

    fn open(&self) -> Result<Cursor>;
    // again not sure we need it, but maybe fn close(&self, cursor: Cursor)

    // not sure what args/return type best_index should have; this one is tough
    fn best_index(&self, ...) -> ...;
}

trait VTabCursor {
    fn eof(&self) -> bool;

    // args here are related to best_index above
    fn filter(...) -> Result<()>;

    fn next(&mut self) -> Result<()>;
    fn column_value(&self, col: c_int) -> Result<Value>;
    fn rowid(&self) -> Result<i64>;
    // ... etc
}

impl Connection {
    fn create_module<Module: VTabModule>(name: &str, aux: Module::Aux) -> Result<()> {
        // we create the sqlite3_module
        // the functions we provide create the necessary sqlite3_vtab / sqlite3_vtab_cursor too
        // ...
    }

    fn create_eponymous_module<Module: EponymousVTabModule>(name: &str, aux: Module::Aux) -> Result<()> { ... }
}
```

awakecoding:
Any update on virtual table support? This PR is large and goes back to 2016

gwenn:
Some methods are still unsafe.
And the `int_array` module should be removed/replaced by the `carray` module.
Also, @jgallagher suggested to avoid using macro but I don't know another way to generate a **static** instance of an `sqlite3_module`.
Sorry.

awakecoding:
I would like to start using this branch to make my own virtual table to wrap system information tables (https://github.com/wayk/siquery-rs/tree/vtab). Are the csv and array tables the best way to go? They appear to be very unsafe and too hard to compile outside of rusqlite. We are fine with making temporary modifications on the branch to get things working with frequent updates as this progresses. The array virtual table appears to be taking a pointer parameter, I wonder if we couldn't pass a trait instead that would could then be used inside a virtual table to iterate over the data for filtering. Any hints on best ways to start using this? Thanks!

gwenn:
@awakecoding Could you please give another try ?

awakecoding:
@gwenn I managed to get the test dummy virtual table compiling inside my project. I had to import both rusqlite + libsqlite3-sys from your git branch. Here is what I added to my Cargo.toml:

```
[dependencies.rusqlite]
git = "https://github.com/gwenn/rusqlite.git"
rev = "b19d050"
features = ["vtab", "bundled"]

[dependencies.libsqlite3-sys]
git = "https://github.com/gwenn/rusqlite.git"
rev = "b19d050"
```

and then added the required extern crate imports:

```
#[macro_use]
extern crate rusqlite;
extern crate libsqlite3_sys;
```

Now that I can compile a virtual table definition outside of rusqlite, I'll work on adapting it to my project so it can query over my own data :) In the meantime, I started adapting my querying modules so they can all return a Vec<Vec<String>>, so this is what I would modify the virtual table module to work with in a first time. I guess a Vec<Vec<String>> virtual table would be a good sample to add in addition to the CSV sample.

gwenn:
@awakecoding Thanks for your feedback. And sorry, but I manage to remove macros (now only `lazy_static` is used) so could you please give another try (when you have time).
Also I remove the `fn vtab(&self) -> &Self::Table;`.
So there will be no `unsafe` code in VTab impls except for the CSVTab.

awakecoding:
@gwenn I tried again and there is less boilerplate code required now, so it is good. However, do you intend to remove the direct dependency on the libsqlite3-sys ffi types? I noticed there are three such types remaining: ffi::sqlite3_vtab, ffi::sqlite3_vtab_cursor, ffi::sqlite3_libversion_number. Ideally, the only dependency should be rusqlite, without bothering about libsqlite3-sys. What do you think?

gwenn:
@awakecoding `ffi::sqlite3_libversion_number` check can be removed by adding `#[cfg(feature = "bundled")]` for example. But I don't see how to remove `ffi::sqlite3_vtab` and `ffi::sqlite3_vtab_cursor`. I will try...

awakecoding:
@gwenn I tried your last commit, and I was able to remove the direct dependency on ffi, thanks! Does this mean this PR could be merged soon enough, now that the ffi dependency has been removed?

PR #131, opened 2016-02-13
Remove load_extension feature from the binding

upsuper:
It doesn't seem to me "load_extension" is used anywhere inside libsqlite3-sys.


jgallagher:
Thanks!

PR #133, opened 2016-02-23
From/ToSql for chrono types.

gwenn:

PR #144, opened 2016-05-16
From/ToSql for chrono types.

jgallagher:
This PR cuts down #133 to the bare necessities:

* Removes support for storing dates and times as ints or floats (only supports strings)
* Prefers to use `chrono`'s RFC3339 functions when possible while still keeping support for SQLite's `datetime()` format

@gwenn This cuts out a lot of code and some significant functionality from #133. Do you think any of the things I cut out are critical to add back in?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,5 @@ script:
     - cargo test --features load_extension
     - cargo test --features trace
     - cargo test --features functions
-    - cargo test --features "backup blob functions load_extension trace"
+    - cargo test --features chrono
+    - cargo test --features "backup blob chrono functions load_extension trace"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ trace = []
 time = "~0.1.0"
 bitflags = "~0.1"
 libc = "~0.2"
+chrono = { version = "~0.2", optional = true }
 
 [dev-dependencies]
 tempdir = "~0.3.4"
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # Version UPCOMING (...)
 
+* Adds support for serializing types from the `chrono` crate. Requires the `chrono` feature.
 * Removes `load_extension` feature from `libsqlite3-sys`. `load_extension` is still available
   on rusqlite itself.
 * Fixes crash on nightly Rust when using the `trace` feature.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,7 @@ build: false
 
 test_script:
   - cargo test --lib --verbose
-  - cargo test --lib --features "backup blob functions load_extension trace"
+  - cargo test --lib --features "backup blob chrono functions load_extension trace"
 
 cache:
   - C:\Users\appveyor\.cargo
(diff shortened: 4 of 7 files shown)


gwenn:
Are you sure you don't want to support `CURRENT_TIMESTAMP` (`YYYY-MM-DD HH:MM:SS`) ?

[date.c](http://www.sqlite.org/cgi/src/artifact/1cc9fb516ec9932c)
```
SQLite processes all times and dates as julian day numbers. The
dates and times are stored as the number of days since noon
in Greenwich on November 24, 4714 B.C. according to the Gregorian
calendar system.
```

[Best way to store only date](http://www.mail-archive.com/sqlite-users%40mailinglists.sqlite.org/msg06239.html)
[Best way to work with dates in Android SQLite](http://stackoverflow.com/questions/7363112/best-way-to-work-with-dates-in-android-sqlite)

It seems to me that supporting only text (especially RFC3339 with the 'T') is too restrictive.

jgallagher:
I left support for that format in place (https://github.com/jgallagher/rusqlite/blob/07801ca3709d6baa92155bb1ff2c12de1d548db7/src/types/chrono.rs#L77-L81). I will add to the tests to ensure it's working correctly.

SQLite also still understands the timestamps with a `T` in the middle (https://www.sqlite.org/lang_datefunc.html, formats 5-7 in the list of "Time Strings"), so I think writing strings with a `T` is fine as long as we understand them coming back without a `T`.

Another change/simplification I'd like to make is to always store dates in UTC; I think that's pretty much always what you want, and if you don't, having to do a little extra work to make sure you're tracking time zones however is appropriate for you application is probably fine. What do you think?

gwenn:
I mean RFC3339 with ' ' (no 'T') does not work anymore for DateTime (https://github.com/gwenn/rusqlite/blob/79376a4ca96f4c645c3fe0aabff12525eff83c0e/src/types/chrono.rs#L184)

[RFC3339](https://www.ietf.org/rfc/rfc3339.txt):
> ISO 8601 defines date and time separated by "T".
> Applications using this syntax may choose, for the sake of
> readability, to specify a full-date and full-time separated by
> (say) a space character.

And no support for database populated by other platforms with julian day or unix time...

jgallagher:
DateTimes with ' ' still work fine (see last commit adding unit tests to confirm). DateTime tries to parse as rfc3339 and if that fails, falls back to the `NaiveDateTime` parser which supports ' ' and 'T'.

Yes, removing support for Julian / Unix was intentional specifically because we can't know the format from just the column type. If a column is a float, is that a Julian day or a Unix timestamp with fractional seconds? If a column is an integer, is that a Unix timestamp or a datetime stored as an integer (e.g., [from the thread you posted earlier](http://www.mail-archive.com/sqlite-users%40mailinglists.sqlite.org/msg06256.html))? Since we can't tell, I think it's safer to just support strings; folks can always provide their own wrappers if they need to import from another system.

jgallagher:
Oh I see, a full RFC3339, including time zone, won't work with a `' '` instead of a `T`, you're right. I'll add that back in.

gwenn:
> specifically because we can't know the format from just the column type.

applies to TEXT too (not only to INTEGER/REAL)...
So maybe it is safer not to support date/time at all :disappointed:

jgallagher:
Hmm. Technically that's true, but I think it's reasonable to support TEXT since there are some common formats (RFC3339/ISO8601, which includes SQLite's `datetime()` default output format). This PR will at least let people write dates/times and read dates/times that have been written by rusqlite itself or `datetime()`.

jgallagher:
Another (key?) difference between supporting TEXT and supporting the number types - the text formats we're supporting here are unambiguous. If someone is storing some other format and tries to parse it using this code, they'll get an error. That may be annoying but isn't the end of the world - they can write their own `FromSql` adapters.

If we support the numeric types, we can't know if we parsed into the wrong type. Users could get a successful parse that contains incorrect data, which is much worse.

PR #148, opened 2016-05-16
Reset as soon as possible

gwenn:


src/convenient.rs
```diff
@@ -24,11 +24,15 @@ impl<'conn> Statement<'conn> {
...
-            None => Ok(false),
-        }
+        let exists = {
+            let mut rows = try!(self.query(params));
+            match rows.next() {
+                Some(_) => Ok(true),
+                None => Ok(false),
+            }
+        };
+        self.reset_if_needed();
```
jgallagher: What's the benefit to resetting here? If the statement is discarded after this call, we didn't need to reset it it. If it's not discarded, it will be reset the next time it's used, right?

gwenn:
If the query returns more that one row, a read lock will be hold until the statement is discarded.
[sqlite3_stmt_busy](http://sqlite.org/c3ref/stmt_busy.html)
> prepared statements that are holding a transaction open

```rust
let mut stmt = db.prepare("SELECT 1 FROM foo WHERE x = ?").unwrap();
loop {
   // ...
   if stmt.exists(...).unwrap() {
      // ...
   }
   // slow tasks
}
```

jgallagher:
Gotcha, ok. Minor stylistic comments - it seems like `exists` should be just a boolean and not a `Result`. We could drop the `return` and put the `Ok` there:

```rust
        let exists = {
            let mut rows = try!(self.query(params));
            match rows.next() {
                Some(_) => true,
                None => false,
            }
        };
        self.reset_if_needed();
        Ok(exists)
}
```

gwenn:
Perfect!

jgallagher:
Now I'm a little worried about `reset_if_needed`'s existence at all. Should we be resetting in `Rows`'s `Drop` impl instead? It feels fragile that we need to do it here (fragile in the sense of "we need it here - do we need it other places too?").

jgallagher:
@gwenn see 4a6c7b5 - does that look correct to you?

gwenn:
I feel the same but currently I have no solution (for example see the `SQLITE_OMIT_AUTORESET` compilation option which is activated on Mac).
Maybe we should create an issue to investigate ?

jgallagher:
I think resetting when a `Rows` is dropped should be correct with or without `SQLITE_OMIT_AUTORESET`:

* A statement can only have one `Rows` "child" active (since it's created from a `&mut self` method)
* `Rows` is only usable via its `Iterator` impl, which returns `None` (or an error) if we get anything except `SQLITE_ROW` from `sqlite3_step`. This should prevent us from hitting any `SQLITE_MISUSE` errors from autoreset (or no autoreset).
* Once the `Rows` is dropped, the statement should be reset so that another query could be performed.

gwenn:
`Drop` seems good.
And why not call `sqlite3_reset` in `next` when an error occurs or `SQLITE_DONE` is returned ?

gwenn:
Currently, what happens when we do ?
```rust
let mut stmt = db.prepare(...).unwrap();
let mut rows = stmt.query().unwrap();
...
let mut other_rows = stmt.query().unwrap();
...
```

jgallagher:
That won't compile. `rows` keeps a "mutable borrow" on `stmt`, so the compiler won't let you call `query()` again until `rows` has been dropped.

jgallagher:
On not calling `sqlite_reset` in `next`:

Most of the time that will reset in a double reset (harmless, but unnecessary?): once when the query gets `SQLITE_DONE` and again when the `Rows` is dropped. We have to keep the reset in `drop` for cases like `exists` where not all rows are fetched.

I think it's reasonable/intuitive that as long as you have a `Rows` alive, the statement is "active". Will SQLite release any locks once it returns `SQLITE_DONE` or an error, or will it still wait for us to reset?

gwenn:
> Will SQLite release any locks once it returns SQLITE_DONE or an error, or will it still wait for us to reset?

I don't know !

jgallagher:
Some brief googling indicates it [might not](http://sqlite.1065341.n5.nabble.com/Re-Prepared-statements-and-locking-td58354.html) and we should call reset ASAP. I guess we can clear out the Rows's statement when we reset it to avoid double-resetting.

PR #152, opened 2016-05-18
Savepoint improvements

### LINKED ISSUE #142: Savepoints with same name
Currently, all the created savepoints have the same name `sp`.
See the following test case:
https://github.com/gwenn/rusqlite/commit/7c5f74447a2a24e322ed0ab41989086ef5229a6e#diff-d0e9b1d1df1c5795eac22a324e40477eR265
for an example that unexpectedly fails:
https://ci.appveyor.com/project/gwenn/rusqlite/build/1.0.64
It seems to me that savepoints must have distinct names.
Regards.

jgallagher:
Closes #142.

We still keep the requirement to nest transactions/savepoints, but users can provide their own savepoint names and savepoints can now be rolled back without consuming them. (Transactions are still consumed on rollback just like they are in SQLite.)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,11 @@
   connection are inherently nested. While a transaction is alive, the parent connection or
   transaction is unusable, so `Transaction` now implements `Deref<Target=Connection>`, giving
   access to `Connection`'s methods via the `Transaction` itself.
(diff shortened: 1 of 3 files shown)


gwenn:
:thumbsup:

PR #161, opened 2016-05-25
Redesign FromSql

### LINKED ISSUE #160: Fix handling of SQLite column types
Experimenting with possibilities brought up on #158, I realized we're leaving open the possibility of using `sqlite3_column_type` incorrect. From [SQLite's docs](https://www.sqlite.org/c3ref/column_blob.html):

> The value returned by sqlite3_column_type() is only meaningful if no type conversions have occurred as described below. After a type conversion, the value returned by sqlite3_column_type() is undefined. Future versions of SQLite may change the behavior of sqlite3_column_type() following a type conversion.

We call `sqlite3_column_type` several places, including many `column_has_valid_sqlite_type` implementations and the `FromSql` impls for `Option<T>` and `Value`. If the user performs an operation that requires SQLite type conversion prior to these being called, we're hitting undefined behavior.

At first I thought maybe we could avoid calling `sqlite3_column_type` altogether, and allow `get`s that hit incorrectly-typed data to fall back onto SQLite's type conversions (which are well defined for all of SQLite's types). That could get rid of `column_has_valid_sqlite_type`, but we still need a way to implement `Option<T>` and `Value`.

The only way I can think of to do this correctly is to call `sqlite3_column_type` ahead of time for each column. This could be cached in the construction of the `Rows` handle, but not up in `Statement` (since a DDL execution between statement queries could result in changes).

@gwenn Any thoughts on this? Is my understanding of this issue correct, and if so, can you think of any other solution? I checked a couple of other SQLite wrappers; go-sqlite3 pulls out all the column values ahead of time during iteration (it calls `sqlite3_column_type` but always as the first operation for a column), and [sqlite-jdbc](https://bitbucket.org/xerial/sqlite-jdbc/src/8de652d3f54cdfd32d5a493130c995f36e16c652/src/main/java/org/sqlite/RS.java?at=default&fileviewer=file-view-default) appears to have the same issue we currently do.

jgallagher:
Closes #160.

`FromSql` can now be implemented without `unsafe`. It also allows us to remove `functions::FromValue` (which was largely a duplicate of `FromSql`), since we can construct a `BorrowedValue` from a `sqlite3_value`, therefore allowing us to reuse `FromSql` types.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+# Version UPCOMING (...)
+
+* BREAKING CHANGE: The `FromSql` trait has been redesigned. It now requires a single, safe
+  method instead of the previous definition which required implementing one or two unsafe
+  methods.
+
 # Version 0.7.2 (2016-05-19)
 
 * BREAKING CHANGE: `Rows` no longer implements `Iterator`. It still has a `next()` method, but
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -54,7 +54,6 @@ use std::ffi::CStr;
 use std::mem;
 use std::ptr;
 use std::slice;
-use std::str;
 use libc::{c_int, c_double, c_char, c_void};
 
 use ffi;
@@ -63,7 +62,7 @@ pub use ffi::sqlite3_value;
 pub use ffi::sqlite3_value_type;
 pub use ffi::sqlite3_value_numeric_type;
 
-use types::Null;
+use types::{Null, FromSql, BorrowedValue};
 
 use {Result, Error, Connection, str_to_cstring, InnerConnection};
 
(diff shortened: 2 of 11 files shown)


gwenn:
May be:
```
diff --git a/src/types/mod.rs b/src/types/mod.rs
index 7853291..83192ed 100644
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -51,7 +51,6 @@
 //! `None`).

 pub use ffi::sqlite3_stmt;
-pub use ffi::sqlite3_column_type;
 pub use ffi::{SQLITE_INTEGER, SQLITE_FLOAT, SQLITE_TEXT, SQLITE_BLOB, SQLITE_NULL};

 pub use self::from_sql::FromSql;
```

jgallagher:
Good catch. Should be able to drop the next line too, right?

gwenn:
Right.

PR #166, opened 2016-05-26
Better error reporting on invalid column type.

gwenn:


jgallagher:
I love the additional reporting, but don't like having to pass the column index to `FromSql` implementors just for filling in the error. A couple of ideas:

* Add another `Error` case for "invalid column type" that has no attached information, the `map_err` instances of that case into the appropriate full-information one inside of `Row::get_checked` and `Context::get`. This is the simplest solution but leaves us with a dangling `Error` case that we won't really expect to ever return out of the library.
* Change the return type of `FromSql`'s error to allow specifying an invalid type. E.g.,

```rust
pub enum FromSqlError {
    case InvalidColumnType
    case Error(::error::Error)
}

pub trait FromSql: Sized {
    fn column_result(value: ValueRef) -> ::Result<Self, FromSqlError>
}
```

which we'd `map_err` to convert to the appropriate `Error` case.

What do you think?

gwenn:
Ok,
I try to fix it by introducing an intermediary InvalidType error...
But now, FromSql implementors cannot return custom errors with the index specified...
Maybe we should wrap all errors (not only InvalidType) ?

jgallagher:
Hmm. The only `Error` cases that are useful for `FromSql` implementors are `InvalidType` and `FromSqlConversionFailure`, right? Should we pull those two out into their own enum and have `column_result` return that so it's crystal clear what errors `FromSql` can produce? We could then keep the column index you've added to `InvalidColumnType` and `InvalidFunctionParameterType` and add one to `FromSqlConversionFailure`.

jgallagher:
Oh, one other thing. I don't think we need `InvalidType` to carry its `(Type)` with it, since we can add that in `get_checked` the same way we add the column index. We just need implementors to tell us the value we gave them didn't have an acceptable type.

gwenn:
Done!

jgallagher:
Nice. What do you think about not having a column index on `FromSqlConversionError`? I'm leaning toward a different error type for `FromSql` implementors, but not strongly.

gwenn:
Maybe we can just wrap all errors returned by `FromSql::column_result` into a `(col, type, cause)` error ?

jgallagher:
What would the type signature of `FromSql` be?

gwenn:
The same as it is today.

jgallagher:
How do you propose doing the wrapping into `(col, type, cause)`?

gwenn:
Something like that:
```rust
diff --git a/src/error.rs b/src/error.rs
index 9274217..b32264e 100644
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,9 +49,9 @@ pub enum Error {
     /// for the statement.
     InvalidColumnName(String),

-    /// Error when the value of a particular column is requested, but the type of the result in
-    /// that column cannot be converted to the requested Rust type.
-    InvalidColumnType(i32, Type),
+    /// Error when the value of a particular column is requested, it cannot be converted to the
+    /// requested Rust type.
+    InvalidColumnValue(i32, Type, Box<Error>),

     /// Error when an SQLite value is requested, but the type of the result cannot be converted to the
     /// requested Rust type.
@@ -67,7 +67,7 @@ pub enum Error {
     /// Error returned by `functions::Context::get` when the function argument cannot be converted
     /// to the requested type.
     #[cfg(feature = "functions")]
-    InvalidFunctionParameterType(usize, Type),
+    InvalidFunctionParameterValue(usize, Type, Box<Error>),

     /// An error case available for implementors of custom user functions (e.g.,
     /// `create_scalar_function`).
diff --git a/src/lib.rs b/src/lib.rs
index 35cfca7..0ee43a7 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1055,8 +1055,8 @@ impl<'a, 'stmt> Row<'a, 'stmt> {
     ///
     /// ## Failure
     ///
-    /// Returns an `Error::InvalidColumnType` if the underlying SQLite column
-    /// type is not a valid type as a source for `T`.
+    /// Returns an `Error::InvalidColumnValue` if the underlying SQLite column
+    /// value is not a valid value as a source for `T`.
     ///
     /// Returns an `Error::InvalidColumnIndex` if `idx` is outside the valid column range
     /// for this row.
@@ -1066,10 +1066,7 @@ impl<'a, 'stmt> Row<'a, 'stmt> {
     pub fn get_checked<I: RowIndex, T: FromSql>(&self, idx: I) -> Result<T> {
         let idx = try!(idx.idx(self.stmt));
         let value = unsafe { ValueRef::new(&self.stmt.stmt, idx) };
-        FromSql::column_result(value).map_err(|err| match err {
-            Error::InvalidType => Error::InvalidColumnType(idx, value.data_type()),
-            _ => err,
-        })
+        FromSql::column_result(value).map_err(|err| Error::InvalidColumnValue(idx, value.data_type(), Box::new(err)))
     }

     /// Return the number of columns in the current row.
```

jgallagher:
I don't think that's ideal for a couple of reasons:

* It requires boxing the error in all cases, which seems unnecessary when the most common error by far (presumably) is an invalid type conversion.
* The current signature of `FromSql` means implementors have to return a `rusqlite::Error`, but there aren't really meaningful error cases within `Error` (at least with the above changes).

What about:

```rust
pub enum FromSqlError {
    InvalidType,
    Other(Box<Error + Send + Sync>),
}

pub trait FromSql: Sized {
    fn column_result(value: ValueRef) -> ::std::result::Result<Self, FromSqlError>
}

// inside of our existing Error enum, keep InvalidColumnType and InvalidFunctionParameterType,
// and modify FromSqlConversionFailure to include the column index and column type.

// modify get_checked into roughly:
pub fn get_checked<I: RowIndex, T: FromSql>(&self, idx: I) -> Result<T> {
     let idx = try!(idx.idx(self.stmt));
     let value = unsafe { ValueRef::new(&self.stmt.stmt, idx) };
     FromSql::column_result(value).map_err(|err| match err {
        FromSqlError::InvalidType => Error::InvalidColumnType(idx, value.data_type()),
        FromSqlError::Other(err) => Error::FromSqlConversionFailure(idx, value.data_type(), err),
    })
}
```

gwenn:
Ok.
And `FromSqlConversionFailure` is not useful anymore, right ?

jgallagher:
It is in case someone wants to return an error _other_ than `InvalidType` (like the conversion errors we return from time/chrono types).

gwenn:
Ok.
`FromSqlConversionFailure` is not used directly anymore in `FromSql` implementations but only in `get_checked`.

jgallagher:
Yeah, exactly.

Issue #89, opened 2015-12-01
Consider building SQLite as part of the build process if we can't find it on the system.


jgallagher:
This is what [libgit2-sys](http://doc.crates.io/build-script.html#case-study-linking-to-system-libraries) does, and was also mentioned as a possible solution to difficulties on Windows in #46.

bottiger:
I would love to see this implemented in a not too distant future.

photex:
I just submitted a PR for sqlite3-src which does this and works on the big 3 at least: https://github.com/stainless-steel/sqlite3-src/pull/1

I'd be happy to submit something similar here after that PR is finished.

jgallagher:
This is now available as the `bundled` feature (see #176 / #197).

PR #176, opened 2016-06-15
Added feature "bundled" to provide a fully self-contained cross-platform build.

photex:
Howdy!

In reference to #46 and #89 I'm submitting this PR which adds a feature "bundled" that will build and link with a static libsqlite3 from the included amalgamation.

If you don't want the amalgamation bundled then I'm happy to toss that and come up with another solution too.


chamakits:
@jgallagher is this something you would be interested in merging in?

I'm asking because it is by far the simplest way to allow sqlite to be used with Rust in windows, and I'd love to see this make it in; and keeping it behind the bundled feature seems like a neat way to allow for this.

I ask this not trivializing the work it'd be to maintain, and totally understanding if you'd like to keep this complexity out. But the work here done by @photex seems pretty great, and would likely allow Windows users (and other platform as well) get up and running with Rust and sqlite insanely quickly.  I'd be willing to help maintain if either of you need the help.

gwenn:
Maybe you can regenerate the `bindgen.rs` file with the bundled headers ?
```sh
/path/to/bindgen -builtins -l sqlite3 -o /path/to/libsqlite3-sys/src/bindgen.rs /path/to/sqlite3.h
# or /path/to/bindgen -builtins -l sqlite3 -o /path/to/libsqlite3-sys/src/bindgen.rs /path/to/sqlite3ext.h
```
Regards

chamakits:
Hi @gwenn , couple of questions.

Are you providing that as a recommendation of an alternative to bundling sqlite3 but allowing to use it windows, or as a necessary step to include this in the PR?

Either way, I tried multiple ways to get this working on windows, and was not able to.  It may entirely be because I'm still pretty new to Rust and it's tool chain.

These are the things I tried:

- Tried to install with 'cargo install bindgen' which failed at first.  Then I installed LLVM, and had to manually place both the libclang.dll and the libclang.lib in the same directory, to be able to indicate the the directory in which they are located with LIBCLANG_PATH.  This allowed bindgen to install, but running it regardless of parameters would exit out without printing anything.

- I downloaded bindgen repo, and built from source.  Building from source still had the same issue.  Saw the source code and was certain from it's use of docopt that it should print out the flag options.  Added a println to make sure there was a problem, and same, early exit, no output.

- Able to run with 'cargo run' and it finally gave useful text output.  Tried what you mentioned by sending parameters after 'cargo run --'  but it was requesting libclang.a, which my LLVM install did not contain.  Tried also an available snapshot installation, and also no luck.  Have searched online to see where the libclang.a file should be, and no such luck.

If this was meant as an alternative to bundling sqlite3, then it seems it might be a good deal more complicated than bundling.  I don't doubt that the failure is entirely on my unfamiliarity with the tool chain, but surely seems a good deal more complicated than bundling if bundling is still a viable option.

If you meant that we need to provide the bindings to be able to merge this in, could you point me to some more reading where I could find out how to do this?  Could you also provide some context as to why this would be needed as part of this merge?

Thanks in advance!

gwenn:
@chamakits It is not mandatory and may have negative impact for code using an old sqlite3 library provided by the system.
It is just a suggestion to have consistent `bindgen.rs` and headers.
May I suggest that you use a linux VM to generate the `bindgen.rs` file.

chamakits:
Ah I see, that makes sense.

Okay, will try on Linux.  Thanks!

chamakits:
@gwenn that was RIDICULOUSLY easy on Linux.  Should have tried that before.

Created a PR for @photex 's master branch to include the bindgen changes and some other minor changes.  Can be seen here:
https://github.com/photex/rusqlite/pull/1

Let me know if either see any issues with it, or if I should change anything.

gwenn:
I don't understand why the `shell.c` is bundled !

chamakits:
Yeah, not sure, @photex might have added it by accident.  @photex , if you don't mind, I'll remove it in my PR as well.

@gwenn anything else we should shave off?

photex:
Yeah I didn't add that intentionally. Merged your PR @chamakits

chamakits:
Awesome.  Let me know if there are any other changes that we should make for this. Looking forward to seeing this make it in.  Thank you both @gwenn and @photex .

gwenn:
@chamakits Maybe the [compilation options](http://sqlite.org/compile.html) ?
 * SQLITE_THREADSAFE
 * SQLITE_USE_URI
 * SQLITE_ENABLE_COLUMN_METADATA
 * SQLITE_ENABLE_FTS5
 * SQLITE_ENABLE_JSON1
 * SQLITE_ENABLE_STAT4
 * SQLITE_ENABLE_DBSTAT_VTAB
 * SQLITE_DEFAULT_FOREIGN_KEYS

?
This list is not exhaustive.

chamakits:
Sure, I can look into that.  Are you saying that these should be used by default, or that they should be configurable as part of the features, or that we should investigate which ones make sense?

chamakits:
@photex @gwenn 
I've begun adding flags, specifically the ones @gwenn mentioned.  The ones that had parameters I just chose what sounded like a 'nice to have', but not sure if that's the right choice or not.  See commit here:
https://github.com/chamakits/rusqlite/commit/ef6390d93210009b08113a4a1f4abd369f4c7af8

Any guidance as to which flags we would want to include? I'm seeing there are 28 flags, do we want to include them all? A good deal of them have parameters as well, any guidance as to how to choose which parameter in the case that they do?

gwenn:
No particular guidance but take a look at compile options used by:
* [go-sqlite3](https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L10)
* [sqlite-jdbc](https://github.com/xerial/sqlite-jdbc/blob/master/Makefile#L61)
* [System.Data.SQLite](http://system.data.sqlite.org/):
```
SQLITE_THREADSAFE=1;SQLITE_USE_URI=1;SQLITE_ENABLE_COLUMN_METADATA=1;SQLITE_ENABLE_STAT4=1;SQLITE_ENABLE_FTS3=1;SQLITE_ENABLE_LOAD_EXTENSION=1;SQLITE_ENABLE_RTREE=1;SQLITE_SOUNDEX=1;SQLITE_ENABLE_MEMORY_MANAGEMENT=1;SQLITE_ENABLE_API_ARMOR=1;SQLITE_ENABLE_DBSTAT_VTAB=1
```
* [sqlite3-ruby](https://github.com/sparklemotion/sqlite3-ruby/blob/master/tasks/vendor_sqlite3.rake#L19)

chamakits:
Okay, will take a look at those, and come up with some flags based on that.

chamakits:
@gwenn and @photex I've added some flags based on those links provided and from what I've seen.  See them here:

https://github.com/chamakits/rusqlite/blob/feature/bundle-update-sqliteflags/libsqlite3-sys/build.rs#L35

Do you both think these make sense? Anything missing, or anything that shouldn't go in?  If it looks good, I'll open another PR for @photex

jgallagher:
Flags look good to me. Incorporated them into #197.

Issue #179, opened 2016-07-10
query_row for Statement and CachedStatement


est31:
It would be cool to have the query_row method for Statement and CachedStatement as well.

jgallagher:
Fixed in #195

est31:
Thanks!

PR #180, opened 2016-07-14
Add Statement.query_row convenient method (#179)

gwenn:


jgallagher:
Merged as part of #195

PR #182, opened 2016-08-29
Update doc links to docs.rs

gwenn:


Cargo.toml
```diff
@@ -13,6 +13,7 @@ license = "MIT"
 name = "rusqlite"
 
 [features]
+default = ["backup", "blob", "chrono", "functions", "load_extension", "serde_json", "trace"]
```
jgallagher: I don't think I want to enable all features by default just to get docs.rs to show everything. Let's table this until onur/docs.rs#29 is addressed (so we can enable these just for the docs and not for all users).

gwenn:
[Docs.rs now supports building non-default features](https://github.com/onur/docs.rs/pull/131)

PR #200, opened 2017-01-04
Make Connection::close() return the connection on failure.

### LINKED ISSUE #199: Can't close database properly when database is locked.
Because `Connection::close` consumes the connection, it's impossible to retry when closing yields an error. I've encountered an "Error code 5: The database file is locked" error when closing. It seems that the proper thing to do is to sleep for some time and then try again, but I can't do this because I no longer own the connection or hold a reference to it.

Very possible I'm misunderstanding something here, but from what I can understand it seems there's something wrong here.

jgallagher:
Closes #199.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+# Version UPCOMING (TBD)
+
+* BREAKING CHANGE: `Connection::close()` now returns a `Result<(), (Connection, Error)>` instead
+  of a `Result<(), Error>` so callers get the still-open connection back on failure.
+
 # Version 0.8.0 (2016-12-31)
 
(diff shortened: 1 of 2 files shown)


gwenn:
It seems to be that there is something missing:
The raw pointer must not be set to null when closing the inner connection:
https://github.com/jgallagher/rusqlite/blob/master/src/lib.rs#L611
```
    fn close(&mut self) -> Result<()> {
        unsafe {
            let r = ffi::sqlite3_close(self.db());
            self.db = ptr::null_mut(); // do this only when r is SQLITE_OK
            self.decode_result(r)
        }
    }

```
Currently, you cannot really retry (passing null to `sqlite3_close` is a NOP).

jgallagher:
Great catch. Will fix shortly.

PR #203, opened 2017-01-06
Silence warnings about deprecations when exporting deprecated types.

jgallagher:


gwenn:
FYI,
There is another one when activating the `load_extension` feature:
```
warning: use of deprecated item: Use LoadExtensionGuard instead, #[warn(deprecated)] on by default
  --> src/lib.rs:93:32
93 | pub use load_extension_guard::{SqliteLoadExtensionGuard, LoadExtensionGuard};
```

jgallagher:
Thanks, fixed in #204

PR #215, opened 2017-01-25
Fix doc comment example of FromSql/ToSql

reddraggone9:
I modeled the changes to the example code after what I did when migrating my own code. Afterwords however, I noticed you used the more explicit (though more verbose) `ToSqlOutput::from` in types/to_sql.rs. Let me know if you'd like me to change the example to use that form.

While I'm at it, thanks for making this! Obviously I've found it useful 😀


jgallagher:
Thanks! `.into()` is fine; I probably should've used that in `types/to_sql.rs`. Always forget it's there.

Issue #220, opened 2017-02-03
Expose limits, such as SQLITE_MAX_VARIABLE_NUMBER


rnewman:
SQLite includes a bunch of compile-time limits that are the backbone of the runtime limits:

https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber

The runtime limits can be retrieved via `sqlite3_limit` (though I don't see the constants passed through in the bindings?).

The compile-time limits are [nowhere to be seen](https://github.com/jgallagher/rusqlite/search?utf8=%E2%9C%93&q=SQLITE_MAX_VARIABLE_NUMBER). Can we expose a wrapper for these? Apparently the values [vary widely, from 999 as the source default to 500,000 on macOS](http://sqlite.1065341.n5.nabble.com/SQLITE-MAX-VARIABLE-NUMBER-td43457.html).

These limits are commonly used to implement chunked passing of a large number of parameters to a query, where the number of parameters might be more than 999.

jgallagher:
Just to clarify - by "compile-time limits", you mean the `#define`'d constants that you pass to `sqlite3_limit`, right? Does #221 address this?

rnewman:
#221 supplies the constants and wrapper function to do dynamic limit checking — fetching the current variable limit with `sqlite3_limit(SQLITE_LIMIT_VARIABLE_NUMBER)`, for example.

Unless the limit has otherwise been changed, that function call will return the value of the compile time preprocessor macro `SQLITE_MAX_VARIABLE_NUMBER`. If it has been changed, `sqlite3_limit` will return the current value.

Each limit cannot be increased above that defined constant (`SQLITE_MAX_VARIABLE_NUMBER` in this particular case), so knowing the constant itself can be useful.

Ideally those maxes themselves would also be available to Rust code, just as they are when compiling against `sqlite3.h`. #221 doesn't address that, but it's certainly much better than nothing!

jgallagher:
Are those available in `sqlite3.h`? I'm not seeing anything related to them in mine (3.16.2):

```
[2713 john@Johns-MacBook-Pro sqlite/3.16.2/include]% ls
sqlite3.h     sqlite3ext.h
[2714 john@Johns-MacBook-Pro sqlite/3.16.2/include]% grep VARIABLE *
sqlite3.h:** [[SQLITE_LIMIT_VARIABLE_NUMBER]]
[...]

PR #221, opened 2017-02-04
Expose limits (#220)

gwenn:


src/limits.rs
```diff
@@ -0,0 +1,70 @@
...
+use libc::c_int;
+
+use ffi;
+pub use ffi::Limit;
+
+use Connection;
+
+impl Connection {
+    /// Returns the current value of a limit.
+    pub fn get_limit(&self, limit: Limit) -> i32 {
```
jgallagher: Total nit, but should this be `pub fn limit(...)` instead of `get_limit`? This isn't quite what [the conventions RFC](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis) is talking about with getters/setters (since we take a `Limit` arg) but it's close.

src/limits.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[cfg(test)]
+mod test {
+    use ffi::Limit;
+    use Connection;
+
+    #[test]
+    fn test_limit() {
+        let db = Connection::open_in_memory().unwrap();
+        assert_eq!(2147483645, db.set_limit(Limit::SQLITE_LIMIT_LENGTH, 1024));
```
jgallagher: I'm not sure these tests are universally applicable; folks might be linking against a system-provided SQLite that set different default limits. Maybe all of these should set first (unchecked) then confirm the get returns the same value? Would be happier to keep these checks but am concerned about failing tests incorrectly; not sure if there's a middle ground.

jgallagher:
It looks like the travis failure is legit:

```
failures:
---- limits::test::test_limit stdout ----
	thread 'limits::test::test_limit' panicked at 'assertion failed: `(left == right)` (left: `2`, right: `-1`)', src/limits.rs:63
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

`src/limits.rs:63` is

```rust
assert_eq!(2, db.limit(Limit::SQLITE_LIMIT_WORKER_THREADS));
```

According to [SQLite's release history](https://www.sqlite.org/changes.html), SQLITE_LIMIT_WORKER_THREADS was added in 3.8.7. If the SQLite version on travis is older than that, it would explain getting -1 back. Maybe we need a check like we have [in src/lib.rs](https://github.com/jgallagher/rusqlite/blob/53b85d2ab1162eafa88dfe2389234009a83844c1/src/lib.rs#L1513-L1515)?

This does bring up two more issues:

* We should expose the SQLite version
* What should we do if we're running with a SQLite version older than our bindgen run?

I'll open separate issues to discuss those. Just adding the version guard is probably fine for this PR.

PR #230, opened 2017-02-08
rustfmt v0.7.1

gwenn:


src/lib.rs
```diff
@@ -417,10 +423,10 @@ impl Connection {
...
         self.flush_prepared_statement_cache();
         {
-            let mut db = self.db.borrow_mut();
-            db.close()
-        }
-        .map_err(move |err| (self, err))
+                let mut db = self.db.borrow_mut();
+                db.close()
+            }
+            .map_err(move |err| (self, err))
```
jgallagher: This is the only one that seems a little off. Will merge and then see if I can finagle this to format more cleanly.

jgallagher:
👍 Thanks

Issue #225, opened 2017-02-07
What should we do when running with a SQLite version older than the one we ran bindgen against?


jgallagher:
This came up on #221 where `SQLITE_LIMIT_WORKER_THREADS` is exposed by rusqlite but unknown if we run linked against a SQLite older than 3.8.7. I'm not sure how we should address this. Options:

* Do nothing (current "strategy")
* Do nothing but attempt to document any exposed features newer than X (some SQLite version that predates what we think the vast majority of platforms ship with)
* Throw an error (when opening new connections?) if the SQLite version is older than X. X here would almost certainly have to be something older than the version we bundle with and run bindgen against - otherwise no one would be able to use their platform-provided libsqlite; this still leaves the problem of what to do if we expose or expect something between X and our bundled version.
* Add SQLite version checks to any rusqlite functions that use features newer than SQLite version X, and return errors from those functions if they're called. E.g. for #221, we could change `limit()`/`set_limit()` to return `Result<i32>` and fail with something like a `SqliteVersionOutdated` error if the caller tried to use `SQLITE_LIMIT_WORKER_THREADS` with a SQLite older than 3.8.7.

The last one seems like the best but also the most work (by far). @gwenn or @rnewman any thoughts?

rnewman:
Random thoughts:

This is a little tricky, because all of these avenues allow for runtime failures (by definition). As such, I think documentation must be at least part of the solution: a developer testing against new SQLite won't find bugs no matter what you do. A developer who _really_ wants to avoid risk will statically link.

Furthermore, a developer expecting to use a system-provided library must be prepared for it to be old, so I don't think it's worth you jumping through too many hoops. For example, Firefox for iOS relies on partial index support, [and checks for it](https://github.com/mozilla-mobile/firefox-ios/blob/master/Storage/SQL/BrowserTable.swift#L122).

I think it's only appropriate for rusqlite to itself handle errors or document individual feature incompatibilities if it's exposing functionality that is more than just a passthrough.

SQLite 3.9.0 and later use semver, at least, so you could robustly check for incompatibilities in newer versions.

3.8.7 is more than two years old. However, Android versions are renowned for shipping old SQLite: Firefox for Android, for example, runs on devices that ship 3.7.x. That means you can't ignore the problem completely.

My proposal:
- Rely on SQLite's own backward-compatibility efforts. As I understand it, #221 simply retrieves `-1` because SQLite doesn't understand a limit — it doesn't fail. That's OK.
[...]

PR #232, opened 2017-02-09
Run bindgen at build time

jgallagher:
This PR changes the build process of `libsqlite3-sys` substantially - it now runs [rust-bindgen](https://github.com/servo/rust-bindgen) as a part of the build process instead of shipping a prebuilt bindgen output as part of the crate. This has pros and cons, including at least:

Pros:
* Greatly reduces the risk discussed on #225 where a user could link against a SQLite version older than the one we ran bindgen against.
* I think we could add a panicking version check if the runtime SQLite version is lower than the buildtime SQLite version? I'm less certain whether this is reasonable for a platform like Android.

Cons:
* Building rusqlite would now require being able to run rust-bindgen, which requires a relatively modern LLVM/Clang. This was slightly annoying to get working on Travis CI, for example.
* We have to be able to find `sqlite3.h` to build. This doesn't apply to `bundled` (since we ship one) and is probably a non-issue on Linux/Mac but is slightly annoying on Windows because the prebuilt Windows zip from SQLite does not include a header. (See changes to appveyor.yml - I downloaded the source zip too just to grab the header. Not sure if there's a better way.)
* ~~Build fails if the SQLite header we find is missing constants/functions/etc rusqlite is expecting. This is mostly a pro (and is mentioned above), but not always. The `functions` feature makes use of the `SQLITE_DETERMINISTIC` flag (added in SQLite 3.8.3), so builds will fail against SQLite 3.7.x, for example. However, setting this flag in older SQLite versions is harmless and doesn't change behavior (it allows for more SQLite optimizations); `functions` _should_ be available for SQLite >= 3.7.3 (the version that added `sqlite3_create_function_v2`). Maybe we can restore support for this by including the `SQLITE_DETERMINISTIC` flag in `libsqlite3-sys` even if it's not present in the header file we run bindgen against? I'll fiddle with this.~~ Update: See 5dbfa28 - sidesteps this problem.

@rnewman and @gwenn I'd love to get your feedback on this direction.


gwenn:
The fact that all the SQLite constants are available is great.

jgallagher:
Yeah, I think the pros outweigh the cons. If being able to build without being able to run bindgen is a serious issue for anyone, we may be able to special case support for that (e.g., point an environment variable at some prebuilt bindings or something). Merging this.

PR #236, opened 2017-02-11
Fix URL Markdown syntax in Connection docs

miedzinski:
Fix malformed Markdown URL syntax in both ``Connection::open_with_flags`` and ``Connection::open_in_memory_with_flags`` docs.


jgallagher:
Thanks!

PR #237, opened 2017-02-16
Remove the dependency on `libc`

sgrif:
Recent versions of bindgen use `std::os::raw` over `libc`, but currently
`libsqlite3-sys` is overriding that. `std::os::raw` is a subset of
`libc` that exports only the relevant type definitions, but not any
functions which require additional linking. This enables
`libsqlite3-sys` to be more easily used on targets that may not have a
libc available (presumably sqlite itself would have been compiled with
musl in that case)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ limits = []
 time = "0.1.0"
 bitflags = "0.7"
 lru-cache = "0.1.0"
-libc = "0.2"
 chrono = { version = "0.3", optional = true }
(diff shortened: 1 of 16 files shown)


jgallagher:
It seems like this is an ongoing point of discussion ([background from internals](https://internals.rust-lang.org/t/solve-std-os-raw-c-void/3268), [open RFC](https://github.com/rust-lang/rfcs/pull/1783)). I don't _think_ rusqlite really has a dog in this fight. I'm fine with this change 👍

PR #257, opened 2017-04-06
Add bindgen bindings for SQLite 3.7.16 to libsqlite3-sys.

jgallagher:
Bump to version 0.10.3 to use libsqlite3-sys 0.8.0.


SergioBenitez:
This is a breaking change and should not have been made without bumping the minor version, in this case. This is because only one version of `libsqlite3-sys` can be linked at once as mentioned in https://github.com/diesel-rs/diesel/issues/844#issuecomment-292257703. I'd propose a yank of this version and a new `0.11`.

jgallagher:
Done 👍

PR #263, opened 2017-04-18
Set HAVE_USLEEP

sfackler:
SQLite otherwise has to sleep for a second at a time when waiting for a
lock (!)

usleep is truly ancient so I don't think there will be platforms that don't actually have it.

There are some other flags like this: https://sqlite.org/compile.html. Might be worth doing some configure-style detection of them?


jgallagher:
Thanks! Do you know if there's an idiomatic way to do that kind of build time detection of platform feature availability?

sfackler:
One approach would be to grab the "full" SQLite dist and use its configure script but that would have portability issues on Windows.

jgallagher:
Yeah, I was hoping to avoid that. :(

Issue #260, opened 2017-04-15
Implementation of sqlite3_update_hook


mqudsi:
I'm posting this because of the comment in the code requesting that issues be created for use cases requiring access to the underlying sqlite3 handle.

I do not see an implementation for `sqlite3_update_hook`. It is exposed in sqlite3_sys as follows:

    pub unsafe extern fn sqlite3_update_hook(p: *mut sqlite3,
                                         f: Option<sqlite3_update_hook_callback>,
                                         p: *mut c_void)
                                         -> *mut c_void

We have a unique-ish use case where integration with a system is performed by opening the sqlite3 db for read while the main module writes to the database. Update notifications are required to trigger a re-scan of rows in question. I don't think this functionality is currently implemented in rusqlite; I don't know whether it would be something you'd be interested in implementing.

Thank you.

gwenn:
rusqlite 0.14 released

mqudsi:
Thanks!

PR #266, opened 2017-04-25
Implementation of sqlite3_update_hook #260

gwenn:
First draft (not tested and a memory leak)


gwenn:
It seems to work only with the `bundled` SQLite version !

PR #267, opened 2017-05-02
Add example for enabling bundled features in Readme

THeK3nger:
I'm quite new to Rust, especially with using `*-sys` crates. Yesterday I spent way more time than necessary trying to link and compile rusqlite on Windows. I knew that the "bundled" features was the way to go, but I was not able to find how to enable it (as I said, noob here!)

I think that adding this 3 lines snippet in the `README` can help new rustaceans! Especially because the `bundled` feature is the only real reliable way to use rusqlite on Windows. :)


jgallagher:
👍 Thanks!

PR #270, opened 2017-05-08
Use the linguist-vendored attribute to vendor sqlite3 sources

gwenn:


jgallagher:
Neat, TIL about [github/linguist](https://github.com/github/linguist). Thanks!

Issue #273, opened 2017-05-11
Best way to use pragmas?


dgrnbrg:
Hi, I'm trying to use pragmas in my SQLite interactions. I am invoking

```rust
try!(conn.execute("PRAGMA synchronous = OFF;", &[]));
try!(conn.execute("PRAGMA journal_mode = OFF;", &[]));
```

However, these lines fail with the error: `Error: Execute returned results - did you mean to call query?`. Should I be calling `query`? If so, I can't figure out from sqlite's docs what the return value is that I should be expecting.

Thank you for your help in using this library correctly!

gwenn:
Just try with the SQLite shell:
```sh
$ sqlite3
sqlite> PRAGMA synchronous = OFF;
sqlite> -- nothing is returned, so it should be okay to use `conn.execute`
sqlite> PRAGMA journal_mode = OFF;
off
sqlite> -- activated mode is returned, so `conn.execute_batch` should be used instead if you don't care of the result.
```

Or read the doc thoroughly :smile:
http://sqlite.org/pragma.html#pragma_journal_mode
> The new journal mode is returned.

Regards.

dgrnbrg:
You are so right, thank you! For future readers of this issue, here is my implementation which casts any errors into the appropriate application-level error type without panicking:

```rust
try!(conn.execute("PRAGMA synchronous = OFF;", &[]));
try!(conn.query_row_and_then("PRAGMA journal_mode = OFF;", &[], |row| -> Result<()> {
    row.get_checked::<_,String>(0)? == "off".to_owned();
    Ok(())
}));
```

xpe:
Good question and self-answer here -- thanks.

For others that find their way here, I recommend considering `PRAGMA journal_mode = WAL;`. Read more about it at:
* https://sqlite.org/wal.html
* https://sqlite.org/pragma.html#pragma_journal_mode

gwenn:
A more appropriate answer is https://docs.rs/rusqlite/0.25.3/rusqlite/?search=pragma

Issue #265, opened 2017-04-23
Feature request: A method to easily access and set the user_version pragma


porglezomp:
I'm working on doing the schema versioning for my database, and I need to use `user_version`. It seems like it would be nice to have an extremely easy to get at it.

gwenn:
Do you mean:
```rust
let user_version: i64 = db.query_row("PRAGMA user_version", &[], |row| row.get(0)).unwrap();
db.execute(&format!("PRAGMA user_version = {}", 1), &[]).unwrap();
```
are too verbose ?

porglezomp:
That's what I'm using right now. I hadn't used it long enough at that point to get the philosophy of this library, I can see now that having lots of specialized methods isn't really the style here.

gwenn:
rusqlite 0.17 released

PR #274, opened 2017-05-18
Draft to ease PRAGMA usage (#273 and #265)

gwenn:


src/pragma.rs
```diff
@@ -0,0 +1,376 @@
...
+        if let Some(schema_name) = schema_name {
+            self.push_schema_name(schema_name);
+            self.push_dot();
+        }
+        self.push_keyword(pragma_name)
+
+    }
+
+    pub fn push_keyword(&mut self, keyword: &str) -> Result<()> {
+        if !keyword.is_empty() && is_identifier(keyword) {
```
jgallagher: This is an interesting solution but feels too general to me; I worry we've missed some way of abusing this. What if we exposed a very limited set of (essentially hard-coded) pragmas, either via an enum or functions, and expanded that list as needed over time, instead of allowing arbitrary strings? Especially since users can already run arbitrary strings as queries if they want.

gwenn:
First, the `Sql` is not exported (outside of rusqlite) so no risk for abuse.
Another approach would have been to bind the [sqlite3_mprintf](http://sqlite.org/c3ref/mprintf.html) but variadic functions are not supported on Rust side.
The most important point is to prevent SQL injection. As `PRAGMA` statements do not support [parameters](http://sqlite.org/lang_expr.html#varparam), it seems wise to help users calling any `PRAGMA` instead of letting them generate SQL manually.

jgallagher:
> First, the Sql struct is not exported (outside of rusqlite) so no risk for abuse.

Ah, yeah, I missed that the `pragma` mod itself wasn't public.

> Another approach would have been to bind the sqlite3_mprintf but variadic functions are not supported on Rust side.

We could work around that with some C code potentially, but I'm not sure `sqlite3_mprintf` would help in this particular case. It looks like it adds `%q`/`%Q`/`%w`/`%z`, but it sounds like all of those are for properly quoting/escaping SQL strings, which wouldn't work for PRAGMAs.

> The most important point is to prevent SQL injection. As PRAGMA statements do not support parameters, it seems wise to help users calling any PRAGMA instead of letting them generate SQL manually.

Agreed! I was suggesting we offer an even more restricted API than this PR adds. Instead of 

```rust
db.pragma_update_and_check(None, "journal_mode", &"OFF", |row| row.get_checked(0))
```

for example, we could offer something enum based (untested, but something like):

```rust
db.pragma_update_and_check(None, Pragma::JournalMode(JournalMode::Off), |row| row.get_checked(0))
```

or with more explicit functions (again untested but something like):

```rust
let journal_mode = db.set_pragma_journal_mode(None, JournalMode::Off)?;
```

This would build a bigger API surface (enum cases or functions for each pragma we wanted to support), but would avoid us having to sanity-check a user-provided identifier before building up the pragma string. Would _probably_ be simpler/easier to use, too, not that either is very complicated.

gwenn:
To query the `JournalMode` without updating it, we will need another enum, no ?
And for PRAGMA which can have an optional parameter, like `integrity_check`, a ``Pragma::IntegrityCheck(Some(200))` or two different entries ?

jgallagher:
Yeah, that's definitely the downside to something more specific. Do you think the route this PR takes is a big win over what users can already do (like you showed on [265](https://github.com/jgallagher/rusqlite/issues/265#issuecomment-296752682))? SQL injection _shouldn't_ be a huge worry with PRAGMAs, since they shouldn't be things that take user input. Maybe we just need to include some examples of pragmas that do and don't return values (like you did on #273) in the docs?

gwenn:
Why is there any problem with `PRAGMA` ?
Because they don't support variable/parameter.
It is probably the first/big surprise for devops.

How do we simulate parameter for `PRAGMA` with no restricted input (`PRAGMA` with restricted input will also take advantage of this) ?

If we prevent using `PRAGMA` with user input,
Then, how do we expose a service to generate an [ERD](https://en.wikipedia.org/wiki/Entity–relationship_model) for a limited set of tables ?
Or how do we expose an API similar to [JDBC](https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)) ?
Or [Diesel](https://github.com/diesel-rs/diesel/blob/258b72597f2b55bdbb03380e5edf81bd147437da/diesel_infer_schema/src/sqlite.rs#L51). Are you sure that the table name will never be `boom'--` ?

gwenn:
SQLite 3.20 introduces [PRAGMA function](http://sqlite.org/pragma.html#pragfunc).
`SELECT * FROM pragma_table_info(?)` works.

thomcc:
> SQL injection shouldn't be a huge worry with PRAGMAs, since they shouldn't be things that take user input

[`PRAGMA key`](https://www.zetetic.net/sqlcipher/sqlcipher-api/#key), when using the `sqlcipher` feature conceivably could for some users.

PR #276, opened 2017-05-20
Add Other variant for the Error type

cactorium:
Hello! This just adds a variant to Error that's usable in ToSql implementations

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -75,8 +75,8 @@ pub enum Error {
     #[allow(dead_code)]
     UserFunctionError(Box<error::Error + Send + Sync>),
 
-    /// Error available for the implementors of the `FromSql` trait.
-    Other(Box<error::Error + Send + Sync>),
+    /// Error available for the implementors of the `ToSql` trait.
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -74,6 +74,9 @@ pub enum Error {
     #[cfg(feature = "functions")]
     #[allow(dead_code)]
     UserFunctionError(Box<error::Error + Send + Sync>),
+
+    /// Error available for the implementors of the `FromSql` trait.
+    Other(Box<error::Error + Send + Sync>),
```
jgallagher: The PR said this was for `ToSql` but this comment says it's for `FromSql`. There is already one error case for `FromSql` conversion errors. Could you confirm if this was for `ToSql` errors or `FromSql` errors?

I'd prefer a name with more clarity than `Other`. If this is for `ToSql`, maybe something like `ToSqlConversionFailure`?

cactorium: Sorry about that, fixed! I renamed it to `ToSqlConversionFailure`

PR #278, opened 2017-05-28
for msvc abi builds, add support for using sqlite from a vcpkg installation if available

mcgoo:
This is with a view to enabling more easy `cargo install` of Diesel on windows.  [Vcpkg](https://github.com/Microsoft/vcpkg) is a newish Microsoft maintained collection of packages, like a ports tree or brew.

There is a script [here](https://github.com/mcgoo/vcpkg_diesel_build) that demonstrates building a totally static diesel.exe with Postgres, MySQL and Sqlite. It's more or less:

```
vcpkg --triplet x64-windows-static install libpq sqlite3 libmysql
set RUSTFLAGS=-Ctarget-feature=+crt-static
cargo install diesel_cli
```
The corresponding PRs for the -sys crates for [PostgreSQL](https://github.com/sgrif/pq-sys/pull/13) and [MySQL](https://github.com/sgrif/mysqlclient-sys/pull/4).

Thanks very much for taking the time to consider this PR.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,12 @@
 environment:
-  TARGET: 1.15.0-x86_64-pc-windows-gnu
-  MSYS2_BITS: 64
+  matrix:
+  - TARGET: 1.15.0-x86_64-pc-windows-gnu
+    MSYS2_BITS: 64
+  - TARGET: 1.15.0-x86_64-pc-windows-msvc
+    VCPKG_DEFAULT_TRIPLET: x64-windows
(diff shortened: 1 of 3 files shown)


gwenn:
I am not sure that it is feasible but the `vcpkg` dependency/feature should be activated only on windows platform.

gwenn:
Thanks !

jgallagher:
I don't have a Windows environment to test on, but if this makes things easier for you and it passes CI, 👍 from me. Should we add anything to the README about this change? I'll go ahead and merge; feel free to reply here or on a different issue.

PR #285, opened 2017-06-02
Removed mutable references in transaction.rs

Mason-Chou:
I've noticed that majority of the code in transaction.rs uses a mut reference when it isn't being utilized in the function call. This should help out newer people and removes the need in making the connection variable mutable.


gwenn:
You should have a look at [this](92834951e36eee06dd1fc1482b113338943f0d76):
> Make the creation of transactions and savepoints take &mut self.
Transactions in SQLite are nested, but the previous API allowed rusqlite
transaction wrappers to be created as "siblings". This resulted in
unexpected (and usually wrong) behavior.

Mason-Chou:
Link is a dead link, can you relink the pull request? 

Thanks!

gwenn:
Link updated...

Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
 name = "rusqlite"
-version = "0.11.0"
-authors = ["John Gallagher <jgallagher@bignerdranch.com>"]
+version = "0.11.1"
+authors = ["John Gallagher <jgallagher@bignerdranch.com>, Mason Chou <masonchou95035@gmail.com>"]
 description = "Ergonomic wrapper for SQLite"
-repository = "https://github.com/jgallagher/rusqlite"
+repository = "https://github.com/Mason-Chou/rusqlite"
```
dashed: stray commit?

Mason-Chou: Yes this is my bad, I cloned the repository to do some edits/changes. Forgot to change it back.

Mason-Chou:
Gwenn, I've looked through your link and I'm still having a hard time understanding why you need it to be a mutable reference. I've ran the tests with my version of the codebase and it passes all of them without failing. Maybe it's a good idea to add a test for that "sibling" examples (Seems like it's already been made)? 

I understand if you need it to need to change variables inside Connection/Transaction/Savepoint structs but examples like below doesn't seem to help on what you are trying to prevent. Having random mutable parameters in function calls is messy and shouldn't be the way forward.

```rust
/// Begin a new transaction. Cannot be nested; see `savepoint` for nested transactions.
pub fn new(conn: &mut Connection, behavior: TransactionBehavior) -> Result<Transaction> {
    let query = match behavior {
        TransactionBehavior::Deferred => "BEGIN DEFERRED",
        TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
        TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
    };

    // conn isn't even being changed at all in here. Nested Savepoints should work 
    // the same even if conn is immutable .
    conn.execute_batch(query)
        .map(move |_| {
                 Transaction {
                     conn: conn,
                     drop_behavior: DropBehavior::Rollback,
                     committed: false,
                 }
             })
}
```

gwenn:
Just try to create two transactions (siblings) from the same connection.

Mason-Chou:
I'm still not sure what you are trying to get at. In this snippet, I've created two transactions from the same connection and it compiles and runs fine. This is using the version without the mutable references.

```rust
extern crate rusqlite;

use rusqlite::Connection;

fn main() {
    
    db.execute("CREATE TABLE foo(x INT NOT NULL);", &[]).unwrap();
    
    let tx = db.transaction().unwrap();
    tx.execute_batch("INSERT INTO foo VALUES(1);").unwrap();
    tx.commit().unwrap();
    
    let tx2 = db.transaction().unwrap();
    {
        let sp1 = tx2.savepoint().unwrap();
        sp1.execute_batch("INSERT INTO foo VALUES(2);").unwrap();
        sp1.commit().unwrap();
    }
    {
        let mut sp2 = tx2.savepoint().unwrap();
        sp2.execute_batch("INSERT INTO FOO VALUES(4);").unwrap();
        sp2.rollback().unwrap();
        sp2.execute_batch("INSERT INTO FOO VALUES(5);").unwrap();
        sp2.execute_batch("INSERT INTO FOO VALUES(6);").unwrap();
        sp2.commit().unwrap(); 
    }
    tx2.commit().unwrap();
    
    //This should print 14.
    println!("{}", db.query_row("SELECT SUM(x) FROM foo;", &[], |r|-> i32 { r.get(0) }).unwrap());
}
```

gwenn:
Please, just try to move the line `tx.commit().unwrap();` at the end.

Mason-Chou:
I'm not a big fan of this approach. I understand the need to have it error out at compile time, but the implementation is vague and confusing. Might need to drop a few comment lines on why some of these parameters are mutable references.

I'm going to close this request as I don't see it proceeding any further. Thanks for the help!

PR #286, opened 2017-06-16
msvc/vcpkg dynamic builds now require explicit opt-in

mcgoo:
The build helper was changed to require explicit opt-in for linking dynamically, otherwise `cargo install` using the msvc toolchain would prefer dynamically linking if Vcpkg is installed resulting in binaries that won't work without a PATH adjustment. This commit makes the appveyor build opt-in.


mcgoo:
@jgallagher This will fix the appveyor breakage for new pull requests coming in. Sorry for the hassle.

PR #290, opened 2017-07-19
Added drop for Connection to prevent fd leak

edelangh:
Added call to flush_prepared_statement_cache
otherwise InnerConnection receives DatabaseBusy on sqlite3_close

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -871,14 +871,17 @@ impl InnerConnection {
 }
 
 impl Drop for InnerConnection {
-    #[cfg(not(test))]
     #[allow(unused_must_use)]
     fn drop(&mut self) {
-        self.close();
-    }
-    #[cfg(test)]
-    fn drop(&mut self) {
-        self.close().expect("Error while closing SQLite connection");
+        use std::thread::panicking;
+
+        if let Err(e) = self.close() {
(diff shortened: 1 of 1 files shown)


gwenn:
Do you mind adding the following changes to your PR which make tests failed _without_ your patch:
```Rust
@@ -894,10 +894,15 @@ impl InnerConnection {
 }

 impl Drop for InnerConnection {
+    #[cfg(not(test))]
     #[allow(unused_must_use)]
     fn drop(&mut self) {
         self.close();
     }
+    #[cfg(test)]
+    fn drop(&mut self) {
+        self.close().expect("Error while closing SQLite connection");
+    }
 }

 /// Old name for `Statement`. `SqliteStatement` is deprecated.
```
?
Thanks.

edelangh:
Absolutely ! Thank you

gwenn:
I saw this [link](https://github.com/diesel-rs/diesel/blob/036985ed2c2d2ac1b927f810b89af54d5852826d/diesel/src/sqlite/connection/stmt.rs#L156-L170) on reddit this morning.
May be we should panic in all cases (not just when running tests)?

edelangh:
What do you think about that ?

PR #299, opened 2017-10-16
[ENH] Added support for chrono 0.4.0

vojtechcima:
Hello,

here is a patch fixing compatibility with the latest version of chrono crate (0.4.0), where "UTC" type has been renamed to "Utc". Note that this patch will not work with older versions of chrono.

Best regards,
Vojtech


jgallagher:
Equivalent changes were merged in as #302. Thanks!

PR #315, opened 2017-12-03
Remove the database keyword/category from libsqlite3-sys

sgrif:
Currently `libsqlite3-sys` is the first result for both the "database"
keyword, and the "Database interfaces" category. This makes sense, as it
is a dependency of both Diesel and rusqlite. However, this library is
not intended to be used directly. While it can technically be called a
database interface, FFI is clearly the category that applies more than
anything else. Someone browsing this keyword or category is likely
looking for a Rust library they can use, not a C one.


jgallagher:
👍

PR #317, opened 2017-12-06
Use sqlite3_unlock_notify behind a feature

### LINKED ISSUE #294: Add support to unlock notification
See [Unlock Notification](http://sqlite.org/unlock_notify.html) and [Shared-Cache Mode](http://sqlite.org/sharedcache.html)

sfackler:
Closes #294 

Ideally we'd support user-defined callbacks but that'll be pretty tricky to define in a safe way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ script:
   - cargo test --features serde_json
   - cargo test --features bundled
   - cargo test --features sqlcipher
+  - cargo test --features unlock_notify
   - cargo test --features "backup blob chrono functions limits load_extension serde_json trace"
   - cargo test --features "backup blob chrono functions limits load_extension serde_json trace buildtime_bindgen"
   - cargo test --features "backup blob chrono functions limits load_extension serde_json trace bundled"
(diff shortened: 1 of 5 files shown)


gwenn:
Three month ago, I pushed [this](https://github.com/gwenn/rusqlite/compare/master...gwenn:unlock_notify).
But no PR.

gwenn:
I managed to polish my original changes [here](https://github.com/jgallagher/rusqlite/compare/master...gwenn:unlock-notify).
* `SQLITE_ENABLE_UNLOCK_NOTIFY` compilation option is defined only when `unlock_notify` feature  is activated
* Code remains the same when `unlock_notify` feature  is not enabled.

There is one thing where I am not sure:
[SQLite Unlock-Notify API](http://sqlite.org/unlock_notify.html)
```C
    if( !un.fired ){
      pthread_cond_wait(&un.cond, &un.mutex);
    }
```
versus
```rust
        while !*guard {
             guard = state.cvar.wait(guard).unwrap();
         }
```
Should we wait in a loop or retry sqlite3_prepare/sqlite3_step ?
Do you know what [spurious wakeups](https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait) mean ?
Thanks.

PR #318, opened 2017-12-24
Make `transaction()` take an immutable &Connection

traviscross:
Previously we were requiring the caller to pass in a `&mut Connection`
even though in the `Transaction` structure we were storing the
`Connection` as an immutable reference.  Since we're not mutating the
`Connection`, requiring this mutable reference only creates problems
for callers.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -94,7 +94,7 @@ pub struct Savepoint<'conn> {
 
 impl<'conn> Transaction<'conn> {
     /// Begin a new transaction. Cannot be nested; see `savepoint` for nested transactions.
-    pub fn new(conn: &mut Connection, behavior: TransactionBehavior) -> Result<Transaction> {
+    pub fn new(conn: &Connection, behavior: TransactionBehavior) -> Result<Transaction> {
         let query = match behavior {
             TransactionBehavior::Deferred => "BEGIN DEFERRED",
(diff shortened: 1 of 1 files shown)


gwenn:
See #285

traviscross:
Hi @gwenn; thanks for the pointer to the earlier ticket.  To summarize, by marking the Connection taken by the Transaction constructor as mutable, we prevent at compile time any attempt to open nested or concurrent transactions on the same connection.  SQLite does not support nested transactions, so attempting to do this would be an error at runtime.

That is indeed a nice property.  Unfortunately it does make certain factorings of code tricky.

I'd suggest we add a comment over the constructor to hopefully make more clear to the next interested reader what is going on.  See #322.

thomcc:
It's worth noting that this doesn't actually prevent nested transactions. You can always bypass it with execute_batch, which doesn't require `mut`.

It does prevent a number of refactorings though, and IMO makes the API more difficult to work with.

gwenn:
https://github.com/jgallagher/rusqlite/pull/285#issuecomment-308510303
> But we can make Rust rejects nested transactions at compile time.

Or more precisely, we can make Rust rejects nested programmatic [`Transaction`](https://docs.rs/rusqlite/0.14.0/rusqlite/struct.Transaction.html) at compile time.

PR #319, opened 2017-12-24
Fix date/time format for SQLite, use RFC 3339

traviscross:
We implement `ToSql` and `FromSql` for `time::Timespec` values.  Our
documentation indicates that we store the value in the same format
used by SQLite's built-in date/time functions, but this was not
correct.

We were using the format:

    %Y-%m-%d %H:%M:%S:%f %Z

This format cannot be interpreted at all by SQLite's built-in
date/time functions.  There are three reasons for this:

- SQLite supports only two timezone formats: `[+-]HH:MM` and the
  literal character `Z` (indicating UTC)

- SQLite does not support a space before the timezone indicator

- SQLite supports a period (`.`) between the seconds field and the
  fractional seconds field, but not a colon (`:`)

SQLite does support the RFC 3339 date/time format, which is standard
in many other places.  As we're always storing a UTC value, we'll
simply use a trailing `Z` to indicate the timezone, as allowed by RFC
3339.  The new format is:

    %Y-%m-%dT%H:%M:%S.%fZ

To avoid breaking applications using databases with values in the old
format, we'll continue to support it as a fallback for `FromSql`.

[1] https://www.sqlite.org/lang_datefunc.html
[2] https://tools.ietf.org/html/rfc3339

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -10,12 +10,13 @@
 //! * Strings (`String` and `&str`)
 //! * Blobs (`Vec<u8>` and `&[u8]`)
 //!
-//! Additionally, because it is such a common data type, implementations are provided for
-//! `time::Timespec` that use a string for storage (using the same format string,
-//! `"%Y-%m-%d %H:%M:%S"`, as SQLite's builtin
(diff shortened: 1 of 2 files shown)


gwenn:
I agree but it's a breaking change (the data stored in SQLite are not the same).

PR #320, opened 2017-12-25
Allow build time bindgen of bundled SQLite

traviscross:
Previously the `buildtime_bindgen` feature was ignored when the
`bundled` feature was set; the `bindgen_bundled_version.rs` file was
always used.

With this commit, when both `bundled` and `buildtime_bindgen` are set,
we perform `bindgen` at build time with the bundled SQLite.


gwenn:
Would you mind closing the PR ?

PR #322, opened 2017-12-31
Add comment to justify `&mut Connection` in `Transaction`

traviscross:
The `Transaction` implementation never actually mutates the `Connection` reference we give it.  In fact, the `Transaction` structure itself only requires an immutable connection.  So it can be surprising to readers that the constructor requires a `&mut Connection`.  We do this so as to prevent at compile-time nested or concurrent transactions on the same connection as these are not allowed by SQLite.

In this commit, we add a comment explaining this nuance.


gwenn:
Thanks and sorry for the late response.
I guess that the AppVeyor error can be ignored.

PR #351, opened 2018-05-07
Add flag to allow opening DB with write-ahead logging

tjkirch:
Background: https://www.sqlite.org/wal.html

The OPEN_WAL flag is in the 3.7.3+ bindgen bindings, the sqlite3.h header, and in the bundled_version bindings, so this seemed safe.  The constant is specified as 524288 there; I obviously went for consistency here.

The unit tests still pass, and I was able to use the flag in a local project with an open_with_flags call specifying OpenFlags::SQLITE_OPEN_WAL.

cc: @jgallagher and @gwenn for adding the other constants in this section.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -605,6 +605,7 @@ bitflags! {
         const SQLITE_OPEN_FULL_MUTEX    = ffi::SQLITE_OPEN_FULLMUTEX;
         const SQLITE_OPEN_SHARED_CACHE  = 0x0002_0000;
         const SQLITE_OPEN_PRIVATE_CACHE = 0x0004_0000;
+        const SQLITE_OPEN_WAL           = 0x0008_0000;
     }
 }
 


tjkirch:
The appveyor msvc failure doesn't look related; it failed to download 7zip before building.  I don't see appveyor runs for other recent PRs trying to download 7zip, so I don't know what's different.

gwenn:
I may be wrong but the SQLITE_OPEN_WAL flag should not be used with `sqlite3_open_v2`.
http://sqlite.org/c3ref/c_open_autoproxy.html
> #define SQLITE_OPEN_WAL              0x00080000  /* VFS only */

tjkirch:
Hmm, I guess I don't understand.  I don't see anything related to the vfs interface in rusqlite.  Would we need something to represent a sqlite3_vfs, and the ability to pass that to the open call?

gwenn:
`OpenFlags` should only have 'Ok for sqlite3_open_v2()' flags.
If you want to activate WAL mode, you should use `PRAGMA journal_mode=WAL;`
See http://sqlite.org/wal.html

PR #367, opened 2018-07-27
Roll back uncommitable transactions

### LINKED ISSUE #366: Commit failure unhandled, database left in unusable state
When sqlite is used from multiple processes at the same time, `COMMIT` may fail with "database is locked" error.

After that happens, all subsequent transactions in rusqlite fail with "cannot start a transaction within a transaction".

I think this is because:

```rust
    fn commit_(&mut self) -> Result<()> {
        self.committed = true;
        self.conn.execute_batch("COMMIT")
    }
```

Sets `self.committed = true;` without checking whether commit has succeeded, and when `COMMIT` fails, the transaction remains open — forever.

It should probably be: 

```rust
    fn commit_(&mut self) -> Result<()> {        
        self.conn.execute_batch("COMMIT")?;
        self.committed = true;
        Ok(())
    }
```

kornelski:
This fixes #366 for me

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -197,7 +197,7 @@ impl<'conn> Transaction<'conn> {
             return Ok(());
         }
         match self.drop_behavior() {
-            DropBehavior::Commit => self.commit_(),
+            DropBehavior::Commit => self.commit_().or_else(|_| self.rollback_()),
             DropBehavior::Rollback => self.rollback_(),
             DropBehavior::Ignore => Ok(()),
             DropBehavior::Panic => panic!("Transaction dropped unexpectedly."),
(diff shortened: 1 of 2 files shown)


src/transaction.rs
```diff
@@ -195,7 +197,7 @@ impl<'conn> Transaction<'conn> {
             return Ok(());
         }
         match self.drop_behavior() {
-            DropBehavior::Commit => self.commit_(),
+            DropBehavior::Commit => self.commit_().or_else(|_| self.rollback_()),
```
gwenn: I am not sure but the `.or_else(|_| self.rollback_())` should only be done in `drop` (in case the user wants to wait and retry for example after `finish`).
And maybe the `rollback_` should only be called for some specific error code(s).

src/transaction.rs
```diff
@@ -307,7 +310,7 @@ impl<'conn> Savepoint<'conn> {
             return Ok(());
         }
         match self.drop_behavior() {
-            DropBehavior::Commit => self.commit_(),
+            DropBehavior::Commit => self.commit_().or_else(|_| self.rollback()),
```
gwenn: Above comment applies also here.

src/transaction.rs
kornelski: It's not possible to retry if `finish()` fails, because it consumes `self` and drops it.

gwenn: Right, sorry.

kornelski:
Added

gwenn:
Thanks.

PR #391, opened 2018-09-05
Add an impl for FromSql for Cow<str>

mqudsi:
Some core rust APIs return a copy-on-write str reference instead of a
string (e.g. Path::display()). This allows them to be directly inserted
into the database without needing to call `.to_string()` explicitly.

### DIFF
--- a/src/types/value.rs
+++ b/src/types/value.rs
@@ -1,4 +1,5 @@
 use super::{Null, Type};
+use std::borrow::Cow;
 
 /// Owning [dynamic type value](http://sqlite.org/datatype3.html). Value's type is typically
 /// dictated by SQLite (not by the caller).
@@ -74,6 +75,12 @@ impl From<String> for Value {
 impl From<Vec<u8>> for Value {
     fn from(v: Vec<u8>) -> Value {
         Value::Blob(v)
+    }
+}
+
+impl<'a> From<Cow<'a, str>> for Value {
+    fn from(s: Cow<'a, str>) -> Value {
+        Value::Text(s.to_string())
     }
 }
 


gwenn:
Could you please provide an example ?
Because, if you want to insert a Cow<str>, you need `ToSql` which is already implemented...

mqudsi:
I'm sorry, it wasn't inserts that weren't possible, it was lookups like `WHERE path = ?1` binding to `path.display()`.

gwenn:
The difference between insert values or a select clause doesn't matter.
The `?1` parameter is bound to a rust value.
So you need a `ToSql` (Rust - to -> SQL(ite)).
`FromSql` is used to extract values from SQL(ite).

mqudsi:
I think that was a typo in the title. Actual code:

```rust
impl<'a> From<Cow<'a, str>> for Value {
```

(from rust to sqlite value)

gwenn:
Ok,
So could you please provide a use case where you need to transform a `Cow<str>` to `Value` ?

mqudsi:
```rust
extern crate rusqlite;

use std::path::PathBuf;
use rusqlite::Connection;

fn main() {
    let path = PathBuf::from("/tmp/foo"); // imagine this came from somewhere else

    db.query_row("SELECT * FROM foo WHERE foo.path LIKE ?1",
                 &[&path.to_string_lossy()],
                 |row| {
                     let foo: String = row.get(0);
                     foo
                 }).unwrap();
}
```

```
   Compiling rusqlite-test v0.1.0 (file:///mnt/c/Users/mqudsi/random/rusqlite-test)
error[E0277]: the trait bound `std::borrow::Cow<'_, str>: rusqlite::types::ToSql` is not satisfied
  --> src/main.rs:11:20
   |
11 |                  &[&path.to_string_lossy()],
   |                    ^^^^^^^^^^^^^^^^^^^^^^^ the trait `rusqlite::types::ToSql` is not implemented for `std::borrow::Cow<'_, str>`
   |
   = note: required for the cast to the object type `dyn rusqlite::types::ToSql`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `rusqlite-test`.

To learn more, run the command again with --verbose.
```

gwenn:
Which version of rusqlite is used ?
https://docs.rs/rusqlite/0.14.0/rusqlite/types/trait.ToSql.html
> impl<'a> ToSql for Cow<'a, str>

gwenn:
Could you please confirm that the following example works with `rusqlite` 0.14.0:
```rust
extern crate rusqlite;

use std::path::PathBuf;
use rusqlite::Connection;

fn main() {
    let path = PathBuf::from("/tmp/foo"); // imagine this came from somewhere else

    db.execute_batch("CREATE TABLE foo (path TEXT);").unwrap();
    db.execute("INSERT INTO foo (path) VALUES (?)", &[&path.to_string_lossy()]).unwrap();

    db.query_row("SELECT * FROM foo WHERE foo.path LIKE ?1",
                 &[&path.to_string_lossy()],
                 |row| {
                     let foo: String = row.get(0);
                     foo
                 }).unwrap();
}
```
?
And close the PR accordingly.
Thanks.

mqudsi:
@gwenn thank you. It seems I was using 0.13 before I checked out the repo to implement this PR.

Is there anything that would require `From<T> for Value` where `ToSql for T` is already implemented?

PR #402, opened 2018-09-13
Add query_scalar convenience method to Statement

rmini:
I have a bunch of code in a few different projects looking up IDs that generally looks like:

```rust
let mut stmt = conn.prepare_cached("SELECT id FROM table WHERE k1 = ?1 AND k2 = ?2")?;
match stmt.query_map(&[&k1, &k2], |r| r.get_checked(0))?.next() {
    Some(v) => Ok(v??),
    None => Ok(None)
}
```

I have a helper function I use to clean this up, but I thought this might be generally useful in the API.

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -10,7 +10,7 @@ use super::str_to_cstring;
 use super::{
     AndThenRows, Connection, Error, MappedRows, RawStatement, Result, Row, Rows, ValueRef,
 };
-use types::{ToSql, ToSqlOutput};
+use types::{ToSql, ToSqlOutput, FromSql};
 #[cfg(feature = "array")]
 use vtab::array::{free_array, ARRAY_TYPE};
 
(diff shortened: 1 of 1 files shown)


gwenn:
What about refactoring / breaking `query_row` such as it returns a `Result<Option<T>>` instead ?

rmini:
If I had something like Statement::query_row but with F being ```F: FnOnce(&Row) -> Result<T>``` I wouldn't need to do as many contortions to get from ```Result<Option<Result<T>>>``` to something useful. It's the fact that Row::get can panic and Row::get_unchecked returns Result<T> that makes query_row awkward beyond just the Option<>.

gwenn:
`NULL` value can be converted to `None`:
https://docs.rs/rusqlite/0.14.0/src/rusqlite/types/from_sql.rs.html#138-145
So there is no way for the caller to make the distinction between no row or a row with a `NULL` value.
Or an ugly one `Result<Option<Option<T>>>`.

rmini:
I've found I can generally let type inference do the dirty work to keep things from getting too ugly as far as explicit typing goes.

gwenn:
Ok,
When there is no row matching the specified query: 
- `query_row` uses a specific error `QueryReturnedNoRows`
- `query_row_opt` uses an `Option` instead

When a `NULL` value is extracted:
- `query_row` returns a `Result<Option<T>>`
- `query_row_opt` returns a `Result<Option<Option<T>>`

`query_row` is convenient when at least one row is expected.
`query_row_opt` is convenient when there is no row.

`query_scalar` seems redundant to me.

gwenn:
There is also [onecolumn](https://sqlite.org/tclsqlite.html#onecolumn) provided by the TCL API.
So your PR is fine.
I just need some time and make sure we cannot merge `query_row` and `query_row_opt`.
Thanks you.

gwenn:
Just a note to take a look at https://docs.rs/diesel/1.3.3/diesel/query_dsl/trait.RunQueryDsl.html#method.first
> If the query truly is optional, you can call .optional() on the result of this to get a Result<Option<U>>.

https://docs.rs/diesel/1.3.3/diesel/result/trait.OptionalExtension.html#tymethod.optional
> By default, Diesel treats 0 rows being returned from a query that is expected to return 1 row as an error (e.g. the return value of get_result or first). This method will handle that error, and give you back an Option<T> instead.

PR #408, opened 2018-09-26
Add a method of interrupting a query executing on a separate thread, fixes #407

### LINKED ISSUE #407: Feature request: Some way of interrupting queries
It would be nice if rusqlite exposed something like sqlite3_interrupt.

Use case: We'd like to use rusqlite in an browser location bar scenario, however when the user changes what is input into the location bar (e.g. by typing a new key or deleting something) the previously running query is no longer useful and should be aborted.

At the moment the lack of thread-safety in rusqlite limits how this could be done, but it's possible there could be a more limited API exposed.

thomcc:
This adds an InterruptHandle type which is Send+Sync, and exposes a single interrupt method. It holds onto a `Arc<Mutex<*ffi::sqlite3>>`, which allows us to prevent calling sqlite3_close while an interrupt is occurring (which is called out in https://www.sqlite.org/c3ref/interrupt.html as unsafe).

I'm unsure if there's a more robust way to test this, but I'm open to suggestions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,7 +79,7 @@ use std::ptr;
 use std::result;
 use std::str;
 use std::sync::atomic::{AtomicBool, Ordering, ATOMIC_BOOL_INIT};
-use std::sync::{Once, ONCE_INIT};
+use std::sync::{Once, ONCE_INIT, Arc, Mutex};
 
 use cache::StatementCache;
 use error::{error_from_handle, error_from_sqlite_code};
@@ -598,6 +598,12 @@ impl Connection {
         self.db.borrow().db()
     }
 
+    /// Get access to a handle that can be used to interrupt long running queries
+    /// from another thread.
+    pub fn get_interrupt_handle(&self) -> InterruptHandle {
+        self.db.borrow().get_interrupt_handle()
+    }
+
     fn decode_result(&self, code: c_int) -> Result<()> {
         self.db.borrow_mut().decode_result(code)
     }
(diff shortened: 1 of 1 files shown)


thomcc:
Note: I'd be willing to implement this as a `#[feature]`, but the code size is small, and changes should only effect the performance of `close()` which is... hopefully very far from anybody's bottleneck?

(Uncontended mutex access is quite fast anyway, so it's not like it should effect the performance of `close()` very much).

src/lib.rs
```diff
@@ -629,6 +635,13 @@ impl fmt::Debug for Connection {
 
 struct InnerConnection {
     db: *mut ffi::sqlite3,
+    // It's unsafe to call `sqlite3_close` while another thread is performing
+    // a `sqlite3_interrupt`, and vice versa, so we take this mutex during
+    // those functions. This protects a copy of the `db` pointer (which is
+    // cleared on closing), however the main copy, `db`, is unprotected.
+    // Otherwise, a long running query would prevent calling interrupt, as
+    // interrupt would only acquire the lock after the query's completion.
+    interrupt_lock: Arc<Mutex<*mut ffi::sqlite3>>,
```
gwenn: RwLock ?
Many readers (InterruptHandle) but only one writer (during close).

src/lib.rs
```diff
@@ -887,16 +908,28 @@ impl InnerConnection {
             return Ok(());
         }
         self.remove_hooks();
+        let mut shared_handle = self.interrupt_lock.lock().unwrap();
```
gwenn: Maybe we should close the connection even is lock is poisoned ?

src/lib.rs
```diff
@@ -1472,6 +1530,85 @@ mod test {
...
+                    .collect::<Vec<_>>()
+                    .join(";\n"));
+
+            // Do this on the main thread to minimize the amount of time spent
+            // when interrupt won't do anything (because we haven't started
+            // executing the query).
+            let c_sql = str_to_cstring(&sql).unwrap();
+
+            let joiner = thread::spawn(move || {
+                unsafe {
```
gwenn: I am sorry but I don't understand why we should use unsafe code here ?

thomcc: It's to avoid the case where the sqlite_interrupt happens during the `str_to_cstring` done inside execute (it's a pretty big string...). In practice this happened a lot, and this made the test a decent amount more reliable.

src/lib.rs
gwenn: SQLite documentation does not specify if it is safe to call `sqlite3_interrupt` multiple times/concurrently...
So I guess a Mutex is ok.

src/lib.rs
thomcc: I think this could lead to double sqlite3_close calls, which is probably worse than panicking (it's a memory safety problem).

gwenn: I don't see how InnerConnection.close can be called twice ?

thomcc: The only two places that could poison the lock are `InnerConnection::close()` or `InterruptHandle::interrupt()`. ISTM like the former is more likely since it does a lot more. At the moment, there should be no way for it to happen twice, but it seems more likely to me than `interrupt()`.

PR #410, opened 2018-10-07
Upgrade bindgen and SQLite bundled sources

gwenn:


gwenn:
https://github.com/KyleMayes/clang-sys/issues/83 ?

PR #411, opened 2018-10-08
Add a feature for storing i128 as blobs.

thomcc:
Note: I couldn't figure out how to unify this with the https://github.com/jgallagher/rusqlite/pull/392 PR, or even if that is a good thing; it's code is substantially different anyway (Apologies either way). Regardless, most of the differences are due to the fact that I agree with the comments at https://github.com/jgallagher/rusqlite/pull/392#issuecomment-424089697.

This is behind the `i128_blob` feature.

Blobs are stored as 16 byte big-endian values, with their most significant bit flipped. This is so that sorting, comparison, etc all work properly, even with negative numbers. This also allows the representation to be stable across different computers.

It's possible that the `FromSql` implementation should handle the case that the real value is stored in an integer. I didn't do this, but would be willing to make the change. I don't think we should store them this way though (e.g. IMO `ToSql` shouldn't work like this), since I don't think users would be able to sort/compare them sanely.

Support for `u128` is not implemented, as comparison with `i128` values would work strangely. This also is consistent with `u64` not being allowed, not that I think that would be reason enough on it's own.

The `byteorder` crate is used if this feature is flipped, as it's quite small and implements things more or less optimally. If/when `i128::{to,from}_be_bytes` gets stabilized (https://github.com/rust-lang/rust/issues/52963), we should probably use that instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ script:
   - cargo test --features serde_json
   - cargo test --features bundled
   - cargo test --features sqlcipher
+  - cargo test --features i128_blob
   - cargo test --features "unlock_notify bundled"
   - cargo test --features "array bundled csvtab vtab"
   - cargo test --features "backup blob chrono csvtab functions hooks limits load_extension serde_json trace vtab"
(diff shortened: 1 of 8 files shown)


gwenn:
Thanks.

PR #414, opened 2018-10-18
Allow getting a ValueRef out of Row and Context, fixes #259

### LINKED ISSUE #259: Allow retrieving borrowed values from a row
Currently, if I want to deserialize a blob it seems (correct me if I'm wrong) that I need to either

1. retrieve a `Vec<u8>`, and then pass that to a deserialization function
2. do the dance to retrieve a blob as a reader

I would like to propose allowing

3. retrieving the borrowed value of a row (i.e. a `&[u8]`)

Looking at the `FromSql` trait, I'm not sure it's possible to do it within the constraints imposed by that as it stands right now - it expects the return value to have an independent lifetime to the input `ValueRef`. So it seems to me there are three ways to add this:

1. add a lifetime constraint to the `FromSql` trait, which shouldn't affect any existing code (I believe) because no lifetimes are currently returned, to allow returning borrowed values as well
2. create a new trait `FromSqlBorrowed` and appropriate new methods like `get_borrowed` etc
3. add a `get_raw` to `Row` which allows directly retrieving a `ValueRef`

Personally I'm in favour of the first option, as long as I'm correct in my suspicion that it shouldn't affect any existing code.

Thoughts? Have I missed anything?

thomcc:
This takes approach 3 in #259, specifically letting rusqlite return a `ValueRef` from `Row` and `Context` (for my case I care more about `Context` than with `Row`, but in the past I've certainly wanted both). This seems the simplest and most straightforward to me, and I don't really think I see the benefit of trying to make this more generic/flexible, although if you have a concrete suggestion I'm open.

I don't have an opinion on the name `get_raw`, I picked it only because I needed this feature, and it was just the name suggested in #259. It doesn't seem great since it sort of implies unsafety (and it's completely safe), but something like `get_ref` feels like it should be higher level than this is. But again, I don't really care about the name here.

Anyway, this changes a few lifetime bounds:

- The data returned by `ValueRef<'a>`'s `as_str()`/`as_blob()` is bound to `'a` (this just seems like an accident to me that it wasn't, you could already get this as `'a` if you did the match yourself). Without this, `a.get_raw().as_str()` doesn't work.

- `Row<'a, 'stmt>` now requires the `'stmt` outlives `'a`, because we need to return `ValueRef<'a>` from the Row (`ValueRef<'stmt>`, which `stmt.value_ref()` returns, would be allowed to live past calls to `sqlite3_step`, which is bad).

Using `thing.get_raw().as_{str,blob}()` is seems ergonomic enough, but the lack of an automatic conversion from `FromSqlError` (returned by as_str/blob) to `rusqlite::Error` does make this less ergonomic. This is the part of the patch I'm least happy with, but I wasn't completely sure what the best approach for this would be.

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -139,6 +139,16 @@ impl<'a> Context<'a> {
         })
     }
 
+    /// Returns the `idx`th argument as a `ValueRef`.
+    ///
+    /// # Failure
+    ///
+    /// Will panic if `idx` is greater than or equal to `self.len()`.
+    pub fn get_raw(&self, idx: usize) -> ValueRef<'a> {
+        let arg = self.args[idx];
+        unsafe { ValueRef::from_value(arg) }
+    }
+
     /// Sets the auxilliary data associated with a particular parameter. See
     /// https://www.sqlite.org/c3ref/get_auxdata.html for a discussion of
     /// this feature, or the unit tests of this module for an example.
@@ -530,8 +540,10 @@ mod test {
         let is_match = {
             let re = saved_re.unwrap_or_else(|| new_re.as_ref().unwrap());
 
-            let text = try!(ctx.get::<String>(1));
-            re.is_match(&text)
+            let text = ctx.get_raw(1).as_str().map_err(|e|
+                Error::UserFunctionError(e.into()))?;
+
+            re.is_match(text)
         };
 
         if let Some(re) = new_re {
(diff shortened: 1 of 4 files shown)


gwenn:
Thanks.

thomcc:
Any idea when the next release will come out? This, #404, and #408 fix some pain points for me.

gwenn:
I will try on Sunday afternoon.

PR #417, opened 2018-10-24
Update README example.

vladh:
The code example from the documentation's latest version contains some important updates, such as the usage of `NO_PARAMS`. The README should be updated to reflect this.


gwenn:
Thanks.

Issue #431, opened 2018-11-18
Datetimes don't parse correctly


Bodhizafa:
The implementation of FromSql for Timespec uses strptime with the following formats:

const SQLITE_DATETIME_FMT: &str = "%Y-%m-%dT%H:%M:%S.%fZ";
const SQLITE_DATETIME_FMT_LEGACY: &str = "%Y-%m-%d %H:%M:%S:%f %Z";

SQLite's docs (https://sqlite.org/lang_datefunc.html) specify either of those as correct for inserting datetimes into SQLite, but its output formats don't match, they are: 

date(...) |   | strftime('%Y-%m-%d', ...)
-- | -- | --
time(...) |   | strftime('%H:%M:%S', ...)
datetime(...) |   | strftime('%Y-%m-%d %H:%M:%S', ...)
julianday(...) |   | strftime('%J', ...)

so when I try to convert the output of "SELECT DATETIME('now')" into a Timespec, I get an error because the output doesn't match format.

gwenn:
The [original](https://github.com/gwenn/rusqlite/blob/chrono/src/types/time.rs#L11) was ok.
I will check why this is not the case anymore.

gwenn:
https://github.com/jgallagher/rusqlite/pull/305

gwenn:
Fixed by 0.16.0 release.

PR #432, opened 2018-11-18
Add Timespec parsing for DATETIME('now') output format

Bodhizafa:
Fixes issue #431


gwenn:
Would you mind closing the PR ?

PR #444, opened 2018-12-15
Add support for forcing cargo:rustc-link-lib to link as native

thomcc:
This adds support for setting SQLITE_STATIC=1 (or SQLCIPHER_STATIC for sqlcipher builds) in the environment to force us to link sqlite/sqlcipher statically. If this isn't specified, or it's specified to be 0, it uses the default which is dylib.

This is needed\* in order to build a statically linked rust library that includes these (without the bundled feature).

This is still not perfect, a more robust build.rs might inspect what the actual library files are and decide based on that, as the openssl-sys crate does https://github.com/sfackler/rust-openssl/blob/91a78bc38b09b7e2dd2a0b493407e3fb0550072a/openssl-sys/build/main.rs#L572, but that's complex and doesn't seem super worthwhile.

\* Well, "needed". You could always fixup the static lib after the fact, but that sucks.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -108,18 +108,27 @@ mod build {
         bindings::write_to_out_dir(header, out_path);
     }
 
+    fn find_link_mode() -> &'static str {
+        // If the user specifies SQLITE_STATIC (or SQLCIPHER_STATIC), do static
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.

PR #451, opened 2018-12-16
Introduce `OptionalExtension`

gwenn:
Like in Diesel.
Fix #402

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -144,6 +144,26 @@ pub const NO_PARAMS: &[&dyn ToSql] = &[];
 /// A typedef of the result returned by many methods.
 pub type Result<T> = result::Result<T, Error>;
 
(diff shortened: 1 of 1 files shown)


gwenn:
Yep !
Why ?

rainhead:
Well, I was hoping to use it.

rainhead:
The trait is public, but the method is not, which looked like a mistake.

gwenn:
I guess:
https://stackoverflow.com/questions/25273816/why-do-i-need-to-import-a-trait-to-use-the-methods-it-defines-for-a-type
> In Rust a trait must be in scope for you to be able to call its methods.

rainhead:
You're right, it wasn't. I don't understand why I'm allowed to call that method, but I'll go with it. Thanks.

PR #463, opened 2019-01-25
Make get_aux safe by storing the TypeId with the data.

thomcc:
Is there a reason not to do this?


gwenn:
I don't know. I had never used `std::any::` before...

src/functions.rs
```diff
@@ -699,6 +699,28 @@ mod test {
...
+            } else {
+                assert_eq!(ctx.get_aux::<String>(0), Err(Error::GetAuxWrongType));
+                assert_eq!(ctx.get_aux::<i64>(0), Ok(Some(&100)));
+            }
+            Ok(true)
+        })
+        .unwrap();
+
+        let res: bool = db.query_row(
+            "WITH foo(i) AS (VALUES (0), (1)) SELECT example(0, i) FROM foo",
```
gwenn: [WITH clause](https://sqlite.org/releaselog/3_8_3.html) are only supported by SQLite >= 3.8.3.
I suggest you use this syntax instead to fix CI builds:
```sql
SELECT example(0, i) FROM (SELECT 0 as i UNION SELECT 1);
```

PR #464, opened 2019-01-25
Add support for querying sqlite_stmt_status

thomcc:
Useful for e.g. logging warnings if sorts are done but you don't want them to be done.


gwenn:
Thanks.

PR #470, opened 2019-01-27
Derive Debug for Transaction

rainhead:
It seems you can't `unwrap()` an `Ok(Transaction)` (produced in my case by `Rc::try_unwrap()` unless `Transaction` is `Debug`.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -556,6 +556,16 @@ mod test {
         assert_current_sum(8, &db);
     }
 
+    #[test]
+    fn test_rc() {
+        use std::rc::Rc;
+        let mut conn = Connection::open_in_memory().unwrap();
+        let rc_txn = Rc::new(conn.transaction().unwrap());
+
+        // This will compile only if Transaction is Debug
+        Rc::try_unwrap(rc_txn).unwrap();
+    }
+
     fn insert(x: i32, conn: &Connection) {
         conn.execute("INSERT INTO foo VALUES(?)", &[x]).unwrap();
     }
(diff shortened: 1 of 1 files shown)


gwenn:
I see no such constraint in [try_unwrap](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.try_unwrap) signature !

rainhead:
Right, the problem was with `Result::unwrap`.

gwenn:
Still no constraint on [unwrap](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap) !

rainhead:
Here is the code that compiles only with the changes in this PR. I don't know how to make "doesn't compile" into a test.

```rust
let mut conn = rusqlite::Connection::open_in_memory().unwrap();
let rc_txn = Rc::new(conn.transaction().unwrap());
let txn = Rc::try_unwrap(rc_txn).unwrap();
```

gwenn:
Indeed, the transaction can by moved in the error `E` in `Result<T, E>` and
https://doc.rust-lang.org/src/core/result.rs.html#779-838
```rust
impl<T, E: fmt::Debug> Result<T, E> {
```

rainhead:
Great. Do you want something other than the code I've written?

gwenn:
Maybe a simple unit test ?
Like:
```rust
#[test]
fn test_debug() {
    fn assert_debug<T: Debug>() {}
    assert_debug::<Transaction<'static>>();
}
```
Or https://github.com/jgallagher/rusqlite/pull/470#issuecomment-458351836 in a test method.
Thanks.

gwenn:
Thanks.

rainhead:
Thank you!!

PR #471, opened 2019-01-29
Add params/named_params macro, and expose ToSql from top level

### LINKED ISSUE #462: Cannot insert heterogeneous elements with .execute
I have a database created by:
```sql
CREATE TABLE quotes (
    id          INTEGER PRIMARY KEY,
    quote       TEXT NOT NULL,
    author      VARCHAR(255),
    source      TEXT
);
```

and I want to run the following query:

```rust
const SQL_INSERT_QUOTE: &'static str =
" INSERT INTO quotes (quote, author, source) VALUES (?1, ?2, ?3)";
```

With the code:

```rust
pub fn add_quote(connection: &mut Connection, quote: &Quote) -> Result<(), Error> {
    let mut statement = connection.prepare(SQL_INSERT_QUOTE);
    statement.execute(&[&quote.quote, &quote.author, &quote.source])?;
    Ok(())
}
```

Here, `quote.source` is an `Option<String>`, while the others are `String`, so there's no way (that I can see) to insert it, because I can't access `ToSql` to make the elements of the array into `&dyn ToSql`, and the array has to be homogeneous.

What's the correct way to do this?

thomcc:
Nobody asked for `named_params!`, but it would be nice. It would be nicer if it were supported as the params! macro, but doing that would require a lot more work (see https://github.com/serde-rs/json/blob/master/src/macros.rs, for example).

I've also made some of the examples and tests use this, although not all of them as the intent isn't to deprecate passing these explicitly, just to make doing it more convenient.

Re-exposing ToSql from the top level is done just to make it more obvious. It's a fairly important part of the API, and it feels buried for it to live in `rusqlite::types::`.

Fixes #462.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@
 //! extern crate time;
 //!
 //! use rusqlite::types::ToSql;
-//! use rusqlite::{Connection, NO_PARAMS};
+//! use rusqlite::{Connection, params};
 //! use time::Timespec;
 //!
 //! #[derive(Debug)]
@@ -27,7 +27,7 @@
 //!                   time_created    TEXT NOT NULL,
 //!                   data            BLOB
 //!                   )",
-//!         NO_PARAMS,
+//!         params![],
 //!     )
 //!     .unwrap();
 //!     let me = Person {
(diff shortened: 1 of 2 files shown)


gwenn:
https://github.com/jgallagher/rusqlite/graphs/contributors => third.
Maybe you should ask @jgallagher to become a maintainer ?

PR #476, opened 2019-02-03
Draft to ease PRAGMA usage (#273 and #265)

gwenn:


est31:
:+1: this is very important because with it you can avoid bobby-tables risky dynamic generation of the query string in your user code.

gwenn:
@est31 and @thomcc Would you mind giving a try ?
Or just a check/code review.
Because I don't want to introduce an unsafe (regarding SQL injection) API.
Thanks.

est31:
@gwenn I'll have a look on Wednesday is that ok?

gwenn:
@est31 Ok.

thomcc:
This seems to work, and the code looks right, but I haven't fuzzed it or anything to give me strong confidence that it's fine.

src/pragma.rs
```diff
@@ -0,0 +1,446 @@
...
+        pragma_name: &str,
+        f: F,
+    ) -> Result<T>
+    where
+        F: FnOnce(Row<'_, '_>) -> Result<T>,
+    {
+        let mut query = Sql::new();
+        query.push_pragma(schema_name, pragma_name)?;
+        let mut stmt = self.prepare(&query)?;
+        let mut rows = stmt.query(NO_PARAMS)?;
```
est31: Why not just self.query?

src/pragma.rs
```diff
@@ -0,0 +1,446 @@
...
+        pragma_name: &str,
+        mut f: F,
+    ) -> Result<()>
+    where
+        F: FnMut(&Row<'_, '_>) -> Result<()>,
+    {
+        let mut query = Sql::new();
+        query.push_pragma(schema_name, pragma_name)?;
+        let mut stmt = self.prepare(&query)?;
+        let mut rows = stmt.query(NO_PARAMS)?;
```
est31: Same here

src/pragma.rs
```diff
@@ -0,0 +1,446 @@
...
+        pragma_name: &str,
+        pragma_value: &dyn ToSql,
+        mut f: F,
+    ) -> Result<()>
+    where
+        F: FnMut(&Row<'_, '_>) -> Result<()>,
+    {
+        let mut sql = Sql::new();
+        sql.push_pragma(schema_name, pragma_name)?;
+        // The argument is may be either in parentheses
```
est31: Remove "is"

est31:
See my comments, otherwise the code looks fine! I agree with @thomcc  , this is better than the status quo. It can be fuzzed once it's in master as well :).

src/pragma.rs
gwenn: I guess you mean `self.query_row`.
Because [`check_readonly`](http://sqlite.org/c3ref/stmt_readonly.html) does not work for PRAGMA.
I would like to reactivate this `check_readonly` without impacting `pragma_query_value`.

PR #483, opened 2019-02-17
Clarify instruction for building with vcpkg

technic:
Mention `VCPKGRS_DYNAMIC=1` environment variable,
and the fact that you need to run `cargo clean` in order to force rebuild.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,8 +132,6 @@ You can adjust this behavior in a number of ways:
   which must be enabled by setting `VCPKGRS_DYNAMIC=1` environment variable before build.
   `vcpkg install sqlite3:x64-windows` will install the required library.
   
-*Note:* to change the way `libsqlite3-sys` is built, you need to run `cargo clean libsqlite3-sys` first.
-  
 ### Binding generation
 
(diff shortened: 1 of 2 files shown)


gwenn:
Would you mind testing the following change:
Add this line:
```rs
if cfg!(all(feature = "vcpkg", target_env = "msvc")) {
    println!("cargo:rerun-if-env-changed= VCPKGRS_DYNAMIC");
}
```
here: https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/build.rs#L133

Add then check that you don't need to `clean` anymore.

technic:
@gwenn 
I can readily test it on windows. How can tell cargo to use local git instead of upstream crates source?

gwenn:
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#testing-a-bugfix

technic:
Awesome, every time I am gladly surprised how rust developers though through all scenarios in advance.
I would check it!

technic:
@gwenn 
Your fix works. Pull request updated, thanks!

gwenn:
Thanks.

PR #487, opened 2019-02-27
Avoid unnecessary copies/allocations when passing strings to sqlite

thomcc:
SQLite allows you to pass the string length instead of providing a nul-terminated string. We were providing the length and a nul-terminated string, which is fairly redundant.

This also makes some integer size tests more robust, by using `c_int::max_value()` instead of `std::i32::MAX` (which in practice should be the same except on systems where `sizeof(int) != 4`). This was done mostly since it was nearby and I needed similar logic for the other size tests.

This adds a memchr dep which avoids a potential performance regression on large SQL statements (the stdlib uses memchr for the same test). I can implement it without the dep if it bothers you, but it's much faster than the natural check.


gwenn:
Thanks.

PR #490, opened 2019-03-05
Add peekable to Rows

RoGryza:
I couldn't find a way to peek at the next row without advancing the `Rows` iterator outside the crate, so I thought I'd submit this PR to expose this functionality.


RoGryza:
I didn't check the open PR's beforehand, seems like https://github.com/jgallagher/rusqlite/pull/478 solves this better. I'm closing this PR

PR #491, opened 2019-03-10
Add optional support for rust-url

thomcc:
As described on the tin.

P.S. It would be nice to get a release in the semi-near future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,12 +36,13 @@ script:
   - cargo test --features trace
   - cargo test --features chrono
   - cargo test --features serde_json
+  - cargo test --features url
   - cargo test --features bundled
   - cargo test --features sqlcipher
   - cargo test --features i128_blob
   - cargo test --features "unlock_notify bundled"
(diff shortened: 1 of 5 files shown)


gwenn:
`FromSql`/`ToSql` for UDT can be implemented by dependent crates. 
So I am not sure we should add this to `rusqlite`.

thomcc:
Users can't provide an implementation for this for Url, either this crate or the url crate has to. Given that the url crate isn't even willing to add the dep on serde (they provide a separate crate for serde support, which works but is a pain), it makes more sense to add it here.

Besides, `rust-url` is one of the most widely used crates on crates.io, and rusqlite already does this for `serde_json` and `chrono`.

gwenn:
I checked on [this page](https://www.postgresql.org/docs/11/datatype.html) and saw nothing about `url` (but `json` and `date/time` types are).
And I guess you can't define an alias to `Url` and use it everywhere.
So fine, let merge this PR.

gwenn:
@thomcc version 0.17 released with many breaking changes...

PR #499, opened 2019-03-25
Allow non-Unicode file names on Unix-like platforms

xfix:
There are still issues on Windows, but hopefully fixing this issue on Unix-platforms-only is fine?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -276,6 +276,13 @@ fn len_as_c_int(len: usize) -> Result<c_int> {
     }
 }
 
+#[cfg(unix)]
+fn path_to_cstring(p: &Path) -> Result<CString> {
+    use std::os::unix::ffi::OsStrExt;
+    Ok(CString::new(p.as_os_str().as_bytes())?)
+}
+
+#[cfg(not(unix))]
 fn path_to_cstring(p: &Path) -> Result<CString> {
     let s = p.to_str().ok_or_else(|| Error::InvalidPath(p.to_owned()))?;
     str_to_cstring(s)
@@ -836,7 +843,17 @@ unsafe fn db_filename(db: *mut ffi::sqlite3) -> Option<PathBuf> {
     if db_filename.is_null() {
         None
     } else {
-        CStr::from_ptr(db_filename).to_str().ok().map(PathBuf::from)
+        let cstr = CStr::from_ptr(db_filename);
+        #[cfg(unix)]
+        {
+            use std::ffi::OsStr;
+            use std::os::unix::ffi::OsStrExt;
+            Some(OsStr::from_bytes(cstr.to_bytes()).into())
+        }
+        #[cfg(not(unix))]
+        {
+            cstr.to_str().ok().map(PathBuf::from)
+        }
     }
 }
 #[cfg(not(feature = "bundled"))]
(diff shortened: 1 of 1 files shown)


gwenn:
On Mac OS X:
```
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:208
   4: <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get
             at src/libstd/panicking.rs:474
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:381
   6: std::panicking::try::do_call
             at src/libstd/panicking.rs:308
   7: <T as core::any::Any>::type_id
             at src/libcore/panicking.rs:85
   8: core::result::unwrap_failed
             at /rustc/82e2f3ec25a316c7536c33a6b6704366b14cbf2a/src/libcore/macros.rs:16
   9: core::result::Result<T,E>::unwrap
             at /rustc/82e2f3ec25a316c7536c33a6b6704366b14cbf2a/src/libcore/result.rs:798
  10: rusqlite::test::test_invalid_unicode_file_names
             at src/lib.rs:958
```
Linux only ?
And I guess we should update `db_filename` accordingly.

xfix:
Thanks for pointing it out. The test is making an assumption about a file system that isn't always true - in particular, it will break if temp directory uses HFS+ filesystem. I will make a change to a test to skip it if filesystem doesn't support filenames like this.

thomcc:
Note that you can also pass a URI filename when opening the file: https://www.sqlite.org/uri.html

xfix:
I'm not sure what URI filenames have to do with anything.

Yes, `SQLITE_OPEN_URI` option exists. You can pass it to `Connection::open_with_flags`. This change shouldn't break it - `&Path` is a superset of `&str`, URLs as far Rust are concerned are valid `Path`s..

thomcc:
URI filenames are relevant because they should let you pass non-unicode characters in the path by percent-encoding them, no?

thomcc:
Actually, you're right, that doesn't help outside of unix.

xfix:
Supporting unpaired surrogates on Windows is interesting. I may look into that later. One thing of note is that `sqlite3_open16` converts UTF-16 into UTF-8, but I have no idea what it does with unpaired surrogates, may want to analyze that later.

There is no fully generic solution to supporting all file names on all operating systems. Rust only exposes native representations for [Windows](https://doc.rust-lang.org/stable/std/os/windows/ffi/trait.OsStrExt.html) and [Unix](https://doc.rust-lang.org/stable/std/os/unix/ffi/trait.OsStrExt.html) platforms.

Not that it really matters, because such files are unlikely to exist - this is an edge case that mostly doesn't matter, but may as well support it :).

thomcc:
Yes, I think it's worth fixing. I'll be using rusqlite in firefox soon and suspect it's possible that some poor user has weird characters in their profile path. That said, in practice this is probably not an issue for other reasons, so I'll defer to @gwenn's judgement here for now.

> Supporting unpaired surrogates on Windows is interesting

It's more than unpaired surrogates you have to worry about, files on windows are allowed to be arbitrary bitpatterns of u16 -- not even remotely valid UTF16. 
Quoting https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN, emphasis mine.

> There is no need to perform any Unicode normalization on path and file name strings for use by the Windows file I/O API functions because the **file system treats path and file names as an opaque sequence of WCHARs.**

I don't think SQLite can open these paths? Not sure, it would work if it just passes the "UTF-16" (heavy air quotes) string directly to the OS and treating it as opaque....

That said, sqlite3_open16 doesn't take flags so I'm not sure we could use it.

xfix:
> It's more than unpaired surrogates you have to worry about, files on windows are allowed to be arbitrary bitpatterns of u16 -- not even remotely valid UTF16.

So it's just unpaired surrogates :). Other than surrogate codepoints, all 16-bit values are valid UTF-16 codepoints.

thomcc:
Ah, fair.

PR #501, opened 2019-03-27
impl ToSql for Box<dyn ToSql>. Fixes #500

### LINKED ISSUE #500: Should `Box<dyn ToSql>` be an instance of `ToSql`?
I had a function like `fn my_value_to_sql(v: &MyValue) -> Box<dyn ToSql>` which would parse a `MyValue` into one of several different types, but all the types implemented `ToSql`. I had to returned this `Box`ed since otherwise the size of the return value wouldn't have been known.

It would have been nice if `Box<dyn ToSql>` was an instance of `ToSql`.

(I'm fairly new to Rust, so it may be that my `Box` approach is misguided, but it's the best I could come up with.)

thomcc:
I think this request was reasonable, and it definitely cannot be done inside a user's crate due to orphan rules.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1615,5 +1615,17 @@ mod test {
             })
             .unwrap();
         }
+        #[test]
+        fn test_dyn_box() {
+            let db = checked_memory_handle();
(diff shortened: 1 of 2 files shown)


gwenn:
Thanks.

PR #511, opened 2019-04-19
Allow specifying both `sqlcipher` and `bundled`.

thomcc:
This is useful because currently, when using `rusqlite` in a Cargo workspace with one crate that uses `sqlcipher` and another that uses `bundled`, a build error will be triggered by an unqualified `cargo
build` (as cargo will use the union of all features enabled by crates in the workspace).

Instead of panicing, this just emits a warning, before (mostly) ignoring that the `bundled` feature was specified. Note: in this configuration, we still use our bundled bindings, to avoid changing `rusqlite` to
handle this edge case (hence 'mostly').

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -4,16 +4,35 @@ use std::path::Path;
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     let out_path = Path::new(&out_dir).join("bindgen.rs");
-    build::main(&out_dir, &out_path);
+    if cfg!(feature = "sqlcipher") {
+        if cfg!(feature = "bundled") {
+            println!(
+                "cargo:warning={}",
+                "Builds with bundled SQLCipher are not supported. Searching for SQLCipher to link against. \
+                 This can lead to issues if your version of SQLCipher is not up to date!");
+        }
+        build_linked::main(&out_dir, &out_path)
+    } else {
+        // This can't be `cfg!` without always requiring our `mod build_bundled` (and thus `cc`)
+        #[cfg(feature = "bundled")]
+        {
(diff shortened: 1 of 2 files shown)


libsqlite3-sys/build.rs
```diff
@@ -111,7 +129,20 @@ mod build {
...
+        if cfg!(feature = "bundled") && !cfg!(feature = "buildtime_bindgen") {
+            // We can only get here if `bundled` and `sqlcipher` were both
+            // specified (and `builtime_bindgen` was not). In order to keep
+            // `rusqlite` relatively clean we hide the fact that `bundled` can
+            // be ignored in some cases, and just use the bundled bindings, even
+            // though the library we found might not match their version.
+            // Ideally we'd perform a version check here, but doing so is rather
+            // tricky, since we might not have access to executables (and
+            // moreover, we might be cross compiling).
+            std::fs::copy("sqlite3/bindgen_bundled_version.rs", out_path)
```
gwenn: Here, `sqlite3/bindgen_bundled_version.rs` is your custom sqlcipher bindgen ?

thomcc: It's the same as the normal `bindgen_bundled_version.rs` present in the repo. It's not exactly ideal to use it in this case, but for us we always know we're using a recent sqlcipher, and in any case where we're not, it's not really that much worse than using a more recent header file with an older sqlite. We emit a warning if you configure both bundled and sqlcipher for this reason, however.

gwenn: Ok!

thomcc:
We'd appreciate a release going out when you next have the opportunity to do so, to get https://github.com/mozilla/application-services/ off of my branch.

PR #516, opened 2019-04-27
Provide the "regexp" fn.

derekdreery:
Using the rust "regex" crate. Behind the "regexp_fn" feature.

The feature is called regexp_fn because
 1. I had to use a different name to regex
 2. It doesn't actually expose any regex api surface to the user, it just makes the `REGEXP` sql function work.

# Motivation

There are 3 reasons I'd rather use this method to get regex than the `function` feature

 1. It's zero configuration (just turn on the feature)
 2. The lifetime of the regex cache is managed by sqlite. When you close the db, the cache is freed so no memory leak.
 3. The API behind the `function` feature is safe, but I think you could cause UB by using them incorrectly. This patch hides all unsafety from the user and prevents UB.

I'm using this patched version in a project of mine, but I thought I should contribute it back in case it's useful to others.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,10 @@ script:
   - cargo test --features sqlcipher
   - cargo test --features i128_blob
   - cargo test --features uuid
+  - cargo test --features regexp_fn
   - cargo test --features "unlock_notify bundled"
   - cargo test --features "array bundled csvtab vtab"
-  - cargo test --features "backup blob chrono csvtab functions hooks limits load_extension serde_json trace url vtab"
-  - cargo test --features "backup blob chrono csvtab functions hooks limits load_extension serde_json trace url vtab buildtime_bindgen"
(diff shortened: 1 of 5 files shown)


gwenn:
The C implementation uses [auxdata](https://www.sqlite.org/cgi/src/artifact/79345bf03496155a).
And a similar Rust implementation is [here](https://github.com/jgallagher/rusqlite/blob/master/src/functions.rs#L554).

> The API behind the `function` feature is safe, but I think you could cause UB by using them incorrectly. This patch hides all unsafety from the user and prevents UB.
Does this statement apply to `regexp_with_auxilliary` ?
If not, I would prefer that we use `regexp_with_auxilliary` behind a feature.

derekdreery:
I think you are correct, that auxdata should be used. I'll look at reworking the PR to use the functions infrasturcture.

UnknownSuperficialNight:
Any update on this?

Issue #514, opened 2019-04-25
Nightly build fails


gwenn:
https://travis-ci.org/jgallagher/rusqlite/jobs/524571469
```
error[E0277]: the type `std::cell::UnsafeCell<std::option::Option<std::boxed::Box<std::cell::RefCell<regex::exec::ProgramCacheInner>>>>` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
   --> src/functions.rs:635:12
    |
635 |         db.create_scalar_function("regexp", 2, true, move |ctx| {
    |            ^^^^^^^^^^^^^^^^^^^^^^ `std::cell::UnsafeCell<std::option::Option<std::boxed::Box<std::cell::RefCell<regex::exec::ProgramCacheInner>>>>` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
    |
    = help: within `(std::string::String, regex::re_unicode::Regex)`, the trait `std::panic::RefUnwindSafe` is not implemented for `std::cell::UnsafeCell<std::option::Option<std::boxed::Box<std::cell::RefCell<regex::exec::ProgramCacheInner>>>>`
    = note: required because it appears within the type `thread_local::CachedThreadLocal<std::cell::RefCell<regex::exec::ProgramCacheInner>>`
    = note: required because it appears within the type `regex::exec::Exec`
    = note: required because it appears within the type `regex::re_unicode::Regex`
    = note: required because it appears within the type `(std::string::String, regex::re_unicode::Regex)`
    = note: required because of the requirements on the impl of `std::panic::UnwindSafe` for `std::ptr::NonNull<(std::string::String, regex::re_unicode::Regex)>`
    = note: required because it appears within the type `hashbrown::raw::RawTable<(std::string::String, regex::re_unicode::Regex)>`
    = note: required because it appears within the type `hashbrown::map::HashMap<std::string::String, regex::re_unicode::Regex, std::collections::hash_map::RandomState>`
    = note: required because it appears within the type `std::collections::HashMap<std::string::String, regex::re_unicode::Regex>`
    = note: required because it appears within the type `[closure@src/functions.rs:635:54: 653:10 cached_regexes:std::collections::HashMap<std::string::String, regex::re_unicode::Regex>]`
```
Seems related to:
https://github.com/rust-lang/regex/issues/576

gwenn:
Fixed in version 0.19.0

PR #526, opened 2019-06-02
Fix nightly build

gwenn:
error[E0277]: the type
`std::cell::UnsafeCell<std::option::Option<std::boxed::Box<std::cell::RefCell<regex::exec::ProgramCacheInner>>>>`
may contain interior mutability and a reference may not be safely
transferrable across a catch_unwind boundary


gwenn:
Fix #514

Issue #524, opened 2019-05-23
Creating a free-standing extension ends in a segfault


lukaslueg:
I figured it should be possible to create a free-standing extension (a `cdylib`) for sqlite using the rusqlite-API. The basic plan is to

* have a small C-shim, which does the extension-initialization via the macros provided by sqlite. This can't be done in Rust, as those macros are not callable via a binding.
* have any sqlite-client load the extension, which calls the shim, which calls some Rust-code
* ???
* profit

I've implemented a small extension which does the above to provide a new aggregate-function. The basic loading mechanism seems to work fine, yet sqlite crashes with a segfault somewhere in [here](https://github.com/lukaslueg/rusqlite_ext_test/blob/6b46f025169032f714c942de692b7c5b3d71ceb3/src/lib.rs#L24) when calling `sqlite3_create_function_v2`. To reproduce, build the lib and use `.load target/debug/libsqlite3_hyperminhash.so` in  the sqlite-cli.

I was unable to debug this any further as of yet. Am I missing something obvious here?

thomcc:
Unfortunately, the SQLite extension API also re-`#defines` a bunch of normal SQLite methods to call through a provided API pointer, which is done in case the SQLite version you compile against and are loaded into do not match (and so that you don't actually have to link against SQLite).

I investigated this before and concluded that it wasn't really practical to use Rusqlite with a SQLite extension, but I also didn't try very hard.

lukaslueg:
I've checked out what the INIT-macros actually do and stopped investigating this any further... Fine with me to close

jrandall:
I can confirm that #532 should address this issue. 

> have a small C-shim, which does the extension-initialization via the macros provided by sqlite. This can't be done in Rust, as those macros are not callable via a binding.
@lukaslueg with #532 I think it should not be necessary to use a C-shim at all. You should be able to use it to compile a rust cdylib that can be directly loaded as a sqlite extension. 

I'll try to explain here how it works and how to use it in case you want to give it a try. 

The way this works is that there are multiple sets of rust bindings generated for sqlite, one for the normal case and others for the case when the rusqlite code is being used as an extension. This basically accomplishes the same thing that including `sqlite3.h` vs `sqlite3ext.h` in your C code would accomplish. 

The normal rusqlite/libsqlite3 bindings generate an empty/opaque `sqlite3_api_routines` struct:
[...]

PR #532, opened 2019-06-15
Support for sqlite vtables as loadable extensions

jrandall:
Implements support to author sqlite loadable extensions (https://sqlite.org/loadext.html) using rusqlite, including the ability to write vtables in Rust and use them from other languages or the `sqlite3` CLI. 

The two main features that configure this are `loadable_extension` or `loadable_extension_embedded`. The first is used when a rust crate is a standalone loadable extension that itself implements an extension entrypoint (i.e. by implementing an `sqlite3_X_init` function ). The embedded version is intended if the rust crate implements functionality that is statically linked into a loadable extension. For example, a rust crate using `loadable_extension_embedded` could implement a vtable but not provide a loadable extension entrypoint itself, but rather expects to be embedded within a loadable extension.

The generated sqlite bindings are different when loaded as an extension or as an embedded extension (note that in C one would need to include `sqlite3ext.h` instead of `sqlite3.h`), so this change requires additional pre-built bindings to be created. Rather than generate them manually, we opted to write a script to generate them using sources downloaded from sqlite.org (see `libsqlite3-sys/bindgen-bindings/generate-bindgen-bindings.sh`). This PR includes prebuilt bindings for seven versions of sqlite for both the normal, extension, and embedded extension modes.


jrandall:
I see that the travis CI checks are failing. This appears to be because the build script now requires `rustfmt` to format the bindings whereas previously they were being formatted by bindgen. The reason for this change is that the extension and embedded extension bindings rely on some code that are added to the bindgen-generated bindings by the build script. 

Can we add `rustfmt` to the travis CI build environment, and is this an acceptable build-time dependency to have?

gwenn:
Why not update [rusqlite/libsqlite3-sys/upgrade.sh](https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/upgrade.sh) ?

jrandall:
> Why not update [rusqlite/libsqlite3-sys/upgrade.sh](https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/upgrade.sh) ?

Because I didn't realize it existed. I did look for documentation on how to generate the pre-built bindings but didn't find anything. I am happy to incorporate my changes into the existing script, though we may want to add some docs (or possibly change the name) to make it clear that it should be used to (re)generate the prebuilt bindgen bindings.

gwenn:
Do you know if your PR fixes or is related to #524 ?

jrandall:
Yes, this should implement a fix to #524 
I've been rather busy lately but will try to get to merging my changes into the existing `upgrade.sh` script as soon as I can.

jrandall:
I have added some notes on how to use this functionality to implement a rust sqlite extension in #524: https://github.com/jgallagher/rusqlite/issues/524#issuecomment-507787350

gwenn:
Many thanks.

phiresky:
Could this get merged at some point? I'm trying to create a sqlite extension so this would be great.

jrandall:
closing in favor of #631

PR #533, opened 2019-06-15
add ability to open sqlite connection with specified vfs

jrandall:
adds two new functions to create a connection:
`open_with_flags_and_vfs()` and `open_in_memory_with_flags_and_vfs()` that allow a speciifc sqlite VFS (https://sqlite.org/vfs.html) to be used. 

The actual pass-through of the vfs is done within InnerConnection. I assume that is intended to be a private interface so I opted to change the signature of the InnerConnection `open_with_flags()` function to accept a third `Option` argument for the vfs. If that is not acceptable it could easily be refactored into a separate function.

### DIFF
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -57,7 +57,7 @@ impl InnerConnection {
         }
     }
 
-    pub fn open_with_flags(c_path: &CString, flags: OpenFlags) -> Result<InnerConnection> {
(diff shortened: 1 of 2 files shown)


gwenn:
Would you mind removing copyright headers ?

jrandall:
I will check on that and get back to you.

jrandall:
On review, we do not think the changes proposed in this PR include any copyrightable contributions, so we have removed the copyright and license notices in the file headers.

PR #537, opened 2019-06-21
Add bundled-windows feature

kpcyrd:
This allows a dependency to use `build_bundled` on windows and `build_linked` everywhere else. This is currently not possible because cargo doesn't support target-specific features and selecting the bundled feature for windows would enable the bundled feature everywhere.

Related: https://github.com/rust-lang/cargo/issues/1197


thomcc:
FWIW I'm pretty sure you can work around the lack of target-specific features by having a target-specific dependency on a dummy crate that does nothing but turn on the feature.

I don't oppose adding this, though.

kpcyrd:
I've considered using a dummy crate, but figured it would be cleaner to add it to libsqlite3-sys instead, especially when packaging the resulting tree for distros.

gwenn:
Thanks.

PR #541, opened 2019-06-25
Include the name of the column in InvalidColumnType errors

thomcc:
We're hitting this in the wild on firefox-ios and knowing that some query has a Null for column 3 is... not particularly actionable. The column name would at least give us a starting place.

I've also made it perform bounds checking on the index in column_name, even though it just panics on out of bounds for now, and refactored that code a little.

It would be nice to get this in a release after it lands, but if that ends up not being possible timing-wise I can just point at my fork.

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -27,8 +27,7 @@ impl Statement<'_> {
         let n = self.column_count();
         let mut cols = Vec::with_capacity(n as usize);
         for i in 0..n {
-            let slice = self.stmt.column_name(i);
-            let s = str::from_utf8(slice.to_bytes()).unwrap();
+            let s = self.column_name(i);
             cols.push(s);
         }
         cols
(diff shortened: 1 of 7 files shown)


src/vtab/mod.rs
```diff
@@ -472,7 +472,7 @@ impl Values<'_> {
             }
             FromSqlError::OutOfRange(i) => Error::IntegralValueOutOfRange(idx, i),
             #[cfg(feature = "i128_blob")]
-            FromSqlError::InvalidI128Size(_) => Error::InvalidColumnType(idx, value.data_type()),
+            FromSqlError::InvalidI128Size(_) => Error::InvalidColumnType(idx, idx.to_string(), value.data_type()),
```
thomcc: It's dodgy to do `idx.to_string()` here for the name, but I'm not sure it particualrly matters. I almost made the name an Option<String> on the error type, but decided against since it should always be present outside of this case, which should be rare. I'd be willing to change it to do that again if it would be preferred though.

src/column.rs
```diff
@@ -40,6 +39,13 @@ impl Statement<'_> {
         self.stmt.column_count()
     }
 
+    pub(crate) fn column_name(&self, col: usize) -> &str {
+        // Just panic if the bounds are wrong for now, we never call this
+        // without checking first.
+        let slice = self.stmt.column_name(col).expect("Column out of bounds");
```
thomcc: The `expect` is here instead of a panic inside the inner stmt's column_name mostly since I debated exposing this publicly, which would have had `column_name` return a `Option<&str>`. I can change that if you want.

gwenn:
I will try to do a release tomorrow.
And I suggest you ask @jgallagher to grant you access.

thomcc:
> And I suggest you ask @jgallagher to grant you access.
Ah sure, I will. I meant to the last time you suggested but forgot.

thomcc:
Thanks a ton for cutting the release @gwenn!

Issue #548, opened 2019-07-20
sqlite3_column_text may return non-utf-8, rusqlite does not allow to handle it


hpk42:
we got a real-life case where the lines at https://github.com/jgallagher/rusqlite/blob/master/src/statement.rs#L625 cause a panic because sqlite3_column_text can indeed contain non-utf8. FWIW the sqlite3 docs  https://www.sqlite.org/version3.html say this: 

> SQLite is not particular about the text it receives and is more than happy to process text strings that are not normalized or even well-formed UTF-8 or UTF-16. Thus, programmers who want to store IS08859 data can do so using the UTF-8 interfaces. As long as no attempts are made to use a UTF-16 collating sequence or SQL function, the byte sequence of the text will not be modified in any way.

In our case we are reading from a database that was created/populated by a non-rust program so we can't change the fact that a text column contains non-utf8.  We'd like to handle the invalid case ourselves -- could the API be amended?  any recommendations on how we could work around the issue?

dignifiedquire:
According to https://www.sqlite.org/capi3ref.html#sqlite3_column_blob `sqlite3_column_text` does return `utf-8` strings

dignifiedquire:
but it might be better to change the code slightly according to the above documentation 

> The safest policy is to invoke these routines in one of the following ways:
>
>    sqlite3_column_text() followed by sqlite3_column_bytes()
>    sqlite3_column_blob() followed by sqlite3_column_bytes()
>    sqlite3_column_text16() followed by sqlite3_column_bytes16()

gwenn:
I guess we need to change:
https://github.com/jgallagher/rusqlite/blob/master/src/types/value_ref.rs#L110
```rust
impl<'a> ValueRef<'a> {
    pub(crate) unsafe fn from_value(value: *mut crate::ffi::sqlite3_value) -> ValueRef<'a> {
```
to:
```rust
impl<'a> ValueRef<'a> {
    pub(crate) unsafe fn from_value(value: *mut crate::ffi::sqlite3_value) -> FromSqlResult<ValueRef<'a>> {
```
Fix:
```rust
                // sqlite3_value_text returns UTF8 data, so our unwrap here should be fine.
                let s = s
                    .to_str()
                    .expect("sqlite3_value_text returned invalid UTF-8");
```
And update callers accordingly.

thomcc:
[...]

PR #549, opened 2019-07-22
Do not assume `sqlite3_column_text` is valid UTF-8.

gwenn:
Fix Statement::value_ref

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -622,10 +622,7 @@ impl Statement<'_> {
                     CStr::from_ptr(text as *const c_char)
                 };
 
-                // sqlite3_column_text returns UTF8 data, so our unwrap here should be fine.
-                let s = s
-                    .to_str()
-                    .expect("sqlite3_column_text returned invalid UTF-8");
+                let s = s.to_bytes();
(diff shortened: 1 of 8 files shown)


gwenn:
See #548

src/statement.rs
```diff
@@ -624,8 +624,7 @@ impl Statement<'_> {
 
                 // sqlite3_column_text returns UTF8 data, so our unwrap here should be fine.
```
thomcc: Stale comment.

gwenn:
> Another approach might be to return the &[u8] and let the caller handle it. This API means that if you have non utf8-text it can't be accessed.

Indeed !

thomcc:
> Indeed !

Concretely, [this code](https://github.com/mozilla/application-services/blob/master/components/places/src/db/db.rs#L273-L277) spends around 10% of its time validating UTF8. if getting a &[u8] out of a ValueRef were possible, we could avoid this. (That said, it's not particularly hot code, in practice IO takes most of the time, but still...)

Using ValueRef (outside of rusqlite's internals) already implies you're worried about perf some anyway.

PR #554, opened 2019-07-28
Handle old versions of visual studio

derekdreery:
On old versions of visual studio, `isnan` isn't supported. This patch tests for the visual studio version, and enables the isnan emulation inside sqlite if necessary


gwenn:
Thanks.

PR #556, opened 2019-08-04
Conversion from FromSqlError to Error

### LINKED ISSUE #555: There's no From for FromSqlError to Error
`query_map` expects to return `rusqlite::Error`, but `ValueRef` returns `FromSqlError`. Because of that, the convenient syntax `.as_str()?` can't be used in the closure. I'd expect these to work nicely together.

kornelski:
`Error` needs column index, but `FromSqlError` doesn't have it. I've fudged it by putting obviously-wrong index value, and then hiding that fact in `Display`.

Ideally the `Error` should use `Option<usize>` for index (and type), but I assume it's better to avoid unnecessary backwards-compat break.

Fixes #555

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,5 @@
 use crate::types::Type;
+use crate::types::FromSqlError;
 use crate::{errmsg_to_string, ffi};
 use std::error;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


kornelski:
The test failure looks like a `vcpkg` installation issue, unlikely to be related to this change.

gwenn:
Thank you.

PR #564, opened 2019-08-22
Make `column_name()` public and forward all column methods in Row and Rows

twistedfall:
Hi,

Having `column_name()` accessible really helps with avoiding allocations. And there was an obvious mismatch between column methods in Statement, Row and Rows.

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -9,14 +9,14 @@ pub struct Column<'stmt> {
     decl_type: Option<&'stmt str>,
 }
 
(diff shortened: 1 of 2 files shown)


gwenn:
Could you please have a look at #568 ?
I made `column_name` returns a Result instead of an option (like `column_index`).
Thanks.

twistedfall:
Looks even better, thank you!

gwenn:
Thanks to you.

Issue #583, opened 2019-10-29
Segmentation fault on `prepare_cached` with an empty query


tomhoule:
With the latest version, the following program segfaults on my machine:

```rust
use rusqlite::*;

fn main() -> Result<()> {
    let conn = Connection::open_in_memory()?;

    conn.prepare_cached("")?;

    Ok(())
}
```

rustqlite version: 0.20.0
sqlite version: 3.30.1-1
OS: Arch Linux

gwenn:
Thanks you.
See PR #584 for a fix.
I am not sure how we should handle this case !

tomhoule:
Should we close this issue, now that the PR is merged? :)

gwenn:
PR is merged but not released yet.
I will try to release a new version this week.

tomhoule:
No hurry from our side, we worked around the issue for now.

gwenn:
Fixed in 0.21

PR #584, opened 2019-10-29
Segmentation fault on `prepare_cached` with an empty query

gwenn:
With an empty query is prepared, sqlite3 returns no error but a null pointer.
And then `sqlite3_sql` returns null.
Which make `CStr::from_ptr` crash.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -135,9 +135,12 @@ impl StatementCache {
 
     // Return a statement to the cache.
     fn cache_stmt(&self, stmt: RawStatement) {
+        if stmt.is_null() {
+            return;
+        }
         let mut cache = self.0.borrow_mut();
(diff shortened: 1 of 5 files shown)


gwenn:
Should we return an error instead of returning an already closed/unusable statement ?

gwenn:
Should fix #583

thomcc:
Returning an error seems better than returning a statement that we'll panic on to me, or am I misinterpreting the alternative?

gwenn:
I will try to do some tests, but I expect no panic.

thomcc:
Ah, I misinterpreted the code then, that's fine, I have no opinion beyond panicking being worse than an error (which is just because the various firefoxes that use rusqlite don't recover from rust panics)

PR #592, opened 2019-11-14
Omit unnecessary code from bundled SQLite build

adamreichold:
Following the explanations from SQLite's [recommended compile-time options](https://www3.sqlite.org/compile.html), this adds a Cargo feature to libsqlite-sys so that the bundled SQLite can be better tailored to the specific needs of rusqlite. The feature is enabled by default and rusqlite does explicitly not request it, so that if any other crate pulls libsqlite-sys into the dependency, no features are removed.

As a first step, this commit
* adds `SQLITE_OMIT_PROGRESS_CALLBACK` as rusqlite does not use `sqlite3_progress_handler`
 * and sets `SQLITE_DEFAULT_MEMSTATUS` to `0` as rusqlite does not use `sqlite3_status`.


adamreichold:
I am not sure, but I suspect that this would imply updating the bundled bindings?

adamreichold:
Oh, it seems the `sqlite3_trace` and `sqlite3_profile` API are used instead of `sqlite3_trace_v2` which was introduced in version 3.14. Is that too new or would a commit to switch to that API be acceptable? Otherwise, I'll remove `SQLITE_OMIT_DEPRECATED` from the compile-time flags.

gwenn:
You can't assume that `rusqlite` is the only crate using `libsqlite3-sys`.
See [here](https://crates.io/crates/libsqlite3-sys/reverse_dependencies).

adamreichold:
> You can't assume that `rusqlite` is the only crate using `libsqlite3-sys`.
> See [here](https://crates.io/crates/libsqlite3-sys/reverse_dependencies).

I certainly agree. But this does affect only the the bundled build. Is this also used outside of rusqlite?

Changes like the default threading mode can also be overwritten at runtime. (I was thinking of following the WAL commit mode recommendation as well?)

So how to continue:
* Shall I survey the above crates of e.g. usage of API like `sqlite3_status`?
* Shall I provide a Cargo feature which can be used to tailor the bundled build to rusqlite's needs?
* Drop this entirely as it is unlikely to get merged in any form?

gwenn:
> But this does affect only the the bundled build. Is this also used outside of rusqlite?

https://github.com/search?q=libsqlite3-sys+bundled+language%3Atoml&type=Code
So it seems unwise to change `-DSQLITE_THREADSAFE=1`.
> Changes like the default threading mode can also be overwritten at runtime.

`rusqlite` must do this in case the `bundled` feature is not used.

The safest options are SQLite default options.

adamreichold:
> rusqlite must do this in case the bundled feature is not used.

Ah yes, `NO_MUTEX` is actually part of the default flags passed rusqlite so this makes sense.

> The safest options are SQLite default options.

So reducing the discussion to those recommended options that only remove code that is not used by rusqlite: How about a feature flag that can be use by rusqlite to request a reduced build e.g. omitting the progress callback since its not exposed anyway? (It would be a default feature of libsqlite-sys that rusqlite would not enable, but it would be present as soon as another dependency requests it via the default features.)

adamreichold:
Changed the PR to only omit the code in question if only rusqlite and no other crate requesting the default features is part of the dependency tree.

adamreichold:
@gwenn Any thoughts on the minimal variant removing memory statistics and progress callbacks if _only_ rusqlite depends on libsqlite-sys using the "inverted" feature flag? Please let me know if you are at all comfortable with diverging from the default build options.

thomcc:
It might be better to add a environment variable code can set to configure cflags more directly. Possibly o

This would probably be more accurate for things like this which are disabling options (in general cargo features work best when they are enabling things).

libsqlite3-sys/build.rs
```diff
@@ -108,6 +108,10 @@ mod build_bundled {
         if cfg!(feature = "session") {
             cfg.flag("-DSQLITE_ENABLE_SESSION");
         }
+        if !cfg!(feature = "bundled-complete") {
```
thomcc, line 111: This name is confusing as it actually makes the bundled build less complete -- it disables things.

adamreichold: Note the `!` in front, things are disabled if this feature is omitted, not if it is enabled (which it is by default).

adamreichold:
> This would probably be more accurate for things like this which are disabling options (in general cargo features work best when they are enabling things).

Which is why the feature flag enables these features and is a default feature, i.e. the are present and are left out if only rusqlite depends on libsqlite-sys as it does not request the default features after this change.

> both disable useful features (that rusqlite doesn't use, but others may). 

Which is why they stay enabled by default and are only disabled if rusqlite is the only thing depending on libsqlite-sys in the whole dependency tree.

> I definitely think these shouldn't be disabled by default though -- (they are on by default for most sqlite builds after all)

They are enabled by default after this change. If only rusqlite is used, they are dead weight as rusqlite does not expose these anyway. Also note that SQLite officially "recommends" to disable these as per the above link - if you don't need them that is and rusqlite itself does not need them.

> You can always use a non-bundled build and statically link to a build explicitly with the SQLITE3_STATIC, SQLITE3_LIB_DIR, SQLITE3_INCLUDE_DIR environment variables (or via .cargo/config in many cases)

This is about making an application that uses only rusqlite faster by default without loosing anything  and also without further setup.

thomcc:
> Also note that SQLite officially "recommends" to disable these as per the above link

It also recommends you turn off thread safety (SQLITE_THREADSAFE=0), recursion limits when parsing (SQLITE_MAX_EXPR_DEPTH=0). These are recommended only if you meet the conditions for them, you shouldn't assume they're all good defaults.

thomcc:
That said the rest of your points are fair. I don't care that much either way here in that case -- I had misread the `!`. It's worth noting that this is still not actually `-complete` though. See conversations in https://github.com/jgallagher/rusqlite/pull/687

adamreichold:
> These are recommended only if you meet the conditions for them, you shouldn't assume they're all good defaults.

Exactly. rusqlite not providing API that exposes memory statistics or progress callbacks implies that rusqlite meets the conditions to disable those. As soon as another crate depends on libsqlite-sys that reasoning is invalid and hence the default feature kicks in.

adamreichold:
> It's worth noting that this is still not actually -complete though.

I am completely not invested in the naming and guess that any other opposite of reduced would be fine. Do you have something in mind?

thomcc:
You can probably do this with https://github.com/rusqlite/rusqlite/commit/a9b700c8413ee8cfdacf504112da8b43e4fad7b2 now, I suspect, no?

adamreichold:
> You can probably do this with a9b700c now, I suspect, no?

Yes, this can be achieved by passing flags manually, but the main motivation here is independent of that possibility, e.g. as written before

> This is about making an application that uses only rusqlite faster by default without loosing anything and also without further setup.

Especially since this encodes the knowledge about which functionality can be omitted as rusqlite does not expose it in the features instead of relying on the library user to figure it out.

thomcc:
As discussed I'm pretty unenthusiastic about this.

I think it's also confusing to have `bundled-full` and `bundled-complete` which are totally unrelated. I'm hesitant to turn it on (it's a breaking change) and I suspect people who want this will would be better served controlling their flags more completely... It's likely many people want to disable `trace`, `fts3` and `fts5`, etc. if they don't use them.

This also, as mentioned, makes it harder for rusqlite to add support for e.g. progress callback in the future.

adamreichold:
> I think it's also confusing to have bundled-full and bundled-complete which are totally unrelated.

As written above, I am very much open to any suggestions for a different name. (I think the "complete" naming even predates the "full" feature.)

> I'm hesitant to turn it on (it's a breaking change)

Could you elaborate where the breakage is? In which scenario will API be removed that is available now?

> It's likely many people want to disable trace, fts3 and fts5, etc. if they don't use them.

The difference is that things like the progress callback add overhead in core SQLite functions even if they are never used whereas the FTS extensions should not have measurable overhead if they are never excercised.

> This also, as mentioned, makes it harder for rusqlite to add support for e.g. progress callback in the future.

Why is that? If rusqlite exposes API that this flag would omit, we can just include that in the "non-complete" libsqlite-sys build as adding API is not a breaking change.

adamreichold:
Considering the lack of enthusiasm and the unintuitive rephrasing of omitting functionality as an additive feature flag, I'll drop this and let the branch predictors sort it out.

thomcc:
Sorry about this, but thanks for being understanding.

PR #594, opened 2019-11-22
Replace deprecated tempdir with tempfile

stevenroose:
`tempdir` has been deprecated for a while and `tempfile` has a new version out that uses a new version of `rand` that doesn't depend on `winapi` anymore.

So this also removes one dependency on the `winapi` crate.  The current version of the `time` crate also depends on `winapi,` but work is being done to replace that version with 0.2 to remove the `winapi` dependency.


gwenn:
Thanks.

PR #595, opened 2019-11-22
Add .editorconfig

stevenroose:
This avoids annoying wrong indentation when people use tabs in their editors (like they are supposed to :speak_no_evil:).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+# see https://editorconfig.org for more options, and setup instructions for yours editor
+
+[*.rs]
+indent_style = space
+indent_size = 4


gwenn:
But only if they use an editor that supports .editorconfig, no ?

stevenroose:
Most do or have plugins. Editorconfig is an attempt at a bare-minimum formatting standard that's very easy to support in any editor.

vn971:
Why is this not merged? o_O
Let's merge this!

(Or write in README that we're in need for contributors (that would also get merge permissions))

gwenn:
Mainly because there is `rustfmt`...

thomcc:
> Or write in README that we're in need for contributors (that would also get merge permissions)

Definitely the issue here isn't that we need someone to click the Merge button for us. It's just not clear what value this adds over rustfmt which (is checked in CI).

I guess I'm not actually opposed to it, though.

vn971:
@gwenn @thomcc editorconfig and rustfmt answer different questions though.
`rustfmt` says how to format well structured Rust code.
Editorconfig tells your editor what should happen when you press Tab or Shift-Tab while editing:
```
fn test() {
    im_forgetting_semilicon_but_that_is_not_relevant()
    <--- cursor here
```

Also, editorconfig is supported by most text editors, even if they don't have rust plugins.

gwenn:
But you don't care of correct indentation because rustc don't and you must use rustfmt anyway !

vn971:
@gwenn humm, interesting... Rust repo currently seems to have editorconfig, but only for clippy: https://github.com/rust-lang/rust/blob/0d97f7a96877a96015d70ece41ad08bb7af12377/src/tools/clippy/.editorconfig

I've asked a question on it here: https://github.com/rust-lang/rust/pull/45002#issuecomment-765226033
Apparently Rust had an .editorconfig, but lost it in a PR that says "Validate miri against the HIR const evaluator #45002". Let's see if anybody will reply on that old thread..

PR #604, opened 2019-12-28
Fix i32 overflow in Connection::busy_timeout

arthurprs:
Found after I tried using `i32::max_value()` as the timeout. Another possibility is to pass i32 max on overflow, that might be better in most cases.


gwenn:
Thanks.

PR #606, opened 2019-12-30
Impl Debug for Row

billop:
https://github.com/jgallagher/rusqlite/issues/572

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -3,7 +3,7 @@ use fallible_streaming_iterator::FallibleStreamingIterator;
 use std::{convert, fmt, result};
 
 use super::{Error, Result, Statement};
-use crate::types::{FromSql, FromSqlError, ValueRef};
+use crate::types::{FromSql, FromSqlError, Value, ValueRef};
 
 /// An handle for the resulting rows of a query.
 pub struct Rows<'stmt> {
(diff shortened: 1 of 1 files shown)


src/row.rs
```diff
@@ -186,6 +186,18 @@ pub struct Row<'stmt> {
     pub(crate) stmt: &'stmt Statement<'stmt>,
 }
 
+impl<'stmt> fmt::Debug for Row<'stmt> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut row = Vec::new();
+        let columns = self.stmt.columns();
+        for (count, c) in columns.iter().enumerate() {
+            // unwrap shouldn't panic since count will always be in range here?
+            row.push((count, c, self.get::<_, Value>(count).unwrap()))
```
thomcc: I think this could still probably panic if you had something like non-utf8 strings in the DB (which Value might have trouble with). Probably better to just return a fmt::Error in that case.

src/row.rs
```diff
@@ -186,6 +186,18 @@ pub struct Row<'stmt> {
     pub(crate) stmt: &'stmt Statement<'stmt>,
 }
 
+impl<'stmt> fmt::Debug for Row<'stmt> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut row = Vec::new();
```
thomcc: This (and going right through Value instead of e.g. ValueRef) is a bit wasteful but I guess it wouldn't matter for a Debug impl anyway.

gwenn:
https://github.com/jgallagher/rusqlite/issues/605#issuecomment-569764259
> Definitely. But we have to be careful in case there is a 1 Gb Blob...
Either print only some bytes or by default, dump only the blob size not its content ?

src/row.rs
hismito: The [`std::fmt`] document says

> Formatting implementations should ensure that they propagate errors from the Formatter (e.g., when calling write!). However, they should never return errors spuriously. That is, a formatting implementation must and may only return an error if the passed-in Formatter returns an error.

It may be surprise if panic when using `dbg!`.

How about using [`String::from_utf8_lossy`] or others for non-utf8 strings?

[`std::fmt`]: https://doc.rust-lang.org/std/fmt/index.html#formatting-traits
[`String::from_utf8_lossy`]: https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy

thomcc: Yes, I agree that we should not panic in a formatter.

evbo:
Hi, would love a way to debug print rows without having to individually `.get` each column when rapid prototyping. Any chance this is getting merged in or is there a better approach we should be using?

thomcc:
I think its reasonable to print out the values. For postgres doing so might be more costly, for us they're probably already in memory anyway.

It's not like `Vec<u8>` only prints out the size and not the contents, even though it might be gigabytes.

evbo:
for those looking for this functionality, within a `#cfg[test]` module you can add a helper utility that achieves basic functionality (haven't added blob support yet):

```rust
#cfg[test]
pub fn print_rows(conn: &mut Connection, sql: &str) {
    let mut stmt = conn.prepare(sql).unwrap();
    println!("{}", stmt.column_names().iter().map(|c| format!("{: >15}", c)).collect::<Vec<String>>().join(", "));
    let col_cnt = stmt.column_count();
    let mut rows = stmt.query([]).unwrap();

    while let Ok(Some(row)) = rows.next() {
      let mut cols: Vec<String> = vec![];
      for c in 0..col_cnt {
        cols.push(match row.get::<_, Value>(c).unwrap() {
          Value::Null => "NULL".to_string(),
          Value::Integer(i64) => i64.to_string(),
          Value::Real(f64) => f64.to_string(),
          Value::Text(string) => string,
          Value::Blob(v) => "vec".to_string(),
        });
      }
      println!("{}", cols.iter().map(|c| format!("{: >15}", c)).collect::<Vec<String>>().join(", "));
    }
  }
```

PR #608, opened 2020-01-08
Allow crates with a direct dep on libsqlite3-sys to know if `sqlcipher` has been enabled

thomcc:
If something else (another crate in your workspace when building with --all, for example) turns sqlcipher on, but isn't part of your build, you'll have issues, as the crypto lib might not get linked.

This way, a dependent build.rs can inspect the env var cargo sets and link the crypto lib if necessary. (... This is a pretty annoying situation to be in, but it doesn't complicate the build.rs really, and I don't really think there's really many other ways if you're in that situation).

In the future it would probably be nice to better support sqlcipher's configuration, but there's only so much env vars + cargo features can manage...

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -206,6 +206,13 @@ mod build_linked {
         if cfg!(all(feature = "vcpkg", target_env = "msvc")) {
             println!("cargo:rerun-if-env-changed=VCPKGRS_DYNAMIC");
         }
+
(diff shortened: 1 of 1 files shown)


thomcc:
@gwenn No rush, but could you publish out a 0.17.1 of libsqlite3-sys with this when you get a chance?

(rusqlite itself should be fine as-is, of course)

gwenn:
Release in progress... #610

PR #613, opened 2020-01-14
Separate the `modern_sqlite` and `bundled` features. 

### LINKED ISSUE #612: `bundled` feature should be split into two features
The bundled feature does a couple things:

1. Causes the build to include the bundled sqlite.
2. Disables certain checks that are irrelevant if we're using the bundled sqlite, mostly around versions, but I haven't done an exhaustive study here.
3. Causes rusqlite to assume it's using a modern version of sqlite, and lets it expose a number of features (among other things).

There are a lot of cases where someone might want the third part enabled, but not the first two.

I think adding a new feature, `modern-sqlite` which just turned those things on, without enabling bundling, would be good. Then `bundled` would depend on it, and most of the `#[cfg(feature = "bundled")]` inside Rusqlite would chage to `#[cfg(feature = "modern-sqlite")]`.  (Cases where the check is actually only relevant for bundling would stay the same -- for example, the places rusqlite bypasses version checks based on `bundled` should stay the way they are).

I think this applies only to `rusqlite` and not `libsqlite3-sys`, but I could be forgetting something.

thomcc:
Fixes #612.

I also got rid of the vtab_v3 feature since AFAICT it's unnecessary if we change the code in the way I did. The old way could cause problems under buildtime_bindgen, as our definition of sqlite3_module might be from >3.26.

BTW what are the version checks protecting against? My understanding is that SQLite's ABI is pretty stable, so it should always be legal to use a newer header to call older code, so long as you only call functions that are present (if you call functions which aren't present, you get a linker error). E.g. it's not like sqlite changes the signatures of existing functions (hence all the `blah_v2`s that exist). Did it do so in older versions?

As it is, this *is* enough for my needs for now, I think, but it's a bit of a footgun to turn on (without also enabling buildtime_bindgen). At least, that's unless you're also going to disable the version detection, because of how aggressive said version detection is.

That said, it's possible I'm not thinking of a case where there's a safety issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ collation = []
 functions = ["libsqlite3-sys/min_sqlite_version_3_7_7"]
 # sqlite3_log: 3.6.23 (2010-03-09)
 trace = ["libsqlite3-sys/min_sqlite_version_3_6_23"]
-bundled = ["libsqlite3-sys/bundled"]
+bundled = ["libsqlite3-sys/bundled", "modern_sqlite"]
 buildtime_bindgen = ["libsqlite3-sys/buildtime_bindgen"]
 limits = []
 hooks = []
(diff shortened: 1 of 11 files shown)


gwenn:
Thanks.

PR #619, opened 2020-01-24
Upgraded the bundled SQLite version to 3.31.0.

KamranMackey:


libsqlite3-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libsqlite3-sys"
-version = "0.17.1"
+version = "0.18.0"
```
thomcc: Why 0.18.0 and not 0.17.2?

KamranMackey: @thomcc From previous pull requests where the SQLite version was updated, it appears that the major version was increased.

KamranMackey: If needed I can always change it to 0.17.2.

gwenn:
I guess you don't have `rustfmt` installed because `libsqlite3-sys/sqlite3/bindgen_bundled_version.rs` doesn't look good.

KamranMackey:
> I guess you don't have `rustfmt` installed because `libsqlite3-sys/sqlite3/bindgen_bundled_version.rs` doesn't look good.

@gwenn I just tried running rustfmt, it had zero effect on the file.

gwenn:
Normally, [rustfmt](https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/build.rs#L335) is performed when the bindings file is generated. Maybe after, it's too late.

KamranMackey:
@gwenn It was indeed run while the bindings were generated then, as I already had rustfmt installed. Plus, it's just a generated file.

gwenn:
@KamranMackey If I do the same:
https://github.com/jgallagher/rusqlite/compare/master...gwenn:3.31?expand=1
>  42  libsqlite3-sys/sqlite3/bindgen_bundled_version.rs
versus
> 10,463  libsqlite3-sys/sqlite3/bindgen_bundled_version.rs

Could you please check:
> git log -p a0d1ca3e1b1d3318f7ae800fe77cb1c8fb2c619c
?

KamranMackey:
@gwenn I fixed it. Turns out it was the line endings.

gwenn:
I guess diffs related to `__builtin_va_list` is just noise.

gwenn:
Thanks.

PR #631, opened 2020-02-09
Support for sqlite vtables as loadable extensions

gwenn:
Try to fix PR #532


gwenn:
`api_routines_stub` feature does not seem to be used anywhere ! 
Am I missing something ?

jrandall:
The `api_routines_stub` feature existed to indicate that the underlying libsqlite3 version was one that included the typedef stub in `sqlite.h`:
```
/*
** CAPI3REF: Loadable Extension Thunk
**
** A pointer to the opaque sqlite3_api_routines structure is passed as
** the third parameter to entry points of [loadable extensions].  This
** structure must be typedefed in order to work around compiler warnings
** on some platforms.
*/
typedef struct sqlite3_api_routines sqlite3_api_routines;
```


It was not referenced from rust code, only serving as a more semantically meaningful alias that required the earliest version of sqlite that we support that includes that stub in the header file (which is `libsqlite3-sys/min_sqlite_version_3_20_0`). sqlite 3.14.0 is actually the earliest version that has that code, but libsqlite3-sys does not have support for that version, so the `api_routines_stub` feature was aliased to the earliest version that do have support for, which is 3.20.0.

gwenn:
Would it possible to generate only one other bindgen file instead of two ?
Because the difference between the two is only:
```
--- sqlite3/bindgen_bundled_version-ext.rs	2020-02-15 11:50:57.736323438 +0000
+++ sqlite3/bindgen_bundled_version-ext-embed.rs	2020-02-15 11:51:14.337342051 +0000
@@ -7895,11 +7895,13 @@
     );
 }

-// bindings were built with (non-embedded) loadable_extension:
-// we define our own sqlite_api static variable and export it
-// to C
-#[no_mangle]
-pub static mut sqlite3_api: *mut sqlite3_api_routines = 0 as *mut sqlite3_api_routines;
+// bindings were built with loadable_extension_embedded:
+// define sqlite3_api as an extern since this code will be embedded
+// within a loadable extension that defines and exports this itself
+extern "C" {
+    #[no_mangle]
+    pub static mut sqlite3_api: *mut sqlite3_api_routines;
+}

 // sqlite3 API wrappers to support loadable extensions (Note: these were generated from build.rs - not by rust-bindgen)

```

gwenn:
I don't know why we cannot use `loadable_extension` with the `bundled` version ?
```rust
            if cfg!(feature = "loadable_extension") {
                panic!("Building a loadable extension bundled is not supported");
            }
```

gwenn:
Any idea of a minimal test ?

jrandall:
> Would it possible to generate only one other bindgen file instead of two ?
> Because the difference between the two is only:
> 
> ```diff
> --- sqlite3/bindgen_bundled_version-ext.rs	2020-02-15 11:50:57.736323438 +0000
> +++ sqlite3/bindgen_bundled_version-ext-embed.rs	2020-02-15 11:51:14.337342051 +0000
> @@ -7895,11 +7895,13 @@
>      );
>  }
> 
> -// bindings were built with (non-embedded) loadable_extension:
> -// we define our own sqlite_api static variable and export it
> -// to C
> -#[no_mangle]
> -pub static mut sqlite3_api: *mut sqlite3_api_routines = 0 as *mut sqlite3_api_routines;
> +// bindings were built with loadable_extension_embedded:
> +// define sqlite3_api as an extern since this code will be embedded
> +// within a loadable extension that defines and exports this itself
> +extern "C" {
> +    #[no_mangle]
> +    pub static mut sqlite3_api: *mut sqlite3_api_routines;
> +}
> 
>  // sqlite3 API wrappers to support loadable extensions (Note: these were generated from build.rs - not by rust-bindgen)
> ```

I guess we could remove this block entirely and hard-code the code in `libsqlite3-sys/src/lib.rs` instead (with the two versions controlled by `#cfg` directives)? I am not certain why I did not do that in the first place, but at the moment I see no reason why that shouldn't work.

jrandall:
> Any idea of a minimal test ?

The best test would be to have an example vtable implementation that is loaded as an extension and then tested via queries. We are actively using (the old version of this code) with dozens of vtables, so I ought to be able to put together an example. Demonstrating the embedded version is a little more work (in our system we have a core extension that is based in C which registers with the sqlite core and then loads the rust vtable extensions (and non-rust vtable extensions).

I think it ought to be possible to have an example in which there is a non-embedded rust-based vtable that loads another (embedded) rust vtable within it, which should demonstrate all of the functionality.

jrandall:
> I don't know why we cannot use `loadable_extension` with the `bundled` version ?
> 
> ```rust
>             if cfg!(feature = "loadable_extension") {
>                 panic!("Building a loadable extension bundled is not supported");
>             }
> ```

This is probably a mistake, apologies - I suspect this dates from before I added the `-ext-embed.rs` versions of bundled libsqlite-sys.

jrandall:
> On what basis have you removed our copyright statement?

Ah, I see - the Genomics plc copyright statement has been moved to the top-level `LICENSE` file (I missed that as the removal was in a separate commit). That is acceptable for us, although it does confuse the matter of which portions of rusqlite include code under our copyright (we didn't want it to seem that we were claiming copyright over the whole thing). 

No problem here if you are happy to include it in the top-level LICENSE.

jrandall:
> > Would it possible to generate only one other bindgen file instead of two ?
> > Because the difference between the two is only:
> > ```diff
> > --- sqlite3/bindgen_bundled_version-ext.rs	2020-02-15 11:50:57.736323438 +0000
> > +++ sqlite3/bindgen_bundled_version-ext-embed.rs	2020-02-15 11:51:14.337342051 +0000
> > @@ -7895,11 +7895,13 @@
> >      );
> >  }
> > 
> > -// bindings were built with (non-embedded) loadable_extension:
> > -// we define our own sqlite_api static variable and export it
> > -// to C
> > -#[no_mangle]
> > -pub static mut sqlite3_api: *mut sqlite3_api_routines = 0 as *mut sqlite3_api_routines;
> > +// bindings were built with loadable_extension_embedded:
> > +// define sqlite3_api as an extern since this code will be embedded
> > +// within a loadable extension that defines and exports this itself
> > +extern "C" {
> > +    #[no_mangle]
> > +    pub static mut sqlite3_api: *mut sqlite3_api_routines;
> > +}
> > 
> >  // sqlite3 API wrappers to support loadable extensions (Note: these were generated from build.rs - not by rust-bindgen)
> > ```
> 
> I guess we could remove this block entirely and hard-code the code in `libsqlite3-sys/src/lib.rs` instead (with the two versions controlled by `#cfg` directives)? I am not certain why I did not do that in the first place, but at the moment I see no reason why that shouldn't work.

I've now implemented the alternative, eliminating the need to generated the separate `-ext-embed.rs` bindings. See https://github.com/gwenn/rusqlite/pull/2

phiresky:
This PR works beautifully for me. I'm using it to create this extension: https://github.com/phiresky/sqlite-zstd

One question I couldn't figure out by reading https://github.com/jgallagher/rusqlite/issues/524#issuecomment-507787350:

How is the best way to test an extension created this way? Basically I usually want to build my extension as a `.so`/`.dll`, but for tests I probably want to directly load it.


I tried disabling the loadable_extension feature for tests and just calling my init function myself by doing this:

```
[dependencies.rusqlite]
git = "https://github.com/Genomicsplc/rusqlite", branch="pr/loadable-extensions"
features=["functions", "blob", "loadable_extension"]

[dev-dependencies.rusqlite]
git = "https://github.com/Genomicsplc/rusqlite", branch="pr/loadable-extensions"
features=["functions", "blob"]
```


But `cargo test` just crashes with `thread 'zstd_fns::tests::sanity' panicked at 'sqlite3_api is null', cargo-target/sqlite-zstd/debug/build/libsqlite3-sys-13edb0ce376f5abc/out/bindgen.rs:6864:9`, and I can't figure out why.

gwenn:
I don't know how to test the `dummy-extension` module ?
Generate another module with `load_extension` and use a relative path while calling [sqlite3_load_extension](https://docs.rs/rusqlite/0.21.0/rusqlite/struct.Connection.html#method.load_extension) ?

gwenn:
I tried to manually check the generated extension but failed:
```sh
rusqlite % /usr/local/opt/sqlite/bin/sqlite3
SQLite version 3.31.1 2020-01-27 19:55:54
sqlite> .load target/debug/libdummy_extension.dylib
sqlite> SELECT * FROM dummy();
zsh: segmentation fault  /usr/local/opt/sqlite/bin/sqlite3
````

**Comment by gwenn** [comment]
A simple scalar function works:
```
sqlite> .load target/debug/libdummy_extension.dylib
sqlite> SELECT half(6);
3.0
```

**Comment by gwenn** [comment]
There is [test-cdylib](https://crates.io/crates/test-cdylib) but we need incompatible features to test the generated extension !
```toml
[dependencies.rusqlite]
path = ".."
default-features = false
features = ["loadable_extension", "functions", "vtab"]
version = "0.21.0"

[dev-dependencies]
test-cdylib = "1"

[dev-dependencies.rusqlite]
path = ".."
default-features = false
features = ["bundled", "load_extension"]
version = "0.21.0"
```
?

LICENSE
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2014 John Gallagher <johnkgallagher@gmail.com>
```
thomcc, line 1: We s

libsqlite3-sys/build.rs
```diff
@@ -369,4 +519,128 @@ mod bindings {
...
+        // add wrapper macro invocation to be appended to the generated bindings
+        let bare_fn = bare_fn_from_type_path(syn_type);
+        let api_fn_output = &bare_fn.output;
+
+        // prepare inputs
+        let mut api_fn_inputs = bare_fn.inputs.clone();
+
+        // handle variadic api functions
+        if bare_fn.variadic.is_some() {
+            // until rust c_variadic support exists, we can't
```
thomcc, line 595: This isn't really safe. Vararg calling convention isn't guaranteed to be the same as the fixed arg calling convention on all platforms / compilers (https://www.gnu.org/software/libc/manual/html_node/Calling-Variadics.html mentions this).

That said I guess it should be fine in practice, but still is worrisome...

libsqlite3-sys/src/lib.rs
```diff
@@ -49,6 +49,26 @@ pub enum Limit {
     SQLITE_LIMIT_WORKER_THREADS = 11,
 }
 
+#[cfg(feature = "loadable_extension_embedded")]
+// bindings were built with loadable_extension_embedded:
+// define sqlite3_api as an extern since this code will be embedded
+// within a loadable extension that defines and exports this itself
+extern "C" {
+    #[no_mangle]
+    pub static mut sqlite3_api: *mut sqlite3_api_routines;
```
thomcc, line 58: This worries me. I'd feel better if we synchronized access to it somehow...

At a minimum I guess we should document that this may be read from without synchronization by any call into rusqlite, and so the safety contract for use is that it must be initialized strictly before any thread could possibly read from it.

This goes for both declarations of `sqlite3_api`.

jrandall: The `sqlite3_api` symbol is the interface provided by sqlite to its run-time loadable extensions. The wrapper functions we generate (see https://github.com/rusqlite/rusqlite/blob/cf658af142399be749bacb0387d54ce836b85f1a/libsqlite3-sys/build.rs#L635) do a test for it not being null. 

When we are compiled as an embedded extension, the contract with the extension we are embedded within is that the `sqlite3_api` symbol is initialised before our entrypoint is called. 

When we are a non-embedded extension, it is our responsibility to set `sqlite3_api` as the first thing we do within our extension entrypoint (e.g. https://github.com/rusqlite/rusqlite/blob/cf658af142399be749bacb0387d54ce836b85f1a/dummy-extension/src/lib.rs#L19). Even so, we initialise the pointer to null in the declaration and the wrapper function protection should protect us until our entrypoint sets it.

I suppose there is no reason to expose the `sqlite3_api` symbol directly outside of libsqlite3-sys -- the only appropriate direct interactions with it are from the generated wrapper functions and to set it (once) at the start of the extension entrypoint. 

Rather than asking the entrypoint to set the symbol directly, we could expose a pub function to initialise the extension, in which case the symbol could be private to the libsqlite3-sys crate, and I suppose we could also make sure that it is only initialised once. 

I.e. something like:
```
use std::sync::Once;
static SQLITE_EXTENSION_INIT: Once = Once::new();

#[cfg(feature = "loadable_extension_embedded")]
pub fn sqlite_extension_init(_p_api: *mut ffi::sqlite3_api_routines) {}

#[cfg(all(
    feature = "loadable_extension",
    not(feature = "loadable_extension_embedded")
))]
pub fn sqlite_extension_init(p_api: *mut ffi::sqlite3_api_routines) {
    SQLITE_EXTENSION_INIT.call_once(|| {
        unsafe {
            ffi::sqlite3_api = p_api;
        }
    });
}
```

Would that alleviate some of your concerns?

thomcc: More or less. Of course, `sqlite_extension_init` should be an `unsafe fn`.

jrandall:
Extension API function wrappers will be generated for all API functions in the sqlite bindings present at build time. It seems there is a trend towards encouraging building rusqlite against a very recent version of sqlite when any non-ancient features are requested (gating them all with the `modern_sqlite` feature - see #639 ). If this is the case, vtable authors will have little choice but to disable the version check in order to make it possible for end users to use their extension in a variety of environments. This presents a serious safety risk as an attempt to call an API function froma  newer version (i.e. from the wrapper at https://github.com/rusqlite/rusqlite/blob/cf658af142399be749bacb0387d54ce836b85f1a/libsqlite3-sys/build.rs#L638) would result in reading past the end of that struct into memory we don't own, and then executing what we find at the target of a pointer we find there.

If we are going to encourage people to disable the version check by making it difficult to build against a minimum required version of sqlite, then we should probably protect the wrappers by including a runtime version check in them (probably using Once) and refusing to attempt to execute any belonging to a version newer than the runtime environment supports.

jrandall:
> Extension API function wrappers will be generated for all API functions in the sqlite bindings present at build time. It seems there is a trend towards encouraging building rusqlite against a very recent version of sqlite when any non-ancient features are requested (gating them all with the `modern_sqlite` feature - see #639 ). If this is the case, vtable authors will have little choice but to disable the version check in order to make it possible for end users to use their extension in a variety of environments. This presents a serious safety risk as an attempt to call an API function froma newer version (i.e. from the wrapper at
> 
> https://github.com/rusqlite/rusqlite/blob/cf658af142399be749bacb0387d54ce836b85f1a/libsqlite3-sys/build.rs#L638
> 
> ) would result in reading past the end of that struct into memory we don't own, and then executing what we find at the target of a pointer we find there.
> If we are going to encourage people to disable the version check by making it difficult to build against a minimum required version of sqlite, then we should probably protect the wrappers by including a runtime version check in them (probably using Once) and refusing to attempt to execute any belonging to a version newer than the runtime environment supports.

Version requirements for the API routines are documented in comments in `sqlite3ext.h` (e.g. https://github.com/rusqlite/rusqlite/blob/master/libsqlite3-sys/sqlite3/sqlite3ext.h#L287) but I am not sure what the best way would be to capture this and pass it through to the wrapper function generator so we can populate a required version test in each of them.

gwenn:
Sorry, last commits (on Jun 13) are wrong: at least changes on bindgen files...

phiresky:
I've updated this PR to work with the current version of rusqlite here: https://github.com/phiresky/rusqlite/tree/gwenntest
Seems to work, just needs recreation of the bindgen files for the non-bundled versions.

gwenn:
> I've updated this PR to work with the current version of rusqlite here: https://github.com/phiresky/rusqlite/tree/gwenntest
> Seems to work, just needs recreation of the bindgen files for the non-bundled versions.

Thanks. Maybe I should close this PR/branch ?

phiresky:
Uh probably not, I'm just a random person, I don't have access to this branch and not enough knowledge to actually champion these changes (e.g. I might have broken Windows builds with this change https://github.com/phiresky/rusqlite/commit/c409018608e22cf54975143a8d39f77577b78551#diff-c1c013e096409a5aea1f123c04254379057791785963eba5ef253fdd686570b8L526 )

phiresky:
I have opened a new PR here: #868

jrandall:
See a note in #910 regarding this PR not adhering to the conditions of the license under which code contributed in #532 was provided: https://github.com/rusqlite/rusqlite/pull/910#issuecomment-791485983

The issue here appears to have arisen out of a commit in which the "Genomics plc" copyright notice was removed and not (as far as I can tell) recreated elsewhere: https://github.com/rusqlite/rusqlite/pull/631/commits/68bea40e806df992d11ff8d12480738fc34bb611

I am sure this was not done maliciously, but these contributions are not owned by me and I only have permission from the owner (my employer) to contribute it under the terms of the license, including the specific copyright notice.

gwenn:
> See a note in #910 regarding this PR not adhering to the conditions of the license under which code contributed in #532 was provided: [#910 (comment)](https://github.com/rusqlite/rusqlite/pull/910#issuecomment-791485983)
> 
> The issue here appears to have arisen out of a commit in which the "Genomics plc" copyright notice was removed and not (as far as I can tell) recreated elsewhere: [68bea40](https://github.com/rusqlite/rusqlite/commit/68bea40e806df992d11ff8d12480738fc34bb611)
> 
> I am sure this was not done maliciously, but these contributions are not owned by me and I only have permission from the owner (my employer) to contribute it under the terms of the license, including the specific copyright notice.
> 
> Please stop distributing this code without including the required copyright notice. The easiest way to do that might be to revert (or rebase out of existence) the commit [68bea40](https://github.com/rusqlite/rusqlite/commit/68bea40e806df992d11ff8d12480738fc34bb611)

See https://github.com/rusqlite/rusqlite/blob/68bea40e806df992d11ff8d12480738fc34bb611/LICENSE
> Copyright (c) 2019 Genomics plc <info@genomicsplc.com>

So [68bea40] explicitly removed `Genomics plc` from `build.rs` but kept it untouched in global LICENCE file (there is no file with copyright header in this repo)...
And https://github.com/rusqlite/rusqlite/pull/631#discussion_r404001452
`Genomics plc` was removed from global LICENSE file when master branch was merged.

jrandall:
> > See a note in #910 regarding this PR not adhering to the conditions of the license under which code contributed in #532 was provided: [#910 (comment)](https://github.com/rusqlite/rusqlite/pull/910#issuecomment-791485983)
> > The issue here appears to have arisen out of a commit in which the "Genomics plc" copyright notice was removed and not (as far as I can tell) recreated elsewhere: [68bea40](https://github.com/rusqlite/rusqlite/commit/68bea40e806df992d11ff8d12480738fc34bb611)
> > I am sure this was not done maliciously, but these contributions are not owned by me and I only have permission from the owner (my employer) to contribute it under the terms of the license, including the specific copyright notice.
> > Please stop distributing this code without including the required copyright notice. The easiest way to do that might be to revert (or rebase out of existence) the commit [68bea40](https://github.com/rusqlite/rusqlite/commit/68bea40e806df992d11ff8d12480738fc34bb611)
> 
> See https://github.com/rusqlite/rusqlite/blob/68bea40e806df992d11ff8d12480738fc34bb611/LICENSE
> 
> > Copyright (c) 2019 Genomics plc [info@genomicsplc.com](mailto:info@genomicsplc.com)
> 
> So [[68bea40](https://github.com/rusqlite/rusqlite/commit/68bea40e806df992d11ff8d12480738fc34bb611)] explicitly removed `Genomics plc` from `build.rs` but kept it untouched in global LICENCE file (there is no file with copyright header in this repo)...
> And [#631 (comment)](https://github.com/rusqlite/rusqlite/pull/631#discussion_r404001452)
> `Genomics plc` was removed from global LICENSE file when master branch was merged.

@gwenn apologies for missing the move of our notice from build.rs to LICENSE had already happened by the time of 68bea40e and we are happy with that move. 

The problem actually occurred as you point out when master was merged: https://github.com/rusqlite/rusqlite/pull/631/commits/cb9e7025a45fe98698285336cd9be84d4ecbc4a5#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7

dummy-extension/src/lib.rs
```diff
@@ -0,0 +1,105 @@
+use std::os::raw::{c_char, c_int};
+
+use rusqlite::ffi;
+use rusqlite::vtab::{
+    eponymous_only_module, sqlite3_vtab, sqlite3_vtab_cursor, Context, IndexInfo, Module, VTab,
+    VTabConnection, VTabCursor, Values,
+};
+use rusqlite::{to_sqlite_error, Connection, Result};
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
```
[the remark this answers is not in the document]
jrandall: discussion moved to https://github.com/rusqlite/rusqlite/pull/910/files#r765317101

libsqlite3-sys/build.rs
jrandall: Discussion moved to https://github.com/rusqlite/rusqlite/pull/910/files#r765322613

libsqlite3-sys/src/lib.rs
jrandall: Discussion moved to https://github.com/rusqlite/rusqlite/pull/910#discussion_r765315952

jrandall:
> 

Discussion moved to https://github.com/rusqlite/rusqlite/pull/910/files#r765327393

jrandall:
> Extension API function wrappers will be generated for all API functions in the sqlite bindings present at build time. It seems there is a trend towards encouraging building rusqlite against a very recent version of sqlite when any non-ancient features are requested (gating them all with the `modern_sqlite` feature - see #639 ). If this is the case, vtable authors will have little choice but to disable the version check in order to make it possible for end users to use their extension in a variety of environments. This presents a serious safety risk as an attempt to call an API function froma newer version (i.e. from the wrapper at
> 
> https://github.com/rusqlite/rusqlite/blob/cf658af142399be749bacb0387d54ce836b85f1a/libsqlite3-sys/build.rs#L638
> 
> ) would result in reading past the end of that struct into memory we don't own, and then executing what we find at the target of a pointer we find there.
> If we are going to encourage people to disable the version check by making it difficult to build against a minimum required version of sqlite, then we should probably protect the wrappers by including a runtime version check in them (probably using Once) and refusing to attempt to execute any belonging to a version newer than the runtime environment supports.

Discussion moved to: https://github.com/rusqlite/rusqlite/pull/910/files#r765327393

PR #636, opened 2020-02-21
make RawStatement public

jrandall:
Make the RawStatement struct public to assist vtable implementations in working with underlying rusqlite statements.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,7 +72,7 @@ use std::sync::{Arc, Mutex};
 
 use crate::cache::StatementCache;
 use crate::inner_connection::{InnerConnection, BYPASS_SQLITE_INIT};
-use crate::raw_statement::RawStatement;
+pub use crate::raw_statement::RawStatement;
 use crate::types::ValueRef;
 
 pub use crate::cache::CachedStatement;


src/lib.rs
```diff
@@ -72,7 +72,7 @@ use std::sync::{Arc, Mutex};
 
 use crate::cache::StatementCache;
 use crate::inner_connection::{InnerConnection, BYPASS_SQLITE_INIT};
-use crate::raw_statement::RawStatement;
+pub use crate::raw_statement::RawStatement;
```
thomcc: I think this should probably be behind a feature flag, if we're going to do it.

thomcc:
My hesitation here is mostly around RawStatement being undocumented. I also don't know all the APIs this would expose are fully sound (hopefully they are, but I don't have infinite confidence in this).

jrandall:
The reason we need this is so that we can implement vtables that use shadow tables (underlying sqlite tables that store data for the vtable), in which we initialise a statement in the `filter` method, convert it into a raw statement and store it on the cursor, then get it back in e.g. the `next` method where we call `step` on the `RawStatement`, and again the the `column` method where we call directly to ffi methods to get the value ref we need.

I.e. we have a cursor struct that looks something like this:
```
#[repr(C)]
pub struct SqliteBasedCursor {
    base: ffi::sqlite3_vtab_cursor,
    underlying_raw_statement: Option<RawStatement>,
    current_row_id: i64,
    at_eof: bool,
}
```
And in `filter`, we do something like:
```
        let mut statement = match conn.prepare(&query)?;
        statement.bind_parameters(values)?;
        let raw_statement: RawStatement = statement.into();
        self.underlying_raw_statement = Some(raw_statement);
```

And then in `next`, we do something like:
```
        let raw_statement = self.underlying_raw_statement.as_ref()?;
        raw_statement.step();
```

And then what we have to do to get the values out of the raw_statement in `column` is kind of horrible so I won't try to summarise it here. 

Ideally, rather than having to do any of this, we would be able to pass a reference to a `Statement` (rather than a `RawStatement`) in the cursor, but we were not able to figure out any way to do that. It would certainly be much easier if we could have access to the full semantics of the `Statement` type, especially when accessing values. 

Can you think of alternative approaches to achieving what we are trying to do with rusqlite vtables? 

We do have it all working with RawStatement (with a locally patched version of rusqlite in which it is public).

thomcc:
Since my initial review, I've done more work in this area and I'm pretty unenthusiastic about exposing RawStatement as-is. If nothing else, I don't expect it's API to be particularly stable, and this crate already has more breaking changes than I'd like.

That said, I'm sympathetic towards your use case -- Next week I should have some time to experiment here. If all else fails, perhaps firming up the API of it could be done. Alternatively, maybe there's a way to make it possible for virtual tables to use non-raw statements somehow (I understand the difficulty you hit here though, but it's possible there are other options...).

thomcc:
I'm working on a custom virtual table, and hit a need for something like this, and started the work of documenting RawStatement and cleaning it up to be exposed.

That said this... doesn't actually help that much, at least my case, where I don't think I can perform binding on the Statement always.

Anyway, some issues with the API as it stands:

1. It's API is fundamentally unsafe (since you use all the guarantees of `Statement`), and the [steps required for safe use](https://gist.github.com/thomcc/e601aac4167cc0d2a555d32a551ec19a) that I came up with ultimately ends with "also make sure you don't do anything that SQLite considers to be undefined behavior"...

2. It's a really incomplete and rough API. You'll pretty much have to use `rusqlite::ffi::` functions to get anything done anyway, but at that point... what are you getting out of RawStatement?

3. This isn't really a problem (and is my fault I believe) but it's also not actually that "raw" anymore.. It still has various caches...

All that said it is probably still safer in some cases than using just playing it fast and loose with direct usage of the `*mut sqlite3_stmt`... But really not by a lot, and it's very subtle, unergonomic, and incomplete as an API.

---

All that said, this is just some thoughts I had when I encountered this case. I'm still going to be thinking about it as I move forward.

I'm holding out hope there's some way this kind of thing can be done safely though... And barring that, figuring out a cleaner API to expose for it seems necessary.

thomcc:
So, we can't use Statement, Blob, etc (e.g. the safe api) as-is in a virtual tables is because of the &Connection reference. But also everything else is rough and I really don't want to expose a ton of dubiously-safe implementation details.

So, if we could remove that reference, and re-add it later, that would be great. The issue is that this is only valid if:

It's the same connection, or at lest the same `*mut ffi::sqlite3` (note that close and reopen could produce the same address, so the addr isn't enough... alone anyway), and nothing's happened to invalidate *our* `*mut ffi::sqlite3_stmt`...

This sounds tricky/impossible, but I think there's a way to make it work:

We can use https://www.sqlite.org/c3ref/next_stmt.html to look for a `*mut ffi::sqlite3_stmt` that's identical to the one in our raw statement. The nice thing about this is that it's definitive. If we're in that list, we're definitely valid and good to go (or memory has been corrupted and we're hosed anyway)

For clarity here's a rough sketch of what I mean, including the tricky big where we verify we came from the connection correctly.

```rust
/// A `Statement<'conn>` that has been "frozen", removing it's lifetime
/// parameter. It can be "thawed" at some point in the future by providing the
/// connection it came from.
pub struct FrozenStatement {
    raw: RawStatement,

    // a `*mut ffi::sqlite3 as usize` we use to detect obvious cases of misuse. 

    // Note that this pointer being the same as `Connection::handle()` is not 
    // quite enough on its own, as it is possible to close a connection,
    // and open a new one, and have the new one with the same address as the old
    // one.
    //
    // It's also not required for us to determine safety. The reason we
    // store is that it's cheap and can detect clearly wrong usage without
    // having to walk the stmt list.
    //
    // (Note that despite how unlikely the "close -> reopen -> same ptr" case
    // seems, exploit authors tend to be skilled at that kind of heap feng-shui)
    db_ptr: usize,
}

impl<'conn> Statement<'conn> {
    /// Produce a FrozenStatement, which is not very useful on its
    /// own, but has no lifetime bound, and can later be "thawed"
    /// using the original Connection.
    pub fn freeze(self) -> FrozenStatement {
        // probably a version that doesn't auto-reset
        // too, but by default this is better.
        self.reset();
        FrozenStatement {
            raw: self.stmt,
            db_ptr: self.conn.handle() as usize,
        }
    }
}


impl FrozenStatement {
    /// Attempt to thaw this statement.
    ///
    /// Note: This would probably actually be a slightly different api...
    /// (for example, maybe it would take `&mut self`, and return 
    /// a `ThawedStatement<'frozen, 'conn>` which stores itself back into `self`
    /// on drop (and derefs into `&Statement<'conn>` ofc).

    pub fn thaw(self, conn: &Connection) -> Result<Statement<'_>, FrozenStatement> {
        unsafe {
            let handle = conn.handle();
            if !self.was_created_by(conn) {
                Err(self)
            } else {
                Ok(Statement {
                    conn,
                    raw: self.raw
                })
            }
        }
    }

    /// Verify that our statement is a valid statement from `conn`.
    fn was_created_by(&self, conn: &Connection) -> bool {
        unsafe {
            let db = conn.handle();
            if db as usize != self.db_ptr || db.is_null() {
                return false;
            }

            let looking_for = self.raw.ptr();

            // Walk the next_stmt chain. This works, but is slow if there
            // are many connection handles.
            let mut stmt = ffi::sqlite3_next_stmt(db, ptr::null_mut());
            while !stmt.is_null() {
                if ptr::eq(stmt, looking_for) {
                    return true;
                }
                stmt = ffi::sqlite3_next_stmt(db, stmt);
            }
            false
        }
    }

    /// Safety: `conn` must have produced `self.raw`, and `self.raw` must still
    /// be valid. 
    pub unsafe fn unchecked_thaw(self, conn: &Connection) -> Statement<'_> {
        assert!(conn.handle() as usize == self.db_ptr);
        Ok(Statement {
            conn,
            raw: self.raw
        })
    }
    // ...
}
```

Problems with this API:

1. `FrozenStatement`s would prevent the DB from actually closing -- it would go into zombie state: https://www.sqlite.org/c3ref/close.html. Probably need to have a story here...

2. I'd really like something applicable to other types with conn references too, like Blob and such. THis is totally specific to statements. That's not a reason to not do it though.

3. It's slow. Walking this list IMO can't be the common case path. Something this slow should be reserved for edge-cases -- but as it is we have to try it every time.

    Ideally, we'd find some other test to use in conjunction with it. The other test would need to be fast, but too conservative (if it returns true, we're certain the operation is safe, and only if it returns false would we fall back to walking the linked list)... the opposite success profile to `conn.handle() == frozen.db_ptr`.

    I suspect that Weak/Arc can probably give us something like this, but need to think more about how exactly it would be put together.

3. If we're going to walk the `sqlite3_next_stmt` linked list, it needs to be only in the edge case, as it's slow.

4. Might need a way to handle thread safety issues caused by Connection's Send-ness. This probably isn't too bad and may not be necessary -- havent thought through it fully.

None of these seem that bad on their own though -- and if those issues can be worked out... I believe it's completely sound, can be used without unsafe code, avoids exposing internals, and (Assuming it just derefs into &Statement) doesn't add too much to the API surface.

jrandall:
I like the concept of freezing and thawing the statement, and I sympathize with wanting to find an API for this that can be made safe. However, the slowness of having to go looking for the statement in the linked list of all prepared statements on the connection is going to be problematic for our use-case, in which would be thawing the statement in every call to `next()` and in many cases in every call to `column()` so that we can get values from it - this has to be a very cheap operation for it to work for us. 

I suppose in practice we would just use the proposed unsafe `unchecked_thaw` to avoid those checks -- because we are in a vtable implementation, we are in a good position to know that the connection remains the same and the statement is on that connection.

Issue #609, opened 2020-01-08
Separation of named/positional params lead to API duplication


thomcc:
For example, all the `_named` (vs the non-`_named`) functions on https://docs.rs/rusqlite/0.21.0/rusqlite/struct.Statement.html. In mozilla/application-services, we even have a extension trait which exists/started mainly as a way to get more methods for `_named` variants (also cached) and such: https://github.com/mozilla/application-services/blob/master/components/support/sql/src/conn_ext.rs

Really this is probably my biggest complaint about rusqlite's API. I don't really have a solution (the ones I've thought of all hit coherence issues or restrict the API in ways I wouldn't want to give up), but it seems like it should have an issue on file, since its a pain for maintenance. It's probably worth investigating what other non-orm (and non-macro) SQL crates do.

gwenn:
As far as know,
* `postgresql` does not support named parameters, only `$NNN`. So you can do `SELECT $1, $1 + 10;` 
* `JDBC` does not support named parameters, only `?`. So you must bind several times the same value.
* `SQLite` does support named parameters but [sqlite3_bind()](https://sqlite.org/c3ref/bind_blob.html) functions support only indexes. `sqlite3_bind_parameter_index()` must be used to translate a symbolic parameter name into its equivalent numeric index.

So maybe we should have only non-`_named` functions and add a single method that do the translation: `&[(&str, &dyn ToSql)]` => `IntoIterator<Item=ToSql>` ?

thomcc:
I didn't realize that this was so uncommon, most of my professional experience has been with SQLite or MySQL (which also has them).

I'm aware that there's a cost, but my experience has definitely been that the readability gains can be substantial, and I've never seen the cost take any time compared to the query that generally follows. So I don't really like having there be a separate function for this...

Along the lines of your suggestion, something like

```rust
trait IntoParams {
    type Param: ToSql;
    // It's likely that this would be better as `Item = (i64, Self::Param)`
    // to avoid needing to reorder for named params...
    type Params: IntoIterator<Item = Self::Param>;
    fn into_params(self, stmt: &Statement) -> Self::Params;
}
```

might work (probably with more Result<T> tho). 

E.g. basically `IntoIterator<Item: ToSql>` but that passes the statement into its ctor, and each function could take `params: P) where P: IntoParams` or something (e.g. users probably would never call the trait directly).

[...]

PR #637, opened 2020-02-21
make bind_parameters() on Statement public

jrandall:
When implementing a vtable based on an underlying rusqlite connection, we have found it helpful in the `filter` implementation to be able to build a Statement by passing a Vec<Values> to bind_parameters() without actually executing the query (yet). We then convert the Statement into a RawStatement (see #636) and store that on the VTabCursor. Only if our `next` implementation is later called do we actuallt execute the query (by calling `step()` on the raw statement).

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -424,7 +424,7 @@ impl Statement<'_> {
         Ok(self.stmt.bind_parameter_index(&c_name))
     }
 
-    fn bind_parameters<P>(&mut self, params: P) -> Result<()>
+    pub fn bind_parameters<P>(&mut self, params: P) -> Result<()>
     where
         P: IntoIterator,
         P::Item: ToSql,


thomcc:
Is there any reason you can see not to do this for bind_parameters_named too? If not, can you do that?

thomcc:
I think we should do #668 instead. It's less ergonomic but allows use in more scenarios. It's also easier to document and seems less error prone.

thomcc:
I'm going to close this since we landed #668. The ergonomics of that are rougher, but I think we'd accept a PR for additional functions with better ergonomics if you're still interested.

jrandall:
Ok, after #668 it appears that we should be able to implement our own version of `bind_parameters` that uses the new `raw_bind_parameter` interface. I'm really not sure what the problem is with making `bind_parameters` public, though. 

The `execute` method and the `query` method are both public and pass the `params` argument it exposes directly through to `bind_parameters` so the interface for `bind_parameters` seems like it is already constrained by those public interfaces and nothing is being gained by keeping `bind_parameters` private. Am I missing something?

I did miss that `bind_parameters_named` even exists - I agree they should both be made public.

thomcc:
The main reason I'm hesitant is that it doesn't seem that useful directly; that is, you couldn't implement the functionality in #668 given a stable bind_parameters, but the other way is true.

I also plan on changing this interface somewhat, since I find the duplication everywhere to be annoying (#609). I also consider the necessity of `NO_PARAMS` to be a bit of a design wart. I have ideas for how to fix this, but need to find the time (Perhaps next week I'll have more time?).

I would have done this for 0.23, but I was not really wanting to do a big API overhaul in that release since it had a number of security fixes.

jrandall:
I agree that we *can* now implement our own bind_parameters using the new `raw_bind_parameter` interface from #668 -- that's great and it is good to have a lower level interface available in case we need it. 

However, I don't really see why we can't expose both a low-level interface and a higher level interface for this. I guess the reasoning is probably that you are planning to do away with the `_named` versions and so this will all be changing anyway -- in that case perhaps the replacement API for executing a statement with either named or positional params could include high-level public functions for binding parameters and executing the statement separately (in addition to a convenience function for doing the full execution in one go)?

thomcc:
Yeah, that seems reasonable to me.

PR #639, opened 2020-02-24
Implement additional vtab functionality

jrandall:
Derives debug on the vtab IndexInfo struct as this aids in debugging vtable implementations.

Implements the set_idx_str, set_idx_flags, col_used, and collation methods on the vtab IndexInfo struct.

Add a close() method to the VTabCursor trait and call it from rust_close() so that the vtable
can clean up any state that it has related to the cursor, which in some cases may include unsafe
raw pointers that have been passed over the FFI boundary and need to be reclaimed to avoid leaking
memory.

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -308,6 +308,7 @@ impl From<u8> for IndexConstraintOp {
 /// [`VTab::best_index`] method.
 ///
 /// (See [SQLite doc](http://sqlite.org/c3ref/index_info.html))
+#[derive(Debug)]
 pub struct IndexInfo(*mut ffi::sqlite3_index_info);
 
 impl IndexInfo {
@@ -367,6 +368,15 @@ impl IndexInfo {
         }
     }
 
+    /// String passed if/when filter() is called on the cursor
+    #[inline]
+    pub fn set_idx_str(&mut self, idx_str: &str) {
+        unsafe {
+            (*self.0).idxStr = alloc(&idx_str.to_string());
+            (*self.0).needToFreeIdxStr = true as i32;
+        }
+    }
+
     /// True if output is already ordered
     #[inline]
     pub fn set_order_by_consumed(&mut self, order_by_consumed: bool) {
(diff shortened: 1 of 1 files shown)


gwenn:
It seems unsustainable to introduce new features / new bindgens each time.
I would prefer keeping only the `modern_sqlite` feature.
And you can now use [alloc](https://github.com/jgallagher/rusqlite/pull/644) instead of `CString` + `sqlite3_malloc`.

jrandall:
The `modern_sqlite` feature is probably enough when linking sqlite at build time, but when we build a run-time loadable extension (pending #631) it seems more important to ensure that we've built against the lowest version that we can work with, since we have much less control over exactly what version the end users happen to try to dynamically load us into. 

It seems likely that relying on `modern_sqlite` would push vtable authors towards calling `bypass_sqlite_version_check()` and then checking `version_number()` themselves and attempting to ensure that they don't call any api routines that don't exist. 

That seems much riskier than making the default work for most people by automatically building against the sqlite3 api version that is the minimum necessary to support their required features so that they don't have to disable safety checks in order to make their extension work.

thomcc:
I think the modern_sqlite feature is to aggressive about the version check. We shouldn't depend on it for features we don't use... It's hard to say the right choice here. In general it's not really a safety issue... as it would just cause linker/loader errors...

An exception to this might be if you're in a loadable extension, at which point it could very much be a safety issue -- I think it could lead you to calling a function pointer in uninitialized memory -- if an attacker controls that somehow then you're in a lot of trouble!

In general the situation for versioning in this lib doesn't make a ton of sense, see also #706.

jrandall:
Any further direction in what should be updated to get this into a mergeable state?

src/vtab/mod.rs
```diff
@@ -331,6 +332,21 @@ impl IndexInfo {
...
     }
 
+    /// String passed if/when filter() is called on the cursor
+    pub fn set_idx_str(&mut self, idx_str: &str) {
+        unsafe {
+            let c_idx_str = CString::new(idx_str).
+                expect("Couldn't create C string (probably because idx_str contains a null (\\0) character)").
+                into_bytes_with_nul();
+
+            let dst = ffi::sqlite3_malloc(c_idx_str.len() as i32) as *mut ::std::os::raw::c_char;
```
gwenn, line 342: See [alloc](https://github.com/rusqlite/rusqlite/blob/bb34f3fec8a8ed71d203a452a147cd2d58348748/src/vtab/mod.rs#L1039) and `SqliteMallocString`

jrandall:
Thanks @gwenn - that has been changed.

jrandall:
Just a note to confirm that this PR is a "small" contribution and for that reason I have permission from my employer to contribute it under the terms of the rusqlite license without specific attribution.

src/vtab/mod.rs
```diff
@@ -554,6 +592,11 @@ pub unsafe trait VTabCursor: Sized {
     /// Return the rowid of row that the cursor is currently pointing at.
     /// (See [SQLite doc](https://sqlite.org/vtab.html#the_xrowid_method))
     fn rowid(&self) -> Result<i64>;
+    /// Allow the cursor to clean up after itself
+    /// (See [SQLite doc](https://sqlite.org/vtab.html#the_xclose_method)
+    fn close(&mut self) -> Result<()> {
```
gwenn, line 597: Why not just use `Drop` ?

src/vtab/mod.rs
```diff
@@ -393,10 +403,38 @@ impl IndexInfo {
...
+    /// Determine the collation for a virtual table constraint
+    /// https://sqlite.org/c3ref/vtab_collation.html
+    /// May only be called during the best_index call
+    #[cfg(feature = "modern_sqlite")] // SQLite >= 3.22.0
+    #[inline]
+    pub fn collation(&self, constraint_idx: usize) -> Result<&str> {
+        use std::ffi::CStr;
+        let constraint_idx_c = constraint_idx as c_int;
+        let collation_c_buf = unsafe { ffi::sqlite3_vtab_collation(self.0, constraint_idx_c) };
+        let collation_c_str = unsafe { CStr::from_ptr(collation_c_buf) };
```
gwenn, line 432: `collation_c_buf` may be null, no ?

PR #640, opened 2020-02-24
don't perform threading mode checks on wasm32

jrandall:
The wasm32 target arch does not support threads, and to compile the sqlite core for wasm32 requires setting `SQLITE_OS_OTHER`, which does not include a mutex implementation, meaning that the checks in `ensure_safe_sqlite_threading_mode` would always fail. 

This PR disables those checks on the wasm32 target_arch. I've prepared it in such a way (i.e. using an otherwise pointless `any`) as to ease merging with the similar checks included in #631 (see https://github.com/jgallagher/rusqlite/pull/631/files#diff-405160620b5b458acad6ae62a46a20fbR375).

### DIFF
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -6,7 +6,7 @@ use std::path::Path;
 use std::ptr;
 use std::str;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex, Once};
+use std::sync::{Arc, Mutex};
 
 use super::ffi;
 use super::{str_for_sqlite, str_to_cstring};
@@ -324,7 +324,7 @@ impl Drop for InnerConnection {
 }
 
 #[cfg(not(feature = "bundled"))]
-static SQLITE_VERSION_CHECK: Once = Once::new();
+static SQLITE_VERSION_CHECK: std::sync::Once = std::sync::Once::new();
 #[cfg(not(feature = "bundled"))]
 pub static BYPASS_VERSION_CHECK: AtomicBool = AtomicBool::new(false);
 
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.

thomcc:
@jrandall How exactly did you compile sqlite for wasm32?

jrandall:
@thomcc we are using this in connection with #631 - I haven't tried building a wasm application directly using rusqlite - we use rusqlite only to build vtables as embedded extensions. With these changes, that can target wasm without a problem.

We compile the sqlite core using clang-9. The sqlite3.c amalgamation builds without issue for wasm if you disable some things and set `SQLITE_OS_OTHER`. If you want it to actually work you then need to link it with some code (which we've written in C) to supply `sqlite3_os_init()` and `sqlite3_os_end()` which registers and unregisters an appropriate VFS (we just register a dummy one that doesn't really do anything, but the sqlite core will still provide access to an in-memory main database without using any vfs). In our use-case we do actually also have another custom VFS that we've built using a combination of C code and rust code (which does use wasm-bindgen) to provide a VFS to access remote files over http using the browser's fetch implementation. 

The preprocessor arguments we pass to clang while building sqlite are: `-DSQLITE_CORE -DSQLITE_INTROSPECTION_PRAGMAS=1 -DSQLITE_OMIT_WAL -DSQLITE_OMIT_POPEN -DSQLITE_TEMP_STORE=2 -DSQLITE_THREADSAFE=1 -DSQLITE_USE_URI=1 -DSQLITE_OS_OTHER ` (some of those are probably specific to our use-case and will not be required in general).

Issue #457, opened 2019-01-09
borderline case: statement parameter indices may have gap(s)


gwenn:
```sql
SELECT * FROM foo where col = ?2;
```
[sqlite3_bind_parameter_count](https://www.sqlite.org/c3ref/bind_parameter_count.html)
> If parameters of the ?NNN form are used, there may be gaps in the list.

thomcc:
I suspect this could be handled by letting the user specify parameters separate from executing the query.

(Which would also allow things like binding some parameters by position and others by name, which is something I've definitely wanted to do before).

PR #641, opened 2020-02-25
Allow to set named parameters that are not used in the query

hce:
I have a situation where a user can write a query, but my program supplies certain parameters they can use in their query. Right now, an error is produced when I try to set parameters the user hasn't referenced in their query. This could be mitigated by first looking up every parameter I try to set and throwing those out that aren't used, but this means additional work.

Not sure if this is the right way to do it or if an additional flag allow_unused_named_parameters should be added to the functions/a new set of functions...

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -452,7 +452,7 @@ impl Statement<'_> {
             if let Some(i) = self.parameter_index(name)? {
                 self.bind_parameter(value, i)?;
             } else {
-                return Err(Error::InvalidParameterName(name.into()));
+                // Parameter is not used in the query; ignore
             }
         }
         Ok(())


thomcc:
Hrm, I'm a fan of the current behavior, as it's caught bugs a number of times for us.

I'm also very much opposed to adding a new set of functions for these -- the duplication in the API between named/positional is one of my biggest pet peeves. I guess a flag wouldn't be the worst thing, but given that (AFAICT) you're the first to request this, it seems rather niche.

So I'm inclined to say that you should check.

hce:
It is a niche use -- for a custom DSL. What would you say to adding a feature in Cargo.toml for this and use conditional compilation statements to enforce/ignore this check?

thomcc:
IMO it's not a good use case for features.

One option would be to make bind_parameters_named public (similar to https://github.com/jgallagher/rusqlite/pull/637), and then for you to bind each parameter separately, checking and ignor the any parameter name errors.

That said honestly I don't think this really saves much work compared to filling in a slice with each named parameter.

hce:
I agree using features for this would probably be misusing the features feature it a bit.

What do you mean by "filling a slice with each named parameter?" If you mean by this to set each parameter individually, that'd only work with a publicly available bind_parameters_named.

thomcc:
> that'd only work with a publicly available bind_parameters_named

I think that would be a better approach, there's also a similar PR up for this, and an open issue that's vaguely related: #457.

hce:
Okay, then let's close this one and I vote for that other PR to be accepted. :-)

hce:
I see that the issue you referenced is over a year old.

I'll go back to using my custom branch of rusqlite, then.

thomcc:
The issue is, the PR is #637, which is under 4 days old.

Issue #645, opened 2020-03-01
Invalid treatment for executes with columns


Folling:
Rusqlite currently returns an error when `execute` returns no row, but has columns.
However, both renaming a table, as well as renaming a column (which ought to be fine usecases for `execute`) return no rows, but have a column in the result set.

I'm uncertain why they do, but it's something I've noticed. It might have been a regression with the recent changes to how renames are propagated throughout the database.

This is with version 0.21.0 of rusqlite, and version 3.31.1 of sqlite.

Folling:
Would you mind explaining why you believe this issue to be invalid?
There is an auxiliary check from within `execute_with_bound_parameters`, which is performed when there is no row in the result set, which checks whether columns are there or not, seemingly unnecessary. If you agree with that statement, that's fine - but please at least elaborate on it.

Ref:
```rs
    fn execute_with_bound_parameters(&mut self) -> Result<usize> {
        self.check_update()?;
        let r = self.stmt.step();
        self.stmt.reset();
        match r {
            ffi::SQLITE_DONE => {
                if self.column_count() == 0 { // <--- why is this check necessary?
                    Ok(self.conn.changes())
                } else {
                    Err(Error::ExecuteReturnedResults)
                }
            }
            ffi::SQLITE_ROW => Err(Error::ExecuteReturnedResults),
            _ => Err(self.conn.decode_result(r).unwrap_err()),
        }
    }
```

gwenn:
This issue was `invalid` because some basic informations were missing like a stacktrace or a minimal example to reproduce the error.
And, as far as I know, you cannot bind parameter in an `ALTER TABLE` or any DDL.
Also, the number of rows impacted is meaningless...
So [execute_batch](https://docs.rs/rusqlite/0.21.0/rusqlite/struct.Connection.html#method.execute_batch) should be used.

I will check why the column count is not zero.
Regards.

gwenn:
Minimal test to reproduce:
```rust
        #[test]
        fn test_alter_table() {
            let db = checked_memory_handle();
            db.execute_batch("CREATE TABLE x(t);").unwrap();
            db.execute("ALTER TABLE x RENAME TO y;", params![]).unwrap();
        }
```

gwenn:
https://bugs.php.net/bug.php?id=78813

gwenn:
```rust
        #[test]
        fn test_alter_table() {
            let db = checked_memory_handle();
[...]

PR #646, opened 2020-03-04
Fix error while executing ALTER statement

gwenn:
`execute_bacth` should be used for DDL.
`execute` should still work except when `extra-check` feature is
activated.


gwenn:
Should fix #645

PR #647, opened 2020-03-08
Add playground metadata for `rusqlite`, hopefully fixing it

thomcc:
Fixes https://github.com/integer32llc/rust-playground/issues/580, hopefully.

I've also turned on all features which don't conflict or cause problems, since most are small anyway and it would be annoying not to have user defined function support.


gwenn:
I didn't know that `rusqlite` was available on this playground !

thomcc:
Ah yeah, I pointed out it's absence in https://github.com/integer32llc/rust-playground/issues/517, but... then never followed up to ensure it worked >_> (or I noticed it didn't a couple times but never had time to look)

Hopefully this makes it usable, it would be very handy.

PR #651, opened 2020-03-24
don't truncate text columns that contain nul bytes

dae:
If a text column in sqlite contains a nul byte, the string returned by rusqlite is silently truncated at the first nul byte, instead of the text either being returned in full or an error raised. This patch alters the code to use the blob interface to retrieve the text, which is nul-safe.

I benched the change, and there doesn't seem to be a significant difference in performance:

```rust
    #[bench]
    fn bench_string_column(b: &mut Bencher) {
        db.execute_batch("create table tbl(val text);").unwrap();
        db.execute("insert into tbl values (?)", params!["a".repeat(256)])
            .unwrap();

        db.query_row("select val from tbl", params![], |row| {
            b.iter(|| {
                let s: String = row.get(0).unwrap();
            });
            Ok(())
        })
        .unwrap()
    }
```

```
with bytes:
test bench::bench_string_column ... bench:         152 ns/iter (+/- 8)

with old string code:
test bench::bench_string_column ... bench:         150 ns/iter (+/- 7)
```

Without this change, it should still be possible to retrieve the full value by casting the text to a blob in the SQL query, but given that the current behaviour can result in silent data corruption when a value is read and then written back, I thought it might be better solved in rusqlite. Thoughts?

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -1,6 +1,5 @@
-use std::ffi::CStr;
 use std::iter::IntoIterator;
-use std::os::raw::{c_char, c_int, c_void};
+use std::os::raw::{c_int, c_void};
 #[cfg(feature = "array")]
 use std::rc::Rc;
 use std::slice::from_raw_parts;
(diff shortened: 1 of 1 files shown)


src/statement.rs
```diff
@@ -682,16 +669,23 @@ impl Statement<'_> {
...
-                    ValueRef::Blob(unsafe { from_raw_parts(blob as *const u8, len as usize) })
+                    unsafe { from_raw_parts(blob as *const u8, len as usize) }
                 } else {
                     // The return value from sqlite3_column_blob() for a zero-length BLOB
                     // is a NULL pointer.
-                    ValueRef::Blob(&[])
+                    &[]
+                };
+
+                if matches!(kind, ffi::SQLITE_TEXT) {
```
thomcc: This is probably clearer as just `kind == ffi::SQLITE_TEXT` tbh.

dae:
Thanks!

gwenn:
@thomcc
I am not sure but I am afraid we have lost the conversion from utf-16 to utf-8 !

PR #652, opened 2020-03-24
Include an example with chrono timestamp.

steadylearner:
I was following the README.md example from the repository. But, I couldn't make it work. So, I made an example with chrono.

The example is CLI CRUD app that people can test the results from SQLite datbases file made from your crate.

I liked your crate and hope this example help others to use it more.

### DIFF
--- a/examples/with_chrono/Cargo.toml
+++ b/examples/with_chrono/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "sqlite_rust"
+version = "0.1.0"
+authors = ["steadylearner <32325099+steadylearner@users.noreply.github.com>"]
+edition = "2018"
+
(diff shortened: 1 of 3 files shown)


thomcc:
I'm not sure this is a good fit for the examples... It's kind of unclear what the point is and has a bunch of messiness in the code. Sorry...

Also normally you just put examples directly as files in `examples/` rather than give them their own crate (you can add dependencies with `dev-dependencies` from the main crate if needed).

steadylearner:
Ok. I included this because the current example in README.md is not working. I will close but hope you edit that to make it more clear.

gwenn:
As far as I know, current example is fine because of [CI](https://github.com/jgallagher/rusqlite/commit/8ddacbb27ca2df00b9ebd22d52fdc54e0e7753cd).

Issue #653, opened 2020-03-24
Update time to 0.2 (and put it under a feature flag)


thomcc:
Split from #625, since that sort of was both about the readme and the fact that `time` was out of date. We should also put `time` under a feature flag at the same time, like with most other external trait implementations.

PR #654, opened 2020-03-24
Remove use of `time` crate in README.md

### LINKED ISSUE #625: Unclear usage of "time" in README example
Minor thing - I copypasted the code from the README to begin playing with the library, and got an error

```
error[E0432]: unresolved import `time`
 --> src/db.rs:2:5
  |
2 | use time::Timespec;
  |     ^^^^ help: a similar path exists: `std::time`
```

It took a tiny bit of poking around to work out it was a thirdparty requirement (mostly as searching for "rust time" is a bit messy)

Not sure if the `std::time` could be used without cluttering up the example - but possibly just a comment on that line or similar would handy?

```rust
use time::Timespec; # https://crates.io/crates/time
```

thomcc:
We'll be making this a feature when we get around to #653, so it's probably not best to have used directly in the readme anyway.

Fixes #625.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,11 @@ an interface similar to [rust-postgres](https://github.com/sfackler/rust-postgre
 
 ```rust
 use rusqlite::{params, Connection, Result};
-use time::Timespec;
 
 #[derive(Debug)]
 struct Person {
     id: i32,
     name: String,
-    time_created: Timespec,
     data: Option<Vec<u8>>,
 }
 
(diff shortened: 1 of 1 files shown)


gwenn:
Maybe we should have put `time` in `dev-dependencies` and keep it as an example of a non-std type.

thomcc:
Users would still have to add it as a dependency, which was part of the confusion in #625

gwenn:
Indeed.

PR #656, opened 2020-03-24
Revert "don't truncate text columns that contain nul bytes"

thomcc:
This reverts commit 004c8f23d4d46a7c3ff331d97311918a67b7fb0c from PR #651


thomcc:
I'll file a fresh issue for the 'truncated text columns with internal nuls' issue.

PR #658, opened 2020-03-24
don't truncate text columns that contain nul bytes, round 2

dae:
Sorry about that regression. This PR contains the previous changes plus an extra patch that should ensure text content is coerced to utf-8 first, by calling bytes() prior to blob():

https://github.com/ankitects/rusqlite/commit/c0c264c1694805e4d03774f9ee9090a74c46d4a0

From the sqlite docs: 
"Type conversions and pointer invalidations might occur in the following cases:

The initial content is a BLOB and sqlite3_column_text() or sqlite3_column_text16() is called. A zero-terminator might need to be added to the string.
The initial content is UTF-8 text and sqlite3_column_bytes16() or sqlite3_column_text16() is called. The content must be converted to UTF-16.
**The initial content is UTF-16 text and sqlite3_column_bytes() or sqlite3_column_text() is called. The content must be converted to UTF-8.**"

Tested the benchmark again, and it did not change.

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -1,6 +1,5 @@
-use std::ffi::CStr;
 use std::iter::IntoIterator;
-use std::os::raw::{c_char, c_int, c_void};
+use std::os::raw::{c_int, c_void};
 #[cfg(feature = "array")]
 use std::rc::Rc;
 use std::slice::from_raw_parts;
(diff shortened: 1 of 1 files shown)


gwenn:
I would prefer #659.

Issue #657, opened 2020-03-24
Text values with internal nuls are truncated when read out from the database


thomcc:
An initial attempt to solve this was #651, however it caused a regression on non-utf8 databases (#655).

gwenn:
Maybe:
1. `sqlite3_column_text`
2. `sqlite3_column_bytes` ()
3. [slice::from_raw_parts](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html)
4. [str::from_utf8](https://doc.rust-lang.org/std/str/fn.from_utf8.html)

From "Using SQLite" book:
> Regardless of the database, you can insert or request text values in either UTF-8 or UTF-16.
> [...]SQLite does not assume input text values are null-terminated.

> To avoid problems, an application should first extract the desired type using a sqlite3_col umn_xxx() function, and then call the appropriate sqlite3_column_bytes() function.

And we should also fix `str_for_sqlite` to allow internal nuls...

dae:
The code in #651 passes the following test  - am I missing something?

```rust
    #[test]
    fn test_utf16() {
        let db = Connection::open_in_memory().unwrap();
        db.pragma_update(None, "encoding", &"utf16").unwrap();
        db.execute_batch("create table tbl(val text);").unwrap();

        let rust_str = "テスト".to_string();
        db.execute("insert into tbl values (?)", params![&rust_str])
            .unwrap();
        let text: Vec<u8> = db
            .query_row("select cast(val as blob) from tbl", NO_PARAMS, |row| {
                row.get(0)
            })
            .unwrap();

        //let utf16_bytes: Vec<_> = rust_str.encode_utf16().collect();
        //println!("{:x?}", utf16_bytes);
        let utf16_bytes_as_u8 = &[198, 48, 185, 48, 200, 48];
        
        assert_eq!(text.as_slice(), utf16_bytes_as_u8);
    }
```

dae:
I can't see any mention of utf-16 in rusqlite's docs, so I'm not sure whether strings are intended to work with utf-16, but the follow code fails both before and after #651:

```rust
let text: String = db
            .query_row("select cast(val as blob) from tbl", NO_PARAMS, |row| {
                row.get(0)
            })
            .unwrap();
```

dae:
Ok, I think this should be addressed now in PR#658

dae:
Should be ok to close this now?

gwenn:
Already fixed, no ?

thomcc:
[...]

PR #659, opened 2020-03-25
Handle text with internal nuls

gwenn:
Fix insertion and selection (#657).


dae:
Re: #658, no objections here to using this instead, though I guess this approach means sqlite needs to allocate an extra byte to store the unused trailing \0.

It looks like Error::NulError is not used anywhere else. As there's been a breaking change in function definitions in master, perhaps it's worth removing the enum case as well?

gwenn:
If you try to remove this [block](https://github.com/jgallagher/rusqlite/blob/master/src/error.rs#L156-L160), you can see that `NulError` is actually used.

dae:
Sorry, that's embarrassing - I just quickly grepped through the sources without reading them carefully.

PR #660, opened 2020-03-28
Impls of ToSql for different generic types (Box, Cow, Rc, Arc).

zero-systems:


thomcc:
Probably worth adding FromSql too (and implementations for Arc<str> / Rc<str>), but that could be done later too.

zero-systems:
I also see there is no implementation for `Path`. I think there must be another implementations for `Path` and different generic types like `Arc`, `Rc` (as `thomcc` wrote)), `Box`, `Cow`(there is only implementation for `Cow<'_, str>`)

gwenn:
LGTM

gwenn:
Thanks.

PR #662, opened 2020-03-31
Relaxed Sized bound on Box/Rc/Arc/Cow for ToSql

zero-systems:
In https://github.com/jgallagher/rusqlite/pull/660 I added support for `Box<T>` where `T: ToSql`.
But as I see it does not work with unsized types. Instead of impl for `Box<dyn ToSql>` above is used.

I tried to put my `Box<str>` in `ToSql::to_sql(**here**)` but it tells me that `str` is unsized.
I think there must be impl `ToSql + ?Sized` for `Box<T>` instead of `T: ToSql`. And I also removed impl for `Box<dyn ToSql>` because it conflicts (and didn't work before with my example) with my new impl.

### DIFF
--- a/src/types/to_sql.rs
+++ b/src/types/to_sql.rs
@@ -101,15 +101,13 @@ impl<T: ToSql + ?Sized> ToSql for Box<T> {
 
 impl<T: ToSql> ToSql for std::rc::Rc<T> {
     fn to_sql(&self) -> Result<ToSqlOutput<'_>> {
-        let derefed: &dyn ToSql = &**self;
-        derefed.to_sql()
+        self.as_ref().to_sql()
     }
 }
 
 impl<T: ToSql> ToSql for std::sync::Arc<T> {
     fn to_sql(&self) -> Result<ToSqlOutput<'_>> {
-        let derefed: &dyn ToSql = &**self;
-        derefed.to_sql()
+        self.as_ref().to_sql()
     }
 }
 
--- a/src/types/to_sql.rs
+++ b/src/types/to_sql.rs
@@ -229,7 +229,23 @@ mod test {
     }
 
     #[test]
-    fn test_box() {
+    fn test_box_dyn() {
+        let s: Box<dyn ToSql> = Box::new("Hello world!");
+        let r = ToSql::to_sql(&s);
+
+        assert!(r.is_ok());
+    }
+
+    #[test]
+    fn test_box_deref() {
+        let s: Box<str> = "Hello world!".into();
+        let r = s.to_sql();
+
+        assert!(r.is_ok());
+    }
+
+    #[test]
+    fn test_box_direct() {
         let s: Box<str> = "Hello world!".into();
         let r = ToSql::to_sql(&s);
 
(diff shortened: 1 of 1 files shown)


src/types/to_sql.rs
```diff
@@ -241,7 +231,7 @@ mod test {
     #[test]
     fn test_box() {
         let s: Box<str> = "Hello world!".into();
-        let r = s.to_sql();
+        let r = ToSql::to_sql(&s);
```
thomcc: I assume this is trying to beat deref coersion? I think it will still happen -- If there were `Box::to_sql` (I think) that would prevent it.

Also a comment that says that this is trying to defeat deref coersion.

src/types/to_sql.rs
```diff
@@ -87,37 +87,27 @@ pub trait ToSql {
...
-impl<T: ToSql> ToSql for Box<T> {
+impl<T: ToSql + ?Sized> ToSql for Box<T> {
     fn to_sql(&self) -> Result<ToSqlOutput<'_>> {
-        let derefed: &dyn ToSql = &**self;
-        derefed.to_sql()
+        self.as_ref().to_sql()
     }
 }
 
 impl<T: ToSql> ToSql for std::rc::Rc<T> {
```
thomcc: Guessing we need this for Arc<T> and Rc<T> too?

src/types/to_sql.rs
```diff
@@ -87,37 +87,27 @@ pub trait ToSql {
     fn to_sql(&self) -> Result<ToSqlOutput<'_>>;
 }
 
-impl ToSql for Box<dyn ToSql> {
```
thomcc: Does this still work covered on the other implementation? If not I don't know if we can take this -- it's useful and common.

src/types/to_sql.rs
thomcc: Actually maybe it would still not use deref coersion here... I'd like the comment either way though.

src/types/to_sql.rs
zero-systems: What do you mean?
`Arc` and `Rc` impls already use
```rust
self.as_ref().to_sql()
```

thomcc: I mean they should be `impl<T: ToSql + ?Sized>`

src/types/to_sql.rs
zero-systems: I think yes, because `?Sized` should work for unsized (dyn types) that implement `ToSql`

src/types/to_sql.rs
thomcc: You marked this as resolved but it's not resolved...

thomcc:
Hmm... So this actually fails some tests if I try it locally (try `cargo test --features "array bundled csvtab series vtab"`)

For whatever reason travis tests are busted (Filing a bug). Can you make https://github.com/jgallagher/rusqlite/blob/master/.travis.yml#L44-L48 pass?

thomcc:
Oh, your patch didn't make these broken...

PR #669, opened 2020-04-06
Document which features are required, and add vtab usage examples

thomcc:
Been meaning to do this for a bit now, it's just kind of tedious. I didn't do this for `modern_sqlite` since it's tricky and generally people should just enable `bundled`...

I did this for anything public that might appear on a page that wouldn't otherwise document the feature requirement:

- Module doc comments
- Module-level items even if they're inside a module that already has it (since they can appear on their own page
- Inherent methods defined on types outside of the current module.

I also added usage examples for `rarray` and `csv` vtab functions -- there was some confusion in https://github.com/jgallagher/rusqlite/issues/661#issuecomment-607007329 (and generally required reading the test code to understand, at least for `rarray`).

Rendered examples:

a: Screen Shot 2020-04-05 at 9 24 44 PM

b: Screen Shot 2020-04-05 at 9 25 26 PM


src/vtab/csvtab.rs
```diff
@@ -1,6 +1,25 @@
-//! CSV Virtual Table.
+//! `feature = "csvtab"` CSV Virtual Table: https://www.sqlite.org/csv.html
```
gwenn, line 1: We are not using the official CSV extension...

src/vtab/series.rs
```diff
@@ -1,4 +1,4 @@
-//! generate series virtual table.
+//! `feature = "series"` Generate series virtual table: https://www.sqlite.org/series.html
```
gwenn, line 1: We are not using the official extension.

src/vtab/csvtab.rs
thomcc: Hm, that's true, but it does say it's a port a few lines down, and the docs for this are better than our docs or the link to the source.

Would you prefer this link be placed with the spot that says it's a port?

thomcc:
> It would be nice if cargo doc automatically injects feature information...

Yeah definitely. I think most crates don't use features as heavily as rusqlite, so we feel it more. This works as a workaround though...

Issue #663, opened 2020-04-01
Travis CI tests don't run on PRs


thomcc:
See https://github.com/jgallagher/rusqlite/pull/662 -- It actually causes (right now) the rarray tests to fail, but we have way better coverage in travis than on appveyor/github.

thomcc:
It looks like these tests have started failing in the meantime (unrelated to that PR, actually).

Example: `cargo test --features "array bundled csvtab series vtab"`.

thomcc:
The failing test is fixed but we should look into a way to regain coverage there on PRs.

@gwenn: When you have time, thoughts?

gwenn:
It seems that PR #660 is the reason why travis fails.
Because everything was fine until PR #660.

thomcc:
That makes some sense, as "`Rc<T>` which implements `ToSql`" became ambiguous. This seems fine to me, as I think generally needing to be explicit in `collect()` calls is normal.

That said, we should still get the coverage back.

gwenn:
@jgallagher Sorry to disturb you but there is something wrong with Travis CI.
For example, I just merged some changes in #631.
Associated build (1508) is visible in https://travis-ci.org/github/jgallagher/rusqlite/pull_requests .
But not on #631 page !

thomcc:
Fixed now, you can see in https://github.com/rusqlite/rusqlite/pull/712.

PR #670, opened 2020-04-06
Improve CI test coverage

thomcc:
This should at least ensure we build things, run all the tests, and test on both linux and windows. It's not as complete as the stuff we were doing in travis.

See #663.

### DIFF
--- a/src/collation.rs
+++ b/src/collation.rs
@@ -118,16 +118,15 @@ impl InnerConnection {
             }
 
             let callback: fn(&Connection, &str) -> Result<()> = mem::transmute(arg1);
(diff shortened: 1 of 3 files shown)


src/collation.rs
```diff
@@ -118,16 +118,15 @@ impl InnerConnection {
             }
 
             let callback: fn(&Connection, &str) -> Result<()> = mem::transmute(arg1);
-            if catch_unwind(|| {
+            let res = catch_unwind(|| {
```
thomcc, line 121: Previously this caused a compiler warning.

PR #675, opened 2020-04-06
Return an error instead of panicing when parameter count is wrong.

### LINKED ISSUE #671: Panics when wrong bind arguments provided
Currently Rusqlite asserts that the bind parameters are correct:

https://github.com/jgallagher/rusqlite/blob/master/src/statement.rs#L440

I understand the logic here - it's a programming error after all - but this unfortunately is not ideal for my use case. 

I have a Python app with a plugin ecosystem, and which exposes DB access to plugin authors. Recently I switched out the old Python DB API to use Rusqlite under the hood, as part of a move towards using more Rust in the codebase. This mostly works fine - if the add-on author makes a mistake in their SQL, the prepare statement will return a Rusqlite error which I can propagate back to Python. But if they provide SQL that doesn't match the number of arguments, the app crashes.

Would you be interested in accepting a patch that converts this assert to an error instead? If so, please let me know whether you'd like an existing error to be reused, or whether I should create a new one. If not, is there another way it could be solved that you'd be willing to accept?

thomcc:
Fixes #671

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -104,6 +104,10 @@ pub enum Error {
 
     /// Error when the SQL contains multiple statements.
     MultipleStatement,
+    /// Error when the number of bound parameters does not match the number of
+    /// parameters in the query. The first `usize` is how many parameters were
+    /// given, the 2nd is how many were expected.
+    InvalidParameterCount(usize, usize),
 }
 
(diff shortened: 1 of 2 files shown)


dae:
Thanks Thom!

PR #677, opened 2020-04-06
Run clippy on all-targets. Also, use --workspace instead of --all

thomcc:
The changes in #670 failed to lint tests and such.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,7 +61,7 @@ jobs:
           RUSTFLAGS: -D warnings
         with:
           command: clippy
-          args: --features 'array backup blob bundled chrono collation csvtab extra_check functions hooks i128_blob limits load_extension serde_json series trace url vtab_v3 window' -- -D warnings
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           RUSTFLAGS: -D warnings
         with:
           command: clippy
-          args: --features 'array backup blob bundled chrono collation csvtab extra_check functions hooks i128_blob limits load_extension serde_json series trace url vtab_v3 window' -- -D warnings
+          args: --all-targets --all --features 'array backup blob bundled chrono collation csvtab extra_check functions hooks i128_blob limits load_extension serde_json series trace url vtab_v3 window' -- -D warnings
```
gwenn, line 64:
```
--all
Deprecated alias for --workspace.
```

thomcc:
Ah, right. I'll fix in #687?

thomcc:
Worth noting that `cargo fmt` doesn't seem to support `--workspace` (only `--all`), but I'll use it for clippy though.

PR #679, opened 2020-04-06
Prepare 0.22.0-beta.0 (prerelease/unstable/testing build)

thomcc:
I'm going to do a beta/prerelease of 0.22.0 so I can ensure I've worked all the build issues out with moz-central integration.

Might do another even later today if needed...


thomcc:
Note that you'd have to add `rusqlite = "0.22.0-beta"` to depend on this. It cannot happen by accident.

PR #687, opened 2020-04-07
Improve CI and testing/dev workflow by cleaning up how features are used.

thomcc:
- Make `bundled` feature work with session by ensuring we define all flags we care about in the bundled bindgen. This requires another (point) libsqlite3-sys release.
- Remove `unstable` feature by moving us to use https://crates.io/crates/bencher
- Removed vtab_v3 feature which wasn't used anyway.
- Add `bundled-full`, which is basically bundled + the set of features we can turn on with causing bad interactions.
- Use this in github CI and document it.

Also use `actions-rs` more consistently in github CI for better error messages.


libsqlite3-sys/upgrade.sh
```diff
@@ -17,7 +17,7 @@ export SQLITE3_INCLUDE_DIR=$SQLITE3_LIB_DIR
 cargo update
 # Just to make sure there is only one bindgen.rs file in target dir
 find $SCRIPT_DIR/../target -type f -name bindgen.rs -exec rm {} \;
-cargo build --features "buildtime_bindgen" --no-default-features
+cargo build --features "buildtime_bindgen session unlock_notify preupdate_hook" --no-default-features
```
gwenn, line 20: I would prefer not have `session unlock_notify preupdate_hook` activated by default (like SQLite).

thomcc: The only one of these that actually matters for the build is `session`, so the other two are fine.

That said, without this `cargo build --features "session bundled"` fails to compile. Having these declarations in the bindings doesn't actually mean you need SQLite to have session support -- it works fine if sqlite is compiled without session support (you just get a linker error if you call one of these functions).

That said, if you won't change your mind it I could also output both prebuilt bindings with and without session, and choose which one to include in the build.rs. This seems kind of pointless to me, but would work.

thomcc: Eh, I'll just make two binding files. It might be confusing the other way for people who depend on libsqlite3-sys directly

thomcc: Okay, there are two binding files now. That's kind of annoying but could be worse.

README.md
```diff
@@ -166,6 +166,22 @@ enabled if you turn this on, as otherwise you'll need to keep the version of
...
+Rusqlite has many features, and many of them impact the build configuration in
+incompatible ways. This is unfortunate, and makes testing changes hard.
+
+To help here: you generally should ensure that you run tests/lint for both
+`--features bundled` and `--features bundled-full`
+
+### Checklist
+
+- Run `cargo fmt` to ensure your Rust code is correctly formatted.
+- Ensure `cargo clippy --all-targets --all --features bundled-full` passes without warnings.
```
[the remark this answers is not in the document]
thomcc: Thanks, still need to get that into muscle memory.

thomcc:
What's your hesitation for?

This makes working on rusqlite much less of a hassle than it is currently.

Without this, inorder to make sure I haven't broke anything before submitting a PR I have to remember more things to run. This simplifies that greatly -- just use bundled-full feature.

thomcc:
Building bindgen is slow and so builds that need that are slower. It also requires LLVM to be installed, which can be a pain.

It's also just kind of janky to have a feature that causes compilation failures if you turn it on -- it took me bit to figure out i needed to enable buildtime_bindgen.

The downside here seems pretty minor (A single additional file that gets updated by a script).

thomcc:
I guess without that it's still an improvement. I don't understand your reluctance though.

gwenn:
I may be wrong but if we start adding non default option, we will be asked to add more and the [list](https://sqlite.org/compile.html) is quite long.

thomcc:
> I may be wrong but if we start adding non default option, we will be asked to add more and the [list](https://sqlite.org/compile.html) is quite long.

I guess. This seems different to me because we already have a ton of code in place to support the `session` feature.

It also makes it harder test/build the code using the session feature (this is the thing I care about more, to be honest).

I'll think more about this and open a follow up for discussion. Maybe there are other ways we could simplify the build/testing story instead.

PR #695, opened 2020-04-10
Upgrade bundled sqlite to 3.31.1

thomcc:
No rush on getting a release out with this.

The major bump is because of the change in #694 -- Technically it's breaking even though there was likely no non-broken use of that code.


gwenn:
Just update `libsqlite3-sys` version in parent `Cargo.toml`, no ?

thomcc:
Sorry, hadn't noticed the failures -- silly mistake

PR #696, opened 2020-04-11
Fix ValueRef::from_value

gwenn:


thomcc:
Tested locally, this seems great.

PR #701, opened 2020-04-12
Fix soundess issues with aux API

thomcc:
~~I think this might still be broken. Probably should replace with an easier to use API too.~~

Should work now. Really annoying.

- Auxdata downcasting not previously reliable.
- Context can delete auxdata somewhat unpredictably (on any call to set_auxdata). Hence refcounting.
- Connection is Send so auxdata needs to be Sync.

There are a few crates that could help with this that I might look at later. I should also add tests that cover this case now.

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -109,7 +109,6 @@ unsafe extern "C" fn free_boxed_value<T>(p: *mut c_void) {
 pub struct Context<'a> {
     ctx: *mut sqlite3_context,
     args: &'a [*mut sqlite3_value],
-    // conn: PhantomData<&'conn mut Connection>,
 }
 
 impl Context<'_> {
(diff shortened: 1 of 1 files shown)


gwenn:
> After each call to sqlite3_set_auxdata(C,N,P,X) where X is not NULL, SQLite will invoke the destructor function X with parameter P exactly once, when the metadata is discarded.

thomcc:
The biggest issue is:
```
ctx.set_aux(0, "foo".to_string());
let x = ctx.get_aux(0).unwrap();
ctx.set_aux(0, "bar".to_string());
// `x` is still usable but has been freed
```

gwenn:
Why not just hide `set_aux` and make SQLite "function" return `Result<(T, Option<Aux>)>` ?

thomcc:
You're allowed to have auxdata for each parameter, so you'd need to return a list or something. Also there are a bunch of open questions.

But mostly, the vast majority of functions don't care about or need to use auxdata, and this way it's only a breaking change for users of aux data.

I'll see if I can remove the need for Box<Arc> if it's bothering you a lot.

gwenn:
I just want to implement/enforce what the SQLite documentation says:
> sqlite3_set_auxdata() should be called near the end of the function implementation and the function implementation should not make any use of P after sqlite3_set_auxdata() has been called.

PR #704, opened 2020-04-14
Fix vtab::Module lifetime

### LINKED ISSUE #703: Connection::create_module - lifetime issue
The `Module` reference must outlive the `Connection`.
But it seems not enforce by rust.
Changing signature does not help:
```rust
    pub fn create_module<'a, 'b: 'a, T: VTab>(
        &'a self,
        module_name: &str,
        module: &'b Module<T>,
        aux: Option<T::Aux>,
    ) -> Result<()> {
```

thomcc:
We should always be able to get static lifetimes for these. I've modified the existing functions (read_only_module, eponymous_only_module) so that they return &'static modules. If this is a problem in the future we could always `Box::leak(Box::new(the_mod))`, but I doubt it will be.

This now means we no longer need lazy_static to get static module references, although someone who wants to do that for some reason still can (it would look a little different).

I also removed the version numbers. The version number is of the structure's layout -- e.g. how many fields it has.

Allowing users to pass in a version number could allow specifying a version with more fields than it should have, and having sqlite call a function that doesn't exist.

Module is always either V2 or V3 in the code here, so we just pass in V2 since we don't use xShadowName.

Fixes #703.

### DIFF
--- a/src/vtab/array.rs
+++ b/src/vtab/array.rs
@@ -71,7 +71,7 @@ struct ArrayTab {
     base: ffi::sqlite3_vtab,
 }
 
-impl VTab for ArrayTab {
+unsafe impl VTab for ArrayTab {
     type Aux = ();
     type Cursor = ArrayTabCursor;
 
@@ -149,7 +149,7 @@ impl ArrayTabCursor {
         }
     }
 }
-impl VTabCursor for ArrayTabCursor {
+unsafe impl VTabCursor for ArrayTabCursor {
     fn filter(&mut self, idx_num: c_int, _idx_str: Option<&str>, args: &Values<'_>) -> Result<()> {
         if idx_num > 0 {
             self.ptr = args.get_array(0)?;
--- a/src/vtab/array.rs
+++ b/src/vtab/array.rs
@@ -33,8 +33,8 @@ use std::rc::Rc;
 use crate::ffi;
 use crate::types::{ToSql, ToSqlOutput, Value};
 use crate::vtab::{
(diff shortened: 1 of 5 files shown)


thomcc:
:( VTab and VTabCursor are both unsound. Will think about this.

Whole crate needs safety audit...

src/vtab/array.rs
```diff
@@ -57,11 +57,7 @@ impl ToSql for Array {
 /// `feature = "array"` Register the "rarray" module.
 pub fn load_module(conn: &Connection) -> Result<()> {
     let aux: Option<()> = None;
-    conn.create_module("rarray", &ARRAY_MODULE, aux)
-}
-
-lazy_static::lazy_static! {
-    static ref ARRAY_MODULE: Module<ArrayTab> = eponymous_only_module::<ArrayTab>(1);
+    conn.create_module("rarray", eponymous_only_module::<ArrayTab>(), aux)
```
gwenn, line 60: I don't understand why we can't reuse the same Module anymore !

thomcc: It is the same, that function will always return the same pointer. There's no reason to use lazy_static anymore, it would be redundant. I can keep it if you want, though.

You can still do non-static module by Box::leak(Box::new(module)) too.

thomcc:
> I will check if there is any non static module in SQLite sources.

Every `sqlite3_module` in sqlite's sources is either `static` or `static const`.

Long term I don't think this is how we want the API to be though, but short term this is to get a fix out for the soundness issues...

PR #705, opened 2020-04-14
Optimize named params and other cstring uses

thomcc:
Add new utility type SmallCString wrapping SmallVec. Use it for cases where rust `&str` is going to C.
Cache named params indices in `RawStatement.


thomcc:
> Maybe there is already an existing crate for SmallCString ?

Looked a bit. Was going to write one, when I do or if someone beats me I'll switch this over.

PR #707, opened 2020-04-14
Avoid relying on sqlite to justify str::from_utf8_unchecked

thomcc:
> SQLite is not particular about the text it receives and is more than happy to process text strings that are not normalized or even well-formed UTF-8 or UTF-16. Thus, programmers who want to store IS08859 data can do so using the UTF-8 interfaces. As long as no attempts are made to use a UTF-16 collating sequence or SQL function, the byte sequence of the text will not be modified in any way.

In practice this can also happen from database corruption.

### DIFF
--- a/src/collation.rs
+++ b/src/collation.rs
@@ -53,20 +53,18 @@ impl InnerConnection {
         where
             C: Fn(&str, &str) -> Ordering,
         {
-            use std::str;
-
             let r = catch_unwind(|| {
                 let boxed_f: *mut C = arg1 as *mut C;
                 assert!(!boxed_f.is_null(), "Internal error - null function pointer");
                 let s1 = {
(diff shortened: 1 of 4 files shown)


gwenn:
I would have thought that the statement doesn't applies to identifiers !

thomcc:
Not all of these are identifiers, but the ones that are are likely to be small enough such that overhead here doesn't matter.

That said I'd be pretty surprised if the corruption case didn't apply to them, at least when they're table / column names.

gwenn:
I mean that, for me, SQLite identifiers (database, collation, column, constraint, index, table, trigger, view names) should be valid utf-8 string.

thomcc:
What happens if the disc corrupts? We see this often in Firefox.

gwenn:
Ok, your first comment is just misleading.

thomcc:
Sorry.

PR #708, opened 2020-04-14
Fix format string error in rusqlite::trace::log

thomcc:
:/


thomcc:
No, CString::new NUL-terminates the string as well as checks it for internal NULs.

gwenn:
Sorry ! I see now why we need to pass a dummy format string: in case the message contains `%s` or any other format.

thomcc:
Yes. Also %n is particularly bad since it can be used to write to places on the stack, like the return address of the function!

gwenn:
Thanks for yours explanations.

Issue #706, opened 2020-04-14
Consider updating SQLite min (and default) supported version


thomcc:
Right now it's 3.6.8. This is from 2009! We also offer 4 different pregenerated versions of headers up through 3.7.7... which is from 2011.

We should pick a version that makes our lives easier to support. @gwenn, do you have any thoughts or requirements here? Anybody else?

Worth noting that we could do away with the headers, and always use one which is up to date, and avoid some build complexity. Rusqlite would need to avoid calling functions that aren't present, but it already does that. If you don't call an extern "C" function, it isn't linked. SQLite is extremely ABI stable, so this is safe. `bindgen` is just paranoid about libraries less well-behaved.

gwenn:
https://github.com/jgallagher/rusqlite/issues/293#issuecomment-326995795

thomcc:
In https://github.com/launchbadge/sqlx/issues/191 it's been noted that support back to v3.8.11 is desirable.

In that issue I dumped some of my thoughts on this subject which I'll reproduce here for ease of referencing.

> I've meant to reach out to you about this, and do a write up, but we have an open issue in the rusqlite repository around the minimum/default supported version: https://github.com/rusqlite/rusqlite/issues/706. (FWIW there's no rule that says that rusqlite and libsqlite3-sys need to have the same values here, although it would certainly be convenient for me if they did).
> 
> That said it's been a rough year in general and I haven't found the time to do so yet. That said, if you have opinions here they'd be valuable. Even the knowledge that you care about 3.8.11+ is very helpful.
> 
> In general I think the way libsqlite3-sys / rusqlite handles versions to be suboptimal. In an ideal world, there would be only a few versions that control underlying features, and then we'd have feature flags for them. `clang-sys` does this like https://github.com/KyleMayes/clang-sys/blob/master/Cargo.toml#L22-L42 and is fairly okay as a result, although you can see the fractures of that approach starting to not scale very well.
> 
> However, the world we live in is one in which there are very very many versions we would have to do this way, far more than clang-sys has to deal with...

Anyway, diesel is another quite large user of libsqlite3-sys in the ecosystem, so: @sgrif / @weiznich do you have any thoughts on this topic? Even "I'd prefer no change to the status quo for libsqlite3-sys" is useful feedback. No rush at all here, just trying to take stock of the requirements/options here.

mehcode:
> In launchbadge/sqlx#191 it's been noted that support back to v3.8.11 is desirable.

[...]

PR #719, opened 2020-04-16
Add a wrapper for strings allocated on sqlite heap

thomcc:
Except for `from_raw` (and `from_raw_nonnull`), this is a safe API that's hard to misuse.

I'm pretty sure that returning a &CStr and then freeing it is UB also -- Things behind & aren't allowed to change. Hopefully nothing would ever catch this.

This also can't panic, so it's a much safer choice for a use case like `vtab::alloc` function, which I've used it in.

### DIFF
--- a/src/util/sqlite_string.rs
+++ b/src/util/sqlite_string.rs
@@ -1,3 +1,10 @@
+// This is used when either vtab or modern-sqlite is on. Different methods are
+// used in each feature. Avoid having to track this for each function. We will
+// still warn for anything that's not used by either, though.
+#![cfg_attr(
+    not(all(feature = "vtab", feature = "modern-sqlite")),
+    allow(dead_code)
+)]
 use crate::ffi;
 use std::marker::PhantomData;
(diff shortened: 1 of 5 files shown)


src/util/sqlite_string.rs
```diff
@@ -0,0 +1,234 @@
...
+        let src_ptr: *const c_char = bytes.as_ptr().cast();
+        let src_len = bytes.len();
+        let maybe_len_plus_1 = s.len().checked_add(1).and_then(|v| c_int::try_from(v).ok());
+        unsafe {
+            let res_ptr = maybe_len_plus_1
+                .and_then(|len_to_alloc| {
+                    // `>` because we added 1.
+                    debug_assert!(len_to_alloc > 0);
+                    debug_assert_eq!((len_to_alloc - 1) as usize, src_len);
+                    NonNull::new(ffi::sqlite3_malloc(len_to_alloc) as *mut c_char)
```
thomcc, line 123: `sqlite3_malloc64` will clean this up after #706.

PR #720, opened 2020-04-16
Improve statement cache

thomcc:
Arc<str> has a little overhead (really way less than people think...) but far less than cloning the string every time. This way we never need to clone it.

### DIFF
--- a/src/raw_statement.rs
+++ b/src/raw_statement.rs
@@ -9,32 +9,40 @@ use std::ptr;
 
 // Private newtype for raw sqlite3_stmts that finalize themselves when dropped.
 #[derive(Debug)]
-pub struct RawStatement(*mut ffi::sqlite3_stmt, bool, crate::util::ParamIndexCache);
+pub struct RawStatement {
+    ptr: *mut ffi::sqlite3_stmt,
+    tail: bool,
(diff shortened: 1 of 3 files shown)


src/raw_statement.rs
```diff
@@ -77,21 +112,23 @@ impl RawStatement {
...
 
     pub fn reset(&self) -> c_int {
-        unsafe { ffi::sqlite3_reset(self.0) }
+        unsafe { ffi::sqlite3_reset(self.ptr) }
     }
 
     pub fn bind_parameter_count(&self) -> usize {
-        unsafe { ffi::sqlite3_bind_parameter_count(self.0) as usize }
+        crate::util::get_cached(&self.bind_parameter_count, || unsafe {
+            ffi::sqlite3_bind_parameter_count(self.ptr) as usize
```
gwenn, line 125: https://github.com/sqlite/sqlite/blob/master/src/vdbeapi.c#L1584

thomcc: Ah fine. Not worth it then.

thomcc:
> `sqlite3_bind_parameter_count` has zero cost, no !

It's more expensive if you're dynamically linked but I see your point and removed that bit.

The statement cache is a bigger deal.

PR #732, opened 2020-05-13
impl TryFrom<&Row<'_>> for (...)

Kixunil:
This change implements `TryFrom<&Row>` for tuples up to 16 fields. This
is a convenience function that can be used to map rows more easily.

The change includes tests for 1-tuple, 2-tuple and 16-tuple.

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -334,3 +334,199 @@ impl RowIndex for &'_ str {
         stmt.column_index(*self)
     }
 }
+
+macro_rules! tuple_try_from_row {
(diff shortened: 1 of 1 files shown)


Kixunil:
Those clippy failures are wrong, because avoiding closure is impossible there. Looks like rustc can't unify those types. Should I explicitly whitelist those clippy lints or do something else?

Kixunil:
Damn, that commit message was supposed to contain `#[allow...]`, didn't realize `#` is a comment :laughing: Not a huge issue I think.

PR #734, opened 2020-05-16
Rustfmt

gwenn:


gwenn:
Pretty sure:
https://github.com/rust-lang/rustfmt/blob/master/Configurations.md
> To enable unstable options, set unstable_features = true in rustfmt.toml

But you may prefer to keep the default settings ?

thomcc:
> But you may prefer to keep the default settings ?

I don't really have a strong preference. If you like these settings better they're fine.

gwenn:
> Hm, sure. Are all of those flags stable?

I was wrong: most of this flags are unstable...

thomcc:
Yeah... most of the useful rustfmt flags are unstable :(

That said, I still would prefer not depending on any unstable rustfmt flags. When I've worked with projects that have, it's always been a hassle to get things set up right with my editor.

gwenn:
You are right.

PR #742, opened 2020-05-24
Hide IndexConstraintIter and OrderByIter

gwenn:
Returns `impl Iterator` directly

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -565,7 +565,7 @@ impl Values<'_> {
     }
 
     /// Turns `Values` into an iterator.
-    pub fn iter(&self) -> ValueIter<'_> {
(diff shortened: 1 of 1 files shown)


thomcc:
Hmm, can you elaborate on why? The old code would allow someone to store the iterator somewhere. This also prevents us from implementing e.g. DoubleEndedIterator (or anything else implemented by the underlying slice iterator) without it being a breaking change.

gwenn:
I just want to hide some implementation details.

thomcc:
I kind of feel like we should be returning the slices from these functions anyway, e.g. 

```rust
pub fn order_bys(&self) -> &[OrderBy<'_>] { ... }
```

That avoids exposing the iterators directly and is more flexible (the user can easily turn them into an iterator).

I'm not sure if there's a reason not to do this though.

gwenn:
You are right.

Issue #729, opened 2020-05-03
Intermittent memory leak in test deny_single_threaded_sqlite_config


thomcc:
Example failing CI run (asan test) https://github.com/rusqlite/rusqlite/runs/639694804

Probably not important.

Edit: note I had which test was failing wrong first.

PR #749, opened 2020-05-29
little speedup for bundled sqlite3.c

Dushistov:
Usage of `localtime_r` may be little faster then mutex sqlite3 + `localtime`.


Dushistov:
The ASAN pass failure looks like unrelated to this PR.

thomcc:
Yeah, the asan failure is https://github.com/rusqlite/rusqlite/issues/729

PR #752, opened 2020-05-31
adding ability to work with sqlite compiled with -DSQLITE_OMIT_DECLTYPE

### LINKED ISSUE #751: feature request: add feature flag to link with sqlite builded with -DSQLITE_OMIT_DECLTYPE
I have to link with sqlite3 that part of big C++ library,
and sqlite3 is build with `-DSQLITE_OMIT_DECLTYPE`,
with such explanation:

> Disable returning the declared column type from query (reduces memory usage)

Quick hack to comment out:
```
    pub fn column_decltype(&self, idx: usize) -> Option<&CStr> {
        unsafe {
            let decltype = ffi::sqlite3_column_decltype(self.ptr, idx as c_int);
            if decltype.is_null() {
                None
            } else {
                Some(CStr::from_ptr(decltype))
            }
        }
    }
```

and all usage of it, show that rusqlite "core" not depend on this feature,
so it is easy enough feature, but may be I missed something?

Dushistov:
fix #751

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ in_gecko = ["modern_sqlite", "libsqlite3-sys/in_gecko"]
 bundled-windows = ["libsqlite3-sys/bundled-windows"]
 # Build bundled sqlite with -fsanitize=address
 with-asan = ["libsqlite3-sys/with-asan"]
+column_decltype = []
 
 # Helper feature for enabling both `bundled` and most non-build-related optional
 # features or dependencies. This is useful for running tests / clippy / etc. New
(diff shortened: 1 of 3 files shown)


Dushistov:
@thomcc 

May be I should add new feature to:
```
          args: '--features "bundled-full session buildtime_bindgen"'
```
into github actions/codecov ?

thomcc:
Oh, no, you should add it to the bundled-full though!

Cargo.toml
```diff
@@ -86,6 +87,7 @@ bundled-full = [
     "uuid",
     "vtab",
     "window",
+    "column_decltype",
```
thomcc, line 90: Nit: these are in alphabetical order.

gwenn:
Thanks.

PR #753, opened 2020-06-01
VTabCursor lifetime should be bound to VTab lifetime

gwenn:
```c
struct sqlite3_vtab_cursor {
  sqlite3_vtab *pVtab;      /* Virtual table of this cursor */
```
It seems that we need to introduce a lifetime on `VTab` trait
to express such constraint:
https://users.rust-lang.org/t/associated-type-with-lifetime-bound/7256
But there may be another/new way to do it.

### DIFF
--- a/src/vtab/array.rs
+++ b/src/vtab/array.rs
@@ -27,6 +27,7 @@
 //! ```
 
 use std::default::Default;
+use std::marker::PhantomData;
 use std::os::raw::{c_char, c_int, c_void};
 use std::rc::Rc;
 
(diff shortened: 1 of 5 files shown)


thomcc:
Hmm... with a bit more working with it, I'm not actually sure this helps -- Even with this change, you can still write a VTabCursor who doesn't have a lifetime bound to the VTab.

This does allow you to write one, but there might be other ways to manage that, an it comes at a somewhat annoying cost (lifetime-parameterized traits are *very* unergonomic)

Basically, my concern here is that this doesn't actually make the API more safe for most users, but it is a large breaking change and ergonomic regression for anybody using VTab, but I guess I worry that the benefit is too small compared to the cost here, given that we aren't getting complaints AFAIK. Note that the traits are unsafe anyway, so this is not exactly a requirement for soundness

I guess I'll think more. There are a bunch of good ideas in `ext/misc/vtshim`. A lot of it is unnecessary for rust though, given that it's intended for GCed langs I gather.

The other option might be to use proc-macros to generate code, instead of using a trait.... Ultimately the way `sqlite3_module` works is similar to com-style APIs if you squint some...

Issue #725, opened 2020-04-23
Codecov considers patches that reduce coverage as failures


thomcc:
Need to figure out how to turn this off

PR #758, opened 2020-06-05
Fix intermittent CI errors

### LINKED ISSUE #729: Intermittent memory leak in test deny_single_threaded_sqlite_config

thomcc:
This is an attempt to work around rust-lang/rust#59125.

Fixes #729

Fingers crossed.


thomcc:
I'm confident it fixes #725, and believe it *should* fix #729, but it's hard to be sure (it's passed a few times in a row so I'll land it and hope for the best -- can always reopen 729 if it happens again).

PR #763, opened 2020-06-07
Don't implement Into<RawStatement> for Statement

thomcc:
Our usage of this function is sound, and RawStatement is entirely internal at the moment so it doesn't really matter, but this should be unsafe as RawStatement loses the protection we get from the `&'conn` reference.

### DIFF
--- a/src/raw_statement.rs
+++ b/src/raw_statement.rs
@@ -153,7 +153,7 @@ impl RawStatement {
         r
     }
 
-    #[cfg(feature = "modern_sqlite")] // 3.7.4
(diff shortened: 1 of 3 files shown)


src/raw_statement.rs
```diff
@@ -153,7 +153,7 @@ impl RawStatement {
         r
     }
 
-    #[cfg(feature = "modern_sqlite")] // 3.7.4
+    #[cfg(all(feature = "extra_check", feature = "modern_sqlite"))] // 3.7.4
```
thomcc, line 156: These are unused code depending on the feature, but the trait impl hid it. I guess the linter couldn't tell it was impossible to invoke that trait as it was effectively sealed due to referring to an inaccessible type.

Issue #767, opened 2020-06-24
"unlock_notify" support does not cover all needed cases - execute_batch, transaction creation


hlavaatch:
I have noticed that "unlock_notify" support does not cover transaction creation.

Transaction creation uses `InnerConnection::execute_batch` which does not support unlock_notify properly.

As a result, in Deferred mode, first write on transaction will fail with 6, "database table is locked",
and in Immediate mode transaction creation will fail with 262, "database table is locked"
instead of waiting for lock to be unlocked.

`execute_batch` needs to be fixed, maybe by not using `sqlite3_exec` directly but splitting sql on ; and using existing prepare / execute support that does have execute_batch implemented properly.

hlavaatch:
Here is a small program that demonstrates the issue,
[repro.zip](https://github.com/rusqlite/rusqlite/files/4825232/repro.zip)

gwenn:
See partial fix: #768.
To fix `execute_batch`, we need to handle SQL tail.

hlavaatch:
The partial fix above is effective when starting transaction in the correct `TransactionBehavior::Immediate` mode, thank you.

In the `Deferred` / default mode the test app still fails, but it is for different reason of deadlock as sqlite starts read only transaction when encountering SELECT first, and then tries to upgrade it to write transaction on INSERT, while other transaction has the read lock already, which is the correct behavior (this test transaction requires the Immediate mode to work properly).

Looking forward for this fix to bubble up to a public release :) 👍

thomcc:
> Looking forward for this fix to bubble up to a public release :) 👍

This is probably waiting for #771 and probably #769 (which I need to do, first). As well as anything else that comes in before then... I'd assume a couple weeks?

If you need it urgently a release can probably be cut sooner, though.

hlavaatch:
Checking in after ~5 weeks...
Could you make a minor release with the fix please?
I'm about to release my project and having an unreleased dependency is looking bad...

thomcc:
Ah, sorry, rough month. Will look into cutting a patch release ahead of all the big changes.

thomcc:
FYI I was hit by the layoffs at mozilla recently, and thus have my hands slightly full for a bit as I figure out where I stand, and so I don't think I'll be able to get a release out soon. 

@gwenn might be able to, but they also might be busy too.

Sorry about this, everybody.

gwenn:
I will try on friday or saturday.

darosior:
[...]

PR #768, opened 2020-06-24
Replace execute_batch usages by execute

gwenn:
Because execute_batch does not support unlock_notify

### DIFF
--- a/src/pragma.rs
+++ b/src/pragma.rs
@@ -256,7 +256,8 @@ impl Connection {
         // The two syntaxes yield identical results.
         sql.push_equal_sign();
         sql.push_value(pragma_value)?;
-        self.execute_batch(&sql)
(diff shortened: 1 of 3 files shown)


gwenn:
> This looks good but I think we still should (eventually -- doesn't have to be in this PR) do what's suggested in [#767 (comment)](https://github.com/rusqlite/rusqlite/issues/767#issue-644531761), and make execute_batch support unlock_notify.

To do so, we need to handle SQL tail which is not trivial:
https://github.com/gwenn/gosqlite/blob/b61ec314aff4f2d35d940417fafd93ac29a26269/sqlite.go#L336

thomcc:
Yeah, I see. I might look into it in a week or two (unless you grab it first) -- we see a lot of SQLITE_BUSY errors in practice which I don't *think* this could cause, but IDK, would be nice to remove any doubt of that.

PR #770, opened 2020-06-25
Miscellaneous cleanup

dubiousjim:
See individual commit messages

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -197,7 +197,7 @@ mod build_linked {
     }
 
     fn find_link_mode() -> &'static str {
-        // If the user specifies SQLITE_STATIC (or SQLCIPHER_STATIC), do static
+        // If the user specifies SQLITE3_STATIC (or SQLCIPHER_STATIC), do static
         // linking, unless it's explicitly set to 0.
         match &env::var(format!("{}_STATIC", env_prefix())) {
             Ok(v) if v != "0" => "static",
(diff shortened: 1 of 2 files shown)


libsqlite3-sys/build.rs
```diff
@@ -127,6 +127,19 @@ mod build_bundled {
         }
         println!("cargo:rerun-if-env-changed=SQLITE_MAX_EXPR_DEPTH");
 
+        if let Ok(extras) = env::var("LIBSQLITE3_FLAGS") {
```
thomcc, line 130: Thinking more about this, we need to limit some flags perhaps. Certain runtime checks are disabled for bundled builds (since we know the specific configuration). I think we still check threadsafety, but am unsure. If we do not, we should at least issue an error here if they try to disable it at compile time.

gwenn: > Thinking more about this, we need to limit some flags perhaps. Certain runtime checks are disabled for bundled builds (since we know the specific configuration). I think we still check threadsafety, but am unsure. If we do not, we should at least issue an error here if they try to disable it at compile time.

Only `rusqlite` requires threadsafety. Crates which depend on `libsqlite3-sys` directly may not have such constraint.

thomcc: Oh, good point!

gwenn:
> This looks fine to me. I think @gwenn and I have diagreed about `session` before so maybe he has stronger opinions, though.

I disagreed to activate `session` while generating default bundled bindgen.

thomcc:
> I disagreed to activate session while generating default bundled bindgen.

Fair enough!

Issue #765, opened 2020-06-11
Bundled sqlcipher?


dubiousjim:
I've been experimenting with a fork of this repository where I enable building a bundled version of `sqlcipher`, optionally also bundling in an `openssl`/`libressl` libcrypto. Would you be interested in, or in principle open to a PR with this functionality?

Some related issues: #219, #558.

thomcc:
Hrmm... In principal, yes with caveats.

There are some issues around bundling crypto code being dubious in general though. Statically linking openssl/libressl is ofc a thing that happens, but it causes major problems on linuxes and such where packagers really want crypto to be dynamically linked.

That said, that's not really the important thing to me, as terrible as that sounds.

Quite selfishly, I'd have to request that it not break how I use rusqlite + sqlcipher at work. Specifically, on Firefox for Android and Firefox for iOS we use SQLCipher, linked with a patched build of NSS. It's actually pretty fragile, and I'm hoping that that would get better if more things moved into here. 

That said, an overview of the setup that I'd like to still have a story for:

1. Large workspace, some packages have feature = "sqlcipher", some feature = "bundled". Various ways of turning this workspace into shared objects. Tragically, cargo is eager to combine these two features -- in which the current build.rs for libsqlite3-sys chooses sqlcipher, which is good.

2. We use a specific local build of SQLcipher. I think we have a few custom patches, (checking again it seems possible they've all landed upstream by now, perhaps)

3.

dubiousjim:
My changes *default to* dynamically linking against `libcrypto`, they just also provide an option for bundling that in too (if one is bundling `libsqlcipher`).

I also have some custom patches for SQLCipher. I was wondering how to integrate them into a workflow that this repo could plausibly merge. I've already added an environment variable that permits adding additional -D/-U flags to the compilation of the bundled sqlcipher and/or sqlite3. Perhaps another environment variable interpreted as pointing to a folder of patches, to be applied in alphabetic order? Then a package that depends on `libsqlite3-sys` but wants to patch the sqlcipher source could include the patches in its distribution. (As I'm imagining this, though, it would only work once, patch folders couldn't be chained.)

Re 3. I know my changes would still permit dynamically linking to `libcrypto`. I think they would also permit statically linking to it. Does NSS provide `libcrypto.so/dylib` and `libcrypto.a`?

dubiousjim:
I'm waiting for the author of the `autotools` package to make a release with the latest master of his repo.

[...]

PR #771, opened 2020-06-25
Add features to bundle sqlcipher (and optionally also OpenSSL libcrypto)

dubiousjim:
Discussed in #765


thomcc:
This should work to avoid the clippy allow, no?

```rust
            let (lib_dir, inc_dir) = match (lib_dir, inc_dir) {
                (Some(lib_dir), Some(inc_dir)) => {
                    use_openssl = true;
                    (lib_dir, inc_dir)
                }
                (lib_dir, inc_dir) => {
                    match find_openssl_dir(&host, &target) {
                        None => {
                            if is_windows {
                                panic!("Missing environment variable OPENSSL_DIR or OPENSSL_DIR is not set")
                            } else if is_apple && Path::new("/opt/local/lib/libssl.a").exists() {
                                // TODO: we default to using MacPorts libraries if installed, perhaps
                                // should provide option to use SecurityFoundation instead?
                                use_openssl = true;
                                (
                                    PathBuf::from("/opt/local/lib"),
                                    PathBuf::from("/opt/local/include"),
                                )
                            } else {
                                (PathBuf::new(), PathBuf::new())
                            }
                        }
                        Some(openssl_dir) => {
                            let lib_dir = lib_dir.unwrap_or_else(|| openssl_dir.join("lib"));
                            let inc_dir = inc_dir.unwrap_or_else(|| openssl_dir.join("include"));

                            if !Path::new(&lib_dir).exists() {
                                panic!(
                                    "OpenSSL library directory does not exist: {}",
                                    lib_dir.to_string_lossy()
                                );
                            }

                            if !Path::new(&inc_dir).exists() {
                                panic!(
                                    "OpenSSL include directory does not exist: {}",
                                    inc_dir.to_string_lossy()
                                )
                            }

                            use_openssl = true;
                            (lib_dir, inc_dir)
                        }
                    }
                }
            };
```

dubiousjim:
Yeah, that might work. I had tried an `if let (Some(...), Some(...)) = (..., ...)`, but then got an error about moving in the else branch.

libsqlite3-sys/build.rs
```diff
@@ -81,6 +97,93 @@ mod build_bundled {
...
+                (Some(lib_dir), Some(inc_dir)) => {
+                    use_openssl = true;
+                    (lib_dir, inc_dir)
+                }
+                (lib_dir, inc_dir) => {
+                    match find_openssl_dir(&host, &target) {
+                        None => {
+                            if is_windows {
+                                panic!("Missing environment variable OPENSSL_DIR or OPENSSL_DIR is not set")
+                            } else if is_apple && Path::new("/opt/local/lib/libssl.a").exists() {
```
thomcc, line 123: Macports is not as widely used as e.g. brew.

But... I also am super skeptical about hardcoding this path, as various package managers have been known to leave stale copies of these around (there was a recent homebrew bug about this!)

In general, I'd *really strongly* prefer if we could leave searching for openssl and such to the e.g. openssl-sys crate and the like?

dubiousjim: I'm happy to drop the convenience lookups for MacPorts, and Homebrew too. The latter came from [rusqlcipher](https://lib.rs/crates/rusqlcipher), which was the basis for my PR, but stalled around 2018. We should give its author(s) credit though for the starting point of this PR.

Once we stabilize on the implementation of this, we can expand the docs to give suggestions about what env variables need to be set for linking against which copies of sqlite3 and/or sqlcipher and/or various crypto provider libraries for sqlcipher.

Taking out the hard-coded convenience paths for openssl is easy. Shifting the search for openssl libraries, when non-bundled, isn't, because there's no package to shift the burden over to. (`openssl-sys` is only a dependency when we're asking for the bundled ssl.) Do you want to drop the `pkg-config` calls too? Would you prefer to shift all of the bundling logic to a new package, which libsqlite3-sys has as an optional dependency, and just retrieve the linking information from `DEP_XXX` flags that the new package sets? I expect that'd be do-able, but more cumbersome and complicated than what we're looking at now.

thomcc: > We should give its author(s) credit though for the starting point of this PR.

A "co-authored-by" line in the final commit maybe would not be out of place, then. Unless you have a better idea.

Probably after you're all done squash this down a bit to clean up the history and add that.

dubiousjim: Sounds good. I'll wait until it's clear we've got the PR where you want it.

ubamrein:
Hi there :)

I'm currently trying to build for Android and iOS and I was wondering, why Openssl is not linked statically? I understood the `bundled` as everything being included in the resulting binary, which does not work when `libcrypto` is dynamically linked (at least in my case on iOS. As soon as I changed Line 152 of build.rs to
```rust
  println!("cargo:rustc-link-lib=static=crypto"); // cargo will resolve downstream to the static lib in openssl-sys
```
it was working.)

Further, I was wondering what the state of this PR currently is. Since I need it for my Android/iOS builds, I would be happy to finish this PR :)

emmanueltouzery:
if that helps in the end in my case i did it "by hand" making a docker image based on rust-musl-builder:
https://github.com/emmanueltouzery/projectpad2/blob/master/Dockerfile

dubiousjim:
The `bundled` feature just bundles the sqlite lib, and `bundled-sqlcipher` only bundles the sqlcipher one. If you use `bundled-ssl` it's supposed to bundle both the sqlcipher lib and the needed openssl libs. You shouldn't have to be changing line 152 of build.rs.

This PR was working for me in the spring, it just needs @thomcc's review. But I've been sidetracked since I submitted it, and haven't been able to track evolutions in this repo since then. It may (or may not) need some tweaking to accommodate that.

ubamrein:
Well I mean it instructs cargo to link dynamically. I think in my case the problem is that I'm building a static library myself, and the final linking takes place in xcode. But xcode does not know anything about intermediate build results from cargo... 

So I'm wondering, if there is any downsight of just linking it statically?

dubiousjim:
I see. I can't remember if there was a positive reason to make that line be:

```
println!("cargo:rustc-link-lib=dylib=crypto"); // cargo will resolve downstream to the static lib in openssl-sys
```

rather than:

```
println!("cargo:rustc-link-lib=static=crypto");
```

It's possible I had it be `dylib` just to prove that (when cargo did the linking) the library would still get statically linked. I don't remember now, sorry. Will have to get my head back inside this. And won't be able to do that just now.

dubiousjim:
Closing in favor of #860

PR #773, opened 2020-06-26
Implement our own sqlite3_exec

gwenn:
Should fix issue related to unlock notify: #767
Caveat: many CString allocated.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -1,4 +1,4 @@
-use crate::{Connection, Result, NO_PARAMS};
+use crate::{Connection, Result};
 use std::ops::Deref;
 
 /// Options for transaction behavior. See [BEGIN
@@ -120,7 +120,7 @@ impl Transaction<'_> {
             TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
             TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
         };
-        conn.execute(query, NO_PARAMS).map(move |_| Transaction {
+        conn.execute_batch(query).map(move |_| Transaction {
             conn,
             drop_behavior: DropBehavior::Rollback,
         })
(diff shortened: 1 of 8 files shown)


src/transaction.rs
```diff
@@ -120,7 +120,7 @@ impl Transaction<'_> {
             TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
             TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
         };
-        conn.execute(query, NO_PARAMS).map(move |_| Transaction {
+        conn.execute_batch(query).map(move |_| Transaction {
```
thomcc, line 123: It's probably better to keep these as `execute` no? Given that it's slightly more efficient?

gwenn: > It's probably better to keep these next few calls as `execute` no? Given that it's slightly more efficient?

My wild guess is: `execute` < `execute_batch` < `sqlite3_exec` for a single statement.

thomcc: Yeah, it's hard for me to imagine it would be anything other than that, so maybe we should keep these parts as using execute()?

gwenn: > Yeah, it's hard for me to imagine it would be anything other than that, so maybe we should keep these parts as using execute()?

I mean `execute_batch` is faster than `execute` but slower than `sqlite3_exec`.
```
test bench_execute       ... bench:       2,923 ns/iter (+/- 291)
test bench_execute_batch ... bench:       2,847 ns/iter (+/- 374)
```

thomcc: Huh, very strange, fair enough though, nevermind!

(I'm probably going to dig into this later since it means `execute` with `NO_PARAMS` is doing work it doesn't have to, but that's for another time)

gwenn: > Huh, very strange, fair enough though, nevermind!
> 
> (I'm probably going to dig into this later since it means `execute` with `NO_PARAMS` is doing work it doesn't have to, but that's for another time)

* check_no_tail
* bind_parameter_count
* check_update
* reset

PR #776, opened 2020-06-29
Publically expose Map

thomcc:
Should fix CI failures.


thomcc:
Another option would be just to change the doc comment not to reference it.

gwenn:
Thanks.

Issue #778, opened 2020-06-30
Improve blob docs?


i-schuetz:
The documentation re: inserting/querying a blob [the doc](https://docs.rs/rusqlite/0.23.1/rusqlite/blob/index.html) seems confusing. The example does too many things at once (or is it only about updates?) and it's difficult to see what has to be done for a simple insert or query.

gwenn:
Could you please tell us which line(s) in https://github.com/rusqlite/rusqlite/blob/master/src/blob.rs#L25-L49 are confusing ?
And rusqlite is just a SQLite wrapper. You may find more explanations [here](https://sqlite.org/c3ref/blob_open.html).

thomcc:
I have a WIP branch that improves the blob API and allows more similar use to in SQLite. As it is I think the Write/Read/Seek impls are nice (and should stick around), but have a number of downsides that aren't required with the original pread/pwrite style API. For example, it requires &mut access to a blob perform reads, when underlying SQLite call could implement a read_at call that does not update or mutate internal state, and thus does not. It also requires first seeking to a spot before IO is performed there, which is tedious.

Sadly I'm not on my computer that has this right now. It's also not finished, but I'll see if I can do something about that this weekend.

This might make it more in line with the SQLite docs, as our API for blob IO is kind of different at the moment.

> is it only about updates

It's about queries or updates, but can only really be about inserts if you first insert a zeroblob, and then update using that. This is because blob IO can't change the size of blobs, only read or write them.

PR #780, opened 2020-07-04
Add positional blob i/o and adjust blob i/o example docs

thomcc:
Probably needs more tests, but I mentioned this in #778. I adjusted the example some to make it more clear why we're doing each step, but mostly this is about adding an API that more closely maps the SQLite one.


thomcc:
Hmm, I shouldn't assume i32 == c_int...

PR #782, opened 2020-07-11
Upgrade to time v0.2 and put it behind a feature flag

### LINKED ISSUE #653: Update time to 0.2 (and put it under a feature flag)

benesch:
This also removes the usage of time in the crate's top-level
documentation example, as was done for the README in #625.

Fix #653.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,9 @@ bundled-full = [
     "load_extension",
     "serde_json",
     "series",
+    # time v0.2 does not work with tarpaulin v0.14.0. See time-rs/time#265.
+    # Re-enable when time v0.3 is released with the fix.
+    # "time",
     "trace",
     "unlock_notify",
     "url",
@@ -91,7 +94,7 @@ bundled-full = [
 ]
 
 [dependencies]
-time = "0.1.0"
+time = { version = "0.2", optional = true }
 bitflags = "1.0"
 lru-cache = "0.1"
 chrono = { version = "0.4", optional = true }
@@ -140,7 +143,7 @@ name = "exec"
 harness = false
 
 [package.metadata.docs.rs]
-features = [ "backup", "blob", "chrono", "collation", "functions", "limits", "load_extension", "serde_json", "trace", "url", "vtab", "window", "modern_sqlite", "column_decltype" ]
+features = [ "backup", "blob", "chrono", "collation", "functions", "limits", "load_extension", "serde_json", "time", "trace", "url", "vtab", "window", "modern_sqlite", "column_decltype" ]
 all-features = false
 no-default-features = true
 default-target = "x86_64-unknown-linux-gnu"
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -10,40 +10,39 @@
 //! * Strings (`String` and `&str`)
 //! * Blobs (`Vec<u8>` and `&[u8]`)
 //!
-//! Additionally, because it is such a common data type, implementations are
-//! provided for `time::Timespec` that use the RFC 3339 date/time format,
+//! Additionally, if the `time` feature is enabled, implementations are
+//! provided for `time::OffsetDateTime` that use the RFC 3339 date/time format,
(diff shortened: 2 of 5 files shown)


src/types/time.rs
```diff
@@ -1,40 +1,40 @@
+//! `ToSql` and `FromSql` implementation for [`time::OffsetDateTime`].
 use crate::types::{FromSql, FromSqlError, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
 use crate::Result;
+use time::{OffsetDateTime, PrimitiveDateTime, UtcOffset};
 
 const CURRENT_TIMESTAMP_FMT: &str = "%Y-%m-%d %H:%M:%S";
-const SQLITE_DATETIME_FMT: &str = "%Y-%m-%dT%H:%M:%S.%fZ";
-const SQLITE_DATETIME_FMT_LEGACY: &str = "%Y-%m-%d %H:%M:%S:%f %Z";
+const SQLITE_DATETIME_FMT: &str = "%Y-%m-%dT%H:%M:%S.%NZ";
+const SQLITE_DATETIME_FMT_LEGACY: &str = "%Y-%m-%d %H:%M:%S:%N %z";
```
gwenn, line 8: Maybe we should drop support to `SQLITE_DATETIME_FMT_LEGACY` ?

src/types/mod.rs
```diff
@@ -10,36 +10,32 @@
...
 //! can be parsed by SQLite's builtin
 //! [datetime](https://www.sqlite.org/lang_datefunc.html) functions.  If you
-//! want different storage for timespecs, you can use a newtype. For example, to
-//! store timespecs as `f64`s:
+//! want different storage for datetimes, you can use a newtype. For example, to
+//! store datetimes as `i64`s counting the number of seconds since the Unix
+//! epoch:
 //!
-//! ```rust
+//! ```ignore
```
gwenn, line 22: Not tested, but it seems possible to activate feature:
https://users.rust-lang.org/t/doctests-that-require-a-non-default-feature-is-it-possible/29529/4

Cargo.toml
```diff
@@ -91,7 +91,7 @@ bundled-full = [
 ]
 
 [dependencies]
-time = "0.1.0"
+time = { version = "0.2", optional = true }
```
gwenn, line 94: Please, add this feature to [bundled-full](https://github.com/rusqlite/rusqlite/blob/master/Cargo.toml#L68) and [docs.rs](https://github.com/rusqlite/rusqlite/blob/master/Cargo.toml#L143).

benesch: Done.

src/types/mod.rs
benesch: Sure thing, done.

src/types/time.rs
benesch: That makes sense to me, but not sure what y'all require in the way of backwards compatibility. Just let me know what makes sense!

thomcc:
This seems like it causes coverage generation to fail.

Apparently this is fixed in the 0.3 branch (https://github.com/time-rs/time/issues/267), so maybe we'll wait on that...

That or you could do the legwork to exclude `time` from the feature-set we use to generate coverage, but that seems like it might be more trouble than it's worth...

thomcc:
@gwenn I think it would be better to have some fix for the coverage issue, even if that's just "remove it from bundled-full until time 0.3 is out". I'd rather not take this as-is...

benesch:
Ugh, rock and a hard place. We're only _just_ getting the Rust ecosystem onto time v0.2—I still have a few other transitive dependencies that need to upgrade. It's too bad they've already ceased development on time v0.2. 

If it's alright with you folks, I'll drop it from bundled-full and leave a comment about adding it back on v0.3?

thomcc:
> If it's alright with you folks, I'll drop it from bundled-full and leave a comment about adding it back on v0.3?

This is fine by me.

benesch:
Ok, pushed. 🤞

benesch:
And it's green!

PR #786, opened 2020-07-21
Serialize and deserialize interfaces

TjeuKayim:
This adds support for https://www.sqlite.org/c3ref/deserialize.html and https://www.sqlite.org/c3ref/serialize.html.
It required a lot of unsafe code, unfortunately. Suggestions to improve my coding skills are welcome, because I'm no expert in Rust.

My motivation is to use this for a WebAssembly application. Because the browser has no real file system, I want to pass a SQLite binary file as an in-memory array. However, the function `sqlite3_deserialize` was not yet exposed in `rusqlite`.

Known shortcomings:

- I couldn't find a good way to express that the passed `data: &[u8]` in `unsafe fn deserialize_with_flags` should outlive `Connection`.
- When https://github.com/rust-lang/rust/issues/42774 is implemented, I'd rather use a `Vec<u8, SqliteAlloc>` instead of coding it myself. Maybe a crate exists offering such functionality now?


TjeuKayim:
The CI/CD already caught some of my mistakes.

src/deserialize.rs
```diff
@@ -0,0 +1,405 @@
+//! `feature = "deserialize"` Serialize and deserialize interfaces.
+//!
+//! This API is only available when SQLite is compiled with `SQLITE_ENABLE_DESERIALIZE`.
+//! These functions create and read a serialized file in-memory, useful on platforms without a real file system
+//! like WebAssembly or Cloud Functions.
+//!
+//! ```
+//! # use rusqlite::{Result, Connection, DatabaseName, NO_PARAMS};
+//! let db = Connection::open_in_memory().unwrap();
```
thomcc, line 9: Please reformat these example lines to be a little cleaner. You should also use `?`, like the following:
```rust
//! # fn main() -> Result<()> {
//! // example here
//! # Ok(())
//! # }
```

src/deserialize.rs
```diff
@@ -0,0 +1,405 @@
...
+/// # use rusqlite::deserialize::SerializedDb;
+/// let mut serialized = SerializedDb::new();
+/// serialized.extend(vec![1, 2, 3]);
+/// assert_eq!(serialized[1], 2);
+/// ```
+pub struct SerializedDb {
+    data: NonNull<u8>,
+    len: usize,
+    cap: usize,
+    phantom: PhantomData<[u8]>,
```
thomcc, line 130: What's the point of `PhantomData<[u8]>`?

src/deserialize.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// if the database where backed up to disk.
+    ///
+    /// Returns `Ok(None)` when [`DatabaseName`] does not exist or malloc/prepare fails.
+    ///
+    /// # Safety
+    ///
+    /// If [`SerializeFlags::NO_COPY`] is set, this returns a pointer to memory that SQLite is currently using
+    /// (or `Ok(None)` if this is not available).
+    /// In that case, the returned `SerializedDb` mutably borrows `Connection`,
+    /// [`std::mem::forget()`] one of them to prevent double free.
```
thomcc, line 91: I don't think this is a good thing for us to have in our API. There has to be a better way...

src/deserialize.rs
TjeuKayim: You're right, `u8` does not implement `Drop` so the drop checker should not worry about it.

src/deserialize.rs
TjeuKayim: OK, I made this function private and added `pub fn serialize_no_copy` as a safe alternative.

thomcc:
FYI: I'm going to be unreachable for most/all of next week -- this looks largely good (and thanks for the work), and I'll prioritize getting a review for this done as soon as I'm back. Sorry for that.

TjeuKayim:
OK thomcc. I'm glad you're positive. Thanks for merging my other pull request btw.
The functions I added recently are safe wrappers for different combinations of `SQLITE_DESERIALIZE_*` flags. I'm content with the functionality now. My unit tests should maybe be split in smaller functions, and I could add more documentation examples.
For now, I'll rebase onto master, fix merge conflicts and force push.

thomcc:
> When rust-lang/rust#42774 is implemented, I'd rather use a Vec<u8, SqliteAlloc> instead of coding it myself. Maybe a crate exists offering such functionality now?

Hmm, which type are you referring to here, the one in mem_file.rs? I'll look into this — I don't love rolling our own here, but it might beat a complex dep at least in the fact that we can limit the scope of how much of Vec it reimplements.

P.S. I'm back now but have a bit of a backlog of stuff to deal with before I will likely get to this.

thomcc:
Oh, also, please fix the asan error you get in CI. Right now we do have some issues with asan right now, but the error in this patch is completely different from the "file not found" error in #789 (which I'll get to soonish hopefully, but also is intermittent as you can see from the fact that most of our daily CI runs [have been passing recently](https://github.com/rusqlite/rusqlite/actions)).

Anyway, your error looks like this (from https://github.com/rusqlite/rusqlite/pull/786/checks?check_run_id=942184275 ):

```
=================================================================
==5602==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24584 byte(s) in 1 object(s) allocated from:

    #0 0x556824913949  (/home/runner/work/rusqlite/rusqlite/target/x86_64-unknown-linux-gnu/debug/deps/rusqlite-c0bf155bf27f9dfd+0x340949)
    ...

SUMMARY: AddressSanitizer: 24584 byte(s) leaked in 1 allocation(s).
```

Which seems like you might have some logic that causes a leak....

Err, Ah, I see:

> The CI/CD already caught some of my mistakes. How should I interpret the Address Sanitizer logs?

Yeah, uh, hm. You might have to run the asan tests manually. I guess when I typed the first part of this I was hoping you might have some idea of where it was happening, but if that's not the case you probably need to run the asan tests locally.

This hopefully is just a matter of installing nightly, and running the following (Note that if you're on windows you'll *probably* need to do this in WSL, but if you're on mac or Linux it should be fine):

```
env RUSTFLAGS="-Zsanitizer=address" RUSTDOCFLAGS="-Zsanitizer=address" RUST_BACKTRACE=0 cargo +nightly -Z build-std test --features 'bundled-full session buildtime_bindgen with-asan' --target x86_64-unknown-linux-gnu
```

e.g. You're basically mimicing this: https://github.com/rusqlite/rusqlite/blob/master/.github/workflows/main.yml#L79-L101. You can probably relax those features if they aren't relevant to your asan failure (which would be surprising) -- the important ones for running with asan are `bundled` and `with-asan`.

If this doesn't work, feel free to reach out [in the gitter](https://gitter.im/rusqlite/community). Feel free to ping me.

Once you get it set up, I wouldn't expect it to suddenly have this backtrace info (maybe it will — it would be very nice if so!). So, you probably will have to manually run relevant tests until you find the test responsible, unfortunately.

Again, feel free to reach out via gitter or whatever. (Hope this helps, fixing it is required to land this, I think)

TjeuKayim:
Thanks for the comments. I fixed the memory leak.

> Hmm, which type are you referring to here, the one in mem_file.rs? I'll look into this — I don't love rolling our own here, but it might beat a complex dep at least in the fact that we can limit the scope of how much of Vec it reimplements.

Yes, I was referring to `MemFile`. I tried to use https://docs.rs/alloc-wg/0.9.0/alloc_wg/, but got stuck (https://github.com/TjeuKayim/rusqlite/tree/deserialize-nightly). I gave up on the idea of using a dependency.

TjeuKayim:
I've ported all the `sqlite3_io_methods` from `memdb.c` at https://github.com/TjeuKayim/rusqlite/blob/deserialize-io-methods/src/deserialize/mod.rs. What's left to do is writing unit tests for every function, porting `Connection::serialize` so that it does not copy twice (first to `sqlite3_malloc` and then to `Vec`), and removing `mem_file.rs`.
Is this new approach better in your opinion?

TjeuKayim:
The last commit fixes an unsoundness that I discovered when using this fork in practice.

thomcc (changes requested):
First pass, didn't make it all the way through (only really go through part of deserialize). I have to run now, and I have lots to still think about.

An initial thought is that the threading situation here concerns me greatly:

First off, I'm pretty confident that you should be using Arc<T> instead of Rc<T>.

But... Even if so... Can SQLite access VFS's from multiple threads? I believe the answer is yes. It certainly would if you have multiple connections open, No?

And I think it can even do this simultaneously when only one connection thread is active... However, this may require SQLITE_LIMIT_WORKER_THREADS to be larger (and perhaps even then I'm wrong — do the workers do IO? Or do they just do in-memory sorting... @gwenn might know since he's more familiar with SQLite's intricacies than I am)

P.S. Super sorry about the massive delay here.

src/deserialize.rs
```diff
@@ -0,0 +1,1299 @@
...
+    f: impl FnOnce(&mut VecDbFile) -> c_int + panic::UnwindSafe,
+) -> c_int {
+    panic::catch_unwind(|| f(&mut *(file as *mut VecDbFile))).unwrap_or_else(|_| ffi::SQLITE_ERROR)
+}
+
+/// This will be called when dropping the `Connection` or
+/// when the database gets detached.
+unsafe extern "C" fn x_close(file: *mut ffi::sqlite3_file) -> c_int {
+    catch_unwind_sqlite_error(file, |file| {
+        ptr::drop_in_place(file);
```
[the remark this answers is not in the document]
thomcc: Actually, perhaps you need to deallocate this pointer? I need to look more around and see where it's allocated.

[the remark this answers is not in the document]
TjeuKayim: `catch_unwind_sqlite_error` casts the pointer to `&mut VecDbFile`, the value pointed to holds a `Rc<MemFile>`, that's why it needs a drop.

src/deserialize.rs
```diff
@@ -0,0 +1,1299 @@
...
+///
+/// The vector grows as needed, but is not automatically shrunk when the
+/// database file is reduced in size.
+#[non_exhaustive]
+#[derive(PartialEq)]
+pub enum MemFile<'a> {
+    /// Owned vector.
+    Owned(Vec<u8>),
+    /// Mutably borrowed vector that can be written and resized.
+    Writable(&'a mut Vec<u8>),
```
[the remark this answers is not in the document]
TjeuKayim: I'm also not sure about this. The only place where `VecDbFile` exposes this pointer to SQLite is the function `x_fetch()`. The documentation on Memory-Mapped I/O seems to indicate that SQLite never writes to the fetched memory, we can try to proof wrong that assumption.
See my comment at line 742: https://github.com/rusqlite/rusqlite/blob/3f3d315d82b382aafd08774d51e770102b9b7134/src/deserialize.rs#L742-L744.

src/deserialize.rs
```diff
@@ -0,0 +1,1299 @@
...
+}
+
+/// `sqlite3_file` subclass for the `vec_db` Virtual File System. It's inspired by
+/// the `memdb` VFS but uses the Rust allocator instead of `sqlite3_malloc`.
+/// The database is stored in an owned or borrowed `Vec<u8>`.
+#[repr(C)]
+#[derive(Debug)]
+struct VecDbFile<'a> {
+    methods: *const ffi::sqlite3_io_methods,
+    data: Rc<MemFile<'a>>,
```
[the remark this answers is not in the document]
TjeuKayim: Ok, I agree, let me change that.

TjeuKayim:
Related to this pull request, I posted a feature request in the [SQLite Forum to bring your own buffer to sqlite3_serialize](https://www.sqlite.org/forum/forumpost/ef64cd3039). That would make serializing to a Vec more efficient than using the backup API. Unfortunately, it didn't receive any comments.

TjeuKayim:
Nightly already has `struct Vec<T, #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global>` for a few months now. So, we could use that in `Connection::serialize()` to return a vector to sqlite3_malloc allocated memory instead of using the Backup API. That would simplify the serialize part. What is your opinion on that?
Another question, do you think that it is a good idea to split this pull request into two, one for serializing and one for deserializing. Both features could get their own file and feature flag.

GeeWee:
I'm not related to the implementation on either side here, but as there's no tracking issue I can find, I just wanted to chime in and say that I'm also using WASM and would love to use rusqlite, but cannot as I would also need this functionality.

PR #793, opened 2020-08-17
update to sqlite 3.33.0

dae:
The new "UPDATE FROM" syntax will be useful: https://www.sqlite.org/releaselog/3_33_0.html


gwenn:
Thanks.
It does not matter but you should not have updated the `libsqlite3-sys` version.

dae:
I noticed you didn't do that last time, but the last line of upgrade.sh says:

echo 'You should increment the version in libsqlite3-sys/Cargo.toml'

Might be worth adding more details there if there are times when it should and shouldn't be done :-)

PR #796, opened 2020-08-17
Add support for linking to winsqlite3

nazar554:
There is system sqlite library on Windows 10 machines since Anniversary Update https://blogs.windows.com/windowsdeveloper/2017/02/06/using-sqlite-databases-uwp-apps/

Not sure if using `wrapper.h` is a good idea, otherwise I need to figure out full path to Windows SDK

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ bundled-windows = ["libsqlite3-sys/bundled-windows"]
 with-asan = ["libsqlite3-sys/with-asan"]
 column_decltype = []
 wasm32-wasi-vfs = ["libsqlite3-sys/wasm32-wasi-vfs"]
+winsqlite3 = ["libsqlite3-sys/winsqlite3"]
 
 # Helper feature for enabling both `bundled` and most non-build-related optional
 # features or dependencies. This is useful for running tests / clippy / etc. New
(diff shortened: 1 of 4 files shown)


thomcc:
Is there anything that should be done for our CI coverage here?

nazar554:
@thomcc Maybe enabling builds with `winsqlite3` feature for Windows `-msvc` and `-gnu` targets? Not sure where to do that though

thomcc (approved):
Thanks

thomcc:
Okay. I'll figure out CI in a follow up.

@nazar554 Do you have any links where I can find out version info for versions shipped with win10? It's relevant probably to #706.

I'm very surprised to hear they're shipping such an old version in windows, since SQLite tends not to break stuff with updates...

nazar554:
@thomcc [Here](https://abi-laboratory.pro/compatibility/Windows_8.1_to_Windows_10_1511_10586.494/x86_64/abi_compat_report.html) it was added (`3.8.8.3`, Windows 10 `10.0.10586.0`)

And then [updated](https://abi-laboratory.pro/compatibility/Windows_8.1_to_Windows_10_1511_10586.494/x86_64/abi_compat_report.html) to `3.12.2` in Windows 10 `10.0.14393.0`

I am running `10.0.19041.508` (May 2020 update) right now and `winsqlite3.dll` is version `3.29.0.0`

thomcc:
3.29 wouldn't be a problem to support. 3.8 and 3.12 are both newer than our current min version, although honestly I kind of want us to be able to say something like "we don't support SQLite versions older than 5 years" (in rusqlite — the pain caused by supporting these in libsqlite3-sys is pretty minor).

Oh well, more food for thought I guess.

PR #806, opened 2020-10-04
Generate doc with 'array' feature enabled

### LINKED ISSUE #781: rarray is not documented?
I am looking for a way to make a query
```
SELECT * FROM table WHERE index IN([list of values]);
```
passing list of values directly from rust Vec.

While it looks like there is a support for that here:
https://github.com/rusqlite/rusqlite/blob/master/src/vtab/array.rs#L51

There are no:
- documentation for array module in docs: https://docs.rs/rusqlite/0.23.1/rusqlite/trait.ToSql.html?search=array - empty searach
- no information about implementing ToSql for Array type: https://docs.rs/rusqlite/0.23.1/rusqlite/trait.ToSql.html
- no examples how to make even such simple query

gwenn:
Should fix #781

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,7 +144,7 @@ name = "exec"
 harness = false
 
 [package.metadata.docs.rs]
-features = [ "backup", "blob", "chrono", "collation", "functions", "limits", "load_extension", "serde_json", "time", "trace", "url", "vtab", "window", "modern_sqlite", "column_decltype" ]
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Should any of the other custom virtual tables be here too?

(Also sorry for being missing for so long — I should have more time this coming week to catch up on some reviews and such...)

gwenn:
No one asks for other extensions, so...

thomcc:
Fair enough!

PR #808, opened 2020-10-04
Update dependency versions

thomcc:


gwenn:
Could you please explain why we should update these versions ?
Because for me, `cargo update` does it automatically, no ?

PR #811, opened 2020-10-05
Replace lru-cache with hashlink

### LINKED ISSUE #810: Consider moving away from the lru-cache dependency
[lru-cache](https://github.com/contain-rs/lru-cache) is a project "IN MAINTENANCE MODE, DUE TO INSUFFICIENT MAINTAINER RESOURCES" which depends on [linked-hash-map](https://github.com/contain-rs/linked-hash-map), another project "IN MAINTENANCE MODE …" and with [undefined behavior issues](https://github.com/contain-rs/linked-hash-map/issues/97) that are now **[caught by the stdlib assertions](https://github.com/rust-lang/rust/issues/77585)** in nightly Rust, which makes Firefox built with nightly Rust crash.

It's probably possible to fix lru-cache by switching to [hashlink](https://github.com/kyren/hashlink), but maybe switching to a more actively maintained LRU cache library in rusqlite would be a good idea.

thomcc:
Fixes #810, although hashlink is kinda a heavyweight dep (pulls in hashbrown). Oh well, probably doesn't matter that much, I didn't notice it taking much longer to compile anyway.

We should be able to simplify some of the nonsense with the source `Arc<str>` we do too, since hashlink supports raw_entry. I'll do that some time in the future, maybe.

If https://github.com/kyren/hashlink/pull/4 will land soon then we should probably wait for that, since as it is `hashlink::LruCache` doesn't impl Debug, and our StatementCache does. IDR if StatementCache is public, but either way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,7 +98,7 @@ bundled-full = [
 [dependencies]
 time = { version = "0.2", optional = true }
 bitflags = "1.2"
-lru-cache = "0.1"
+hashlink = "0.5"
 chrono = { version = "0.4", optional = true }
 serde_json = { version = "1.0", optional = true }
 csv = { version = "1.1", optional = true }
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -2,7 +2,7 @@
 
 use crate::raw_statement::RawStatement;
 use crate::{Connection, Result, Statement};
-use lru_cache::LruCache;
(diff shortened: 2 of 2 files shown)


gwenn (approved):
LGTM

thomcc:
Ah, this isn't a pub type so we can just land it.

PR #812, opened 2020-10-06
Prepare release 0.24.1

thomcc:
This is just for #811, and even then because it causes crashes for firefox when built with nightly rust.

@eoger / @mhammond / @rfk after this release is cut, can someone update the vendored rusqlite in m-c? I don't want to set up a m-c dev env. (Sorry for pinging all three of you, but IDK who still is on the team responsible for this)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusqlite"
-version = "0.24.0"
+version = "0.24.1"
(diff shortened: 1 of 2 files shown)


gwenn (approved):
Ideally, README should be updated but that's fine if you don't:
```
:%s/\<0.24.0\>/0.24.1/gc
```

PR #815, opened 2020-10-07
Try to find where github actions windows CI keeps clang so bindgen can use it.

thomcc:
might not work


thomcc:
i mean i found it, but can't make it work. I'll file a bug with clang-sys about this later.

PR #819, opened 2020-10-16
Add support for compiling sqlite3-sys using OS_OTHER

nhynes:
This PR adds the option to compile sqlite-sys using [OS_OTHER](https://sqlite.org/compile.html#os_other), which  is useful when running on exotic platforms.


libsqlite3-sys/build.rs
```diff
@@ -66,23 +66,39 @@ mod build_bundled {
...
             .flag("-DSQLITE_ENABLE_STAT4")
             .flag("-DSQLITE_SOUNDEX")
             .flag("-DSQLITE_THREADSAFE=1")
-            .flag("-DSQLITE_USE_URI")
             .flag("-DHAVE_USLEEP=1")
             .flag("-D_POSIX_THREAD_SAFE_FUNCTIONS") // cross compile with MinGW
             .warnings(false);
 
+        if cfg!(feature = "os-other") {
+            cfg.flag("-DSQLITE_OS_OTHER")
```
nhynes, line 79: other than this line and the `-D_FORTIFY_SOURCE`, eliding the optional features is intended to keep the build small and secure on embedded-like platforms. I can revert those changes, if desired.

libsqlite3-sys/sqlite3/sqlite3.c
```diff
@@ -230515,3 +230515,19 @@ SQLITE_API int sqlite3_stmt_init(
...
+  assert(n <= destlen);
+	return memcpy(dest, src, n);
+}
+
+void *__memset_chk(void *dest, int c, size_t n, size_t destlen)
+{
+  assert(n <= destlen);
+	return memset(dest, c, n);
+}
+#endif
```
nhynes, line 230533: maybe these should be echoed in by the build script to make it easer to update the sqlite amalgamation?

thomcc: We shouldn't modify the sqlite3.c and should just add another .c file for this, I think.

We could even expose them as symbols from rust honestly...

Cargo.toml
```diff
@@ -62,6 +62,7 @@ with-asan = ["libsqlite3-sys/with-asan"]
 column_decltype = []
 wasm32-wasi-vfs = ["libsqlite3-sys/wasm32-wasi-vfs"]
 winsqlite3 = ["libsqlite3-sys/winsqlite3"]
+os-other = ["libsqlite3-sys/os-other"]
```
thomcc, line 65: I think we want a better name for this. We possibly should even detect it from the build.rs based on the value of `std::env::var("TARGET")`...

libsqlite3-sys/build.rs
thomcc: Hmm, I guess this would be a reason not to have it auto-detected based on the value of `TARGET`...

thomcc (commented):
What's the use case here? I'm in favor but which sorts of OSes are you targeting?

Have a few nits. Sorry for delay in reviewing.

libsqlite3-sys/build.rs
nhynes: I think that might work, actually. I'm using this flag to compile to `x86_64-fortanix-unknown-sgx` where mutexes and filesystems don't quite work as they do on unix or windows. Maybe `if cfg!(target_env = "sgx")` is correct–just not as general.

PR #823, opened 2020-10-22
Implement From/ToSql for more types

### LINKED ISSUE #821: Implement ToSql/FromSql for u64
As discussed in #802 it is pretty awkward that `ToSql` and `FromSql` are not implemented for `u64`. The current justification is that SQLite cannot actually store integers between `i64::MAX` and `u64::MAX`, but it would be simpler just to return an error in that case (`FromSqlError::OutOfRange`, or is it `Error::IntegralValueOutOfRange`, or possibly `Error::FromSqlConversionFailure`?).

### LINKED ISSUE #822: Implement ToSql/FromSql for f32
This just seems to be an oversight.

Timmmm:
This implements `FromSql` for `u64`, `usize` and `f32`, and `ToSql` for `f32`. I omitted the possibly controversial change of implementing `ToSql` for `u64` and `usize`, mostly because the `ToSql` trait does not allow failure at the moment and it would have been a ton of work to fix it. This solves most of the annoyances anyway.

I also updated the documentation to describe how it currently works, and changed the implementation to use `try_from` for integral casts rather rather than custom code.

Fixes #822, and solves half of #821.

### DIFF
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -1,7 +1,7 @@
 use super::{Value, ValueRef};
+use std::convert::TryInto;
 use std::error::Error;
 use std::fmt;
-use std::convert::TryInto;
 
 /// Enum listing possible errors from `FromSql` trait.
 #[derive(Debug)]
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -1,6 +1,7 @@
 use super::{Value, ValueRef};
 use std::error::Error;
 use std::fmt;
+use std::convert::TryInto;
 
 /// Enum listing possible errors from `FromSql` trait.
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


thomcc:
Feel free to allow the clippy float comparison lint. Needs rustfmt tho

thomcc (approved):
Thanks for the additional test coverage too.

gwenn:
> `ToSql` trait does not allow failure at the moment

Are you sure ?
https://docs.rs/rusqlite/0.24.1/rusqlite/types/trait.ToSql.html#tymethod.to_sql
```rust
fn to_sql(&self) -> Result<ToSqlOutput<'_>>
```

Timmmm:
> Are you sure ?

Ah good point - I got mixed up. It's actually because internally `ToSql` is implemented by implementing `From<T> for ToSqlOutput` for all the primitive types `T`, and `From` *is* infallible.

I guess the fix is to change it to use `TryFrom` instead. Any objection if I do that in a future commit?

thomcc:
Go for it. I'll merge this for now though.

PR #824, opened 2020-10-25
Expose query progress information

### LINKED ISSUE #803: Expose progress information
I am using rusqlite's `execute_batch` to run database scripts. As these scripts may take a long time to complete (and sometimes lead to errors) I would like to know the current query that is executing, and possibly progress for each query. 

It would be easy to reimplement `execute_batch` myself, except for the fact that the private statement struct (`stmt`) is not accessible, which means I am unable to parse multiple statements:

````
pub fn execute_batch(&self, sql: &str) -> Result<()> {
        let mut sql = sql;
        while !sql.is_empty() {
            let stmt = self.prepare(sql)?;
            if !stmt.stmt.is_null() && stmt.step()? && cfg!(feature = "extra_check") {
                // Some PRAGMA may return rows
                return Err(Error::ExecuteReturnedResults);
            }
            let tail = stmt.stmt.tail();
            if tail == 0 || tail >= sql.len() {
                break;
            }
            sql = &sql[tail..];
        }
        Ok(())
    }
````

A solution would be to provide access to (parts of) the `RawStatement` functionality. Another solution would be to add an interface to perform the parsing (and return a `Vec<String>` or `Vec<Statement>`), after which I can execute the statements one-at-a-time myself.

To obtain progress information for an individual statement, one could use [`sqlite3_progress_handler`](https://www.sqlite.org/c3ref/progress_handler.html) which is not currently exposed either.

### PR DESCRIPTION
**Pull request by gwenn**
Should fix #803

### DIFF
--- a/src/blob/mod.rs
+++ b/src/blob/mod.rs
@@ -136,7 +136,10 @@
 //!
 //! // Insert another BLOB, this time using a parameter passed in from
 //! // rust (potentially with a dynamic size).
-//! db.execute("INSERT INTO test_table (content) VALUES (?)", &[ZeroBlob(64)])?;
+//! db.execute(
+//!     "INSERT INTO test_table (content) VALUES (?)",
+//!     &[ZeroBlob(64)],
+//! )?;
 //!
 //! // given a new row ID, we can reopen the blob on that row
 //! let rowid = db.last_insert_rowid();
@@ -177,7 +180,10 @@
 //!
 //! // Insert another blob, this time using a parameter passed in from
 //! // rust (potentially with a dynamic size).
-//! db.execute("INSERT INTO test_table (content) VALUES (?)", &[ZeroBlob(64)])?;
+//! db.execute(
+//!     "INSERT INTO test_table (content) VALUES (?)",
+//!     &[ZeroBlob(64)],
+//! )?;
 //!
 //! // given a new row ID, we can reopen the blob on that row
 //! let rowid = db.last_insert_rowid();
@@ -196,8 +202,8 @@ use crate::{Connection, DatabaseName, Result};
 
 mod pos_io;
 
-/// `feature = "blob"` Handle to an open BLOB. See [`rusqlite::blob`](crate::blob) documentation for
-/// in-depth discussion.
+/// `feature = "blob"` Handle to an open BLOB. See
+/// [`rusqlite::blob`](crate::blob) documentation for in-depth discussion.
 pub struct Blob<'conn> {
     conn: &'conn Connection,
     blob: *mut ffi::sqlite3_blob,
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -77,8 +77,10 @@ impl Connection {
 
     /// `feature = "hooks"` Register a query progress callback.
     ///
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: src/blob/mod.rs [hunk 1]
```diff
@@ -136,7 +136,10 @@
 //!
 //! // Insert another BLOB, this time using a parameter passed in from
 //! // rust (potentially with a dynamic size).
-//! db.execute("INSERT INTO test_table (content) VALUES (?)", &[ZeroBlob(64)])?;
+//! db.execute(
```
thomcc, line 139: Sorry about these, I forgot you use some unstable rustfmt settings. I'll think about a way to avoid this in the future, even if it's just running it on nightly. I mostly don't want to impose that barrier for contributors, but dont mind it that much myself.

src/lib.rs
```diff
@@ -810,6 +810,61 @@ impl fmt::Debug for Connection {
...
+}
+
+impl<'conn, 'sql> Batch<'conn, 'sql> {
+    /// Constructor
+    pub fn new(conn: &'conn Connection, sql: &'sql str) -> Batch<'conn, 'sql> {
+        Batch { conn, sql, tail: 0 }
+    }
+}
+
+impl<'conn> fallible_iterator::FallibleIterator for Batch<'conn, '_> {
```
thomcc, line 845: Is there a reason not to `impl Iterator<Item=Result<Statement<'conn>, Error>>`? I'm less sold on the benefit of FallibleIterator now that `transpose()` is stable and can trivially change between `Result<Option<T>, E>` and `Option<Result<T, E>>`.

Implementing both wouldn't bother me if you prefer the FallibleIterator API, but IMO not having to bring in the fallible iterator crate and import the trait to use this type is nicer for most users.

src/hooks.rs
```diff
@@ -282,33 +335,61 @@ mod test {
...
         db.execute_batch("INSERT INTO foo VALUES ('lisa')").unwrap();
+        assert!(called);
+    }
+
+    #[test]
+    fn test_progress_handler() {
+        let db = Connection::open_in_memory().unwrap();
+
+        lazy_static! {
+            static ref CALLED: AtomicBool = AtomicBool::new(false);
```
thomcc, line 369: Lazy static isn't needed for atomics. You can just do

```rust
static CALLED: AtomicBool = AtomicBool::new(false);
```

Outside the lazy_static.

thomcc (commented):
Looks good, just some drive-by nits since I was around.

src/hooks.rs
gwenn: Thanks for the tip.

src/lib.rs
gwenn: I guess we can use the [iterator](https://docs.rs/fallible-iterator/0.2.0/fallible_iterator/trait.FallibleIterator.html#method.iterator) method.

gwenn: We should not implements both!

gwenn: Should be ok now: you can use either `while let Some(stmt) = batch {...}` or `for stmt in batch {...}`.

thomcc (approved):
This looks fine to me. Feel free to merge whenever.

Issue #393, opened 2018-09-07
impl Clone for rusqlite::Connection?


mqudsi:
So long as SQLite is not running in single-threaded mode (and the default is that it's not), it should be safe to offer a `.clone()` implementation that would re-open the same database file with same settings/configuration and return the `Sync` result, while playing nicely with `!Sync`.

thomcc:
This seems difficult if the database was in-memory, and probably requires tracking which pragmas were executed.

gwenn:
If the in-memory mode is specified using [URI filenames](https://sqlite.org/uri.html)/[also](https://sqlite.org/c3ref/open.html#urifilenamesinsqlite3open), it seems feasible.

thomcc:
Right, that's what I meant by 'a named in-memory database'. There's not really any guarantee that's true.

gwenn:
If [`Connection.path.is_some()`](https://docs.rs/rusqlite/0.14.0/src/rusqlite/lib.rs.html#199) and if we keep the open flags, we can clone. 
But we can't use the standard trait because [clone](https://doc.rust-lang.org/std/clone/trait.Clone.html#tymethod.clone) cannot fail.

Why not use [r2d2](https://docs.rs/r2d2_sqlite/0.6.0/r2d2_sqlite/index.html#example) ?
Or [this](https://github.com/gwenn/r2d2-sqlite/blob/master/src/lib.rs#L25) if you need open flags.

mqudsi:
Circling back, this is doable if you add a phantom generic type indicating whether or not the connection has a path, since that's explicitly a function of which API call was used to create the connection. (You can then actually move `path`  to a specialized function of the pathed connection, and get rid of the `Option` entirely. ) Then `Clone` can be implemented just for the specialized type with an explicit path.

However, I'm not sure if it should be implemented since this can result in multiple `DatabaseLocked` errors if multiple threads (or processes) simultaneously try merging their changes from a journal back to the database.

fossdd:
Hey, looking back to it: It's still a nice idea.

okpatil4u:
@mqudsi thank you very much for the commit !
How long do you think it will take until this feature is stable ?

thomcc:
So, it hasn't been submitted as a PR, but I took a peek. Here's a preliminary review, assuming @mqudsi intends for a PR. Caveat: I'm somewhat against this, but not to the point where it's wontfix or anything.

---

[...]

PR #825, opened 2020-10-25
Reduce required lifetime in create_scalar_function

nhynes:
The docs explicitly say the function only exists for the duration of the connection. Reducing the lifetime requirement makes it possible to pass in a sufficiently long-lived closure that captures additional context to the function

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -334,15 +334,15 @@ impl Connection {
     /// # Failure
     ///
     /// Will return Err if the function could not be attached to the connection.
-    pub fn create_scalar_function<F, T>(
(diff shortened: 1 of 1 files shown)


gwenn:
Thanks.
Introduced by #383

gwenn (approved):

thomcc (approved):
LGTM. This makes stuff like #393 harder but I was never convinced we should do that anyway, and not sure how it would work for user defined functions.

thomcc:
Going to merge despite the note about 393 since again, IDK if theres a way to make that work with user defined functions.

PR #826, opened 2020-10-26
Add ToSql implementations for u64 and usize

### LINKED ISSUE #821: Implement ToSql/FromSql for u64

Timmmm:
Fixes #821.

Note this line: `// TODO: Include the values in the error message.` I wasn't sure what the best thing to do here was - ideally it would say something like `Value 12342345234534 too big to store in an i64`, but I think [it will just say](https://doc.rust-lang.org/src/core/num/mod.rs.html#5145) `out of range integral type conversion attempted`.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -11,22 +11,22 @@
 //! The number situation is a little complicated due to the fact that all
 //! numbers in SQLite are stored as `INTEGER` (`i64`) or `REAL` (`f64`).
 //!
-//! `ToSql` cannot fail and is therefore implemented for all number types that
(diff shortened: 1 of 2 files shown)


gwenn (approved):

thomcc (approved):
Thanks

PR #830, opened 2020-11-03
Overhaul parameter API

### LINKED ISSUE #609: Separation of named/positional params lead to API duplication

thomcc:
Fixes #609

I've wanted to do this for a long time, but it required a bunch of type-fu that I wasn't fully prepared for how to do. (It required differentiating between slices, references to arrays, 

This fixes two warts in the current API:

1. The `_named` variants of many functions are now redundant — functions accepting query parameters now all support named params too. e.g. you can pass named parameters to `statement.query()`/`query_map`/etc and no longer need to use the `query_named`/`query_map_named`/etc variants. The `_named` variants are now all deprecated, but users should be able call the un-`_named` apis directly (the `_named` ones all just do that now). Alternatively, they can drop `#![allow(deprecated)]` somewhere if they don't care about using old APIs.

2. `rusqlite::NO_PARAMS` is no longer needed, although I left it in there for now. ~~It's probably worth deprecating too, but it's used everywhere so I didn't do this yet (I still removed it from the example code, and from some tests though).~~ I deprecated it — there's no need to keep it anymore, and people who don't have time to do the update can `#![allow(deprecated)]`. (The update is just replacing `NO_PARAMS` with `[]` though)

Both of these have *really* bugged me for quite some time (hence filing 609), but I didn't see a way around them. Anyway, the fix isn't so bad, and basically involves creating a new `Params` trait and implementing it for a lot of stuff.

I also took the opportunity to write more thorough docs on parameter passing since this expands the number of ways to do it, and will likely help people migrate. These docs are the doc comment for the new `Params` trait (which is a lot simpler than the design I came up with in #609).

# Breakage

This is a large change and is breaking, I suspect many (or even most) rusqlite users will need to make *some* changes, but they're intended to be fairly easy, and most code shouldn't be breaking.

I think the only two things (the most notable ones anyway) that used to work that require changing are:

1. `thing.query(&[1i32, 2, 3])` or similar where a slice literal is passed in and each item is not a reference. This should be changed to `thing.query([1i32, 2, 3])` (e.g. takes an array now) or one of a few other options (`&[&foo, &bar]` or `params![foo, bar]` both work too)

2. `thing.query(blah.iter())` or similar where the full iterator functionality of the old API is used. This is a pretty rare use case (only really coming up when dynamically generating SQL for the most part). It's still supported though, but you need to do `thing.query(rusqlite::params_from_iter(blah.iter()))` or something like that.

Aside from that there are several deprecations but they're all easy to fix — either use the new APIs or use `#![allow(deprecated)]`. (I don't really have strong opinions on when to take these out, but breaking it all in one go would be too much IMO)

There might be a few other cases but in general things should actually be more forgiving overall.

Anyway, when making a changelog, we should link here, although maybe I'll throw together something more explicit about updating.

(I also made `RowIndex` sealed while I was at it since I expected to mess with this too, but didn't have — still, it should be sealed, nobody should be implementing it externally.)

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -34,11 +34,14 @@ impl Statement<'_> {
     /// # use rusqlite::{Connection, Result, params};
     /// fn update_rows(conn: &Connection) -> Result<()> {
     ///     let mut stmt = conn.prepare("UPDATE foo SET bar = 'baz' WHERE qux = ?")?;
-    ///
+    ///     // The `rusqlite::params!` macro is mostly useful when the parameters do not
+    ///     // all have the same type, or if there are more than 32 parameters
+    ///     // at once.
     ///     stmt.execute(params![1i32])?;
-    ///     // Similarly...
+    ///     // However, it's not required, many cases are fine as:
     ///     stmt.execute(&[&2i32])?;
-    ///
+    ///     // Or even:
+    ///     stmt.execute([2i32])?;
     ///     Ok(())
     /// }
     /// ```
@@ -46,21 +49,18 @@ impl Statement<'_> {
     /// ### Use with named parameters
     ///
     /// ```rust,no_run
-    /// # use rusqlite::{Connection, Result};
-    /// fn insert(conn: &Connection) -> Result<usize> {
-    ///     let mut stmt = conn.prepare("INSERT INTO test (name) VALUES (:name)")?;
-    ///     stmt.execute(&[(":name", &"one")])
-    /// }
-    /// ```
-    ///
-    /// Note, the `named_params` macro is provided for syntactic convenience,
-    /// and so the above example could also be written as:
-    ///
-    /// ```rust,no_run
     /// # use rusqlite::{Connection, Result, named_params};
-    /// fn insert(conn: &Connection) -> Result<usize> {
-    ///     let mut stmt = conn.prepare("INSERT INTO test (name) VALUES (:name)")?;
-    ///     stmt.execute(named_params!{":name": "one"})
+    /// fn insert(conn: &Connection) -> Result<()> {
+    ///     let mut stmt = conn.prepare("INSERT INTO test (key, value) VALUES (:key, :value)")?;
+    ///     // The `rusqlite::named_params!` macro (like `params!`) is useful for heterogeneous
+    ///     // sets of parameters (where all parameters are not the same type), or for queries
+    ///     // with many (more than 32) statically known parameters.
+    ///     stmt.execute(named_params!{ ":key": "one", ":val": 2 })?;
+    ///     // However, named parameters can also be passed like:
+    ///     stmt.execute(&[(":key", "three"), (":val", "four")])?;
+    ///     // Or even: (note that a &T is required for the value type, currently)
+    ///     stmt.execute(&[(":key", &100), (":val", &200)])?;
+    ///     Ok(())
     /// }
     /// ```
     ///
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -909,7 +909,7 @@ pub enum StatementStatus {
 #[cfg(test)]
 mod test {
     use crate::types::ToSql;
-    use crate::{params_from_iter, Connection, Error, Result, NO_PARAMS};
+    use crate::{params_from_iter, Connection, Error, Result};
 
     #[test]
     #[allow(deprecated)]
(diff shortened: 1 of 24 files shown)


thomcc:
Note: My long term evil plan here is to allow various implementations of `Params` which support stuff like:

1. Binding positional parameters with an offset.
2. Explicitly opting out of param count checking.
3. Composing multiple implementations of `Params` (something conceptually like `impl<A, B> Params for (A, B) where A: Params, B: Params`, although perhaps there are reasons not to do it for tuples)
4. etc.

But that will be in future commits after more code is written working with this. This kind of thing is why `bind_in` is hidden/non-public

This will give a clean and consistent way of solving issues like https://github.com/rusqlite/rusqlite/pull/637 (currently solved by https://github.com/rusqlite/rusqlite/pull/668 but not in an ideal way), and is useful for other cases too.

src/statement.rs
```diff
@@ -1046,16 +1201,16 @@ mod test {
         let mut stmt = db
             .prepare("INSERT OR IGNORE INTO foo (x) VALUES (?)")
             .unwrap();
-        assert_eq!(stmt.insert(&[1i32]).unwrap(), 1);
-        assert_eq!(stmt.insert(&[2i32]).unwrap(), 2);
-        match stmt.insert(&[1i32]).unwrap_err() {
+        assert_eq!(stmt.insert(&[&1i32]).unwrap(), 1);
```
gwenn, line 1204: Oh no !

src/statement.rs
```diff
@@ -1165,28 +1320,40 @@ mod test {
         .unwrap();
         // existing collection:
         let data = vec![1, 2, 3];
-        db.query_row("SELECT ?1, ?2, ?3", &data, |row| row.get::<_, u8>(0))
-            .unwrap();
-        db.query_row("SELECT ?1, ?2, ?3", data.as_slice(), |row| {
+        db.query_row("SELECT ?1, ?2, ?3", params_from_iter(&data), |row| {
```
gwenn, line 1323: Oh no !

thomcc: Do you think this is that bad? The case where you actually are passing in a list that isn't either `&[&dyn ToSql]` is pretty rare IME — the only cases where you actually need to do that are cases where you're dynamically generating SQL I think

gwenn (approved):
Maybe we should replace all `&[&\(\d\)+` by `[\1` where possible ?
And keep only the easy way to convert a `Vec` to parameters ?

thomcc:
> And keep only the easy way to convert a `Vec` to parameters ?

I could add an extension trait that makes it easier to convert iterators into parameters if you think the `params_from_iter` is too awkward.

```rust
trait IterExtension {
    fn to_params(self) -> ParamsFromIter<Self>;
}
```

And then impl it for all `I where I: IntoIterator, I::Item: ToSql`. This would let you do `slice.to_params()` or `vec.to_params()`. I don't really use extension traits myself very much but I wouldn't mind adding this.

Unfortunately, I don't see any other real way to do this honestly without running into coherence issues (most of the designs I came up with for this ended up failing to type check or being really annoying to use)

src/statement.rs
gwenn: Yes I know ! `params!` macro (and all the rest: `IntoIterator`, ...) was introduced only because of this specific case...
But that is fine.

gwenn:
> > And keep only the easy way to convert a `Vec` to parameters ?
> 
> I could add an extension trait that makes it easier to convert iterators into parameters if you think the `params_from_iter` is too awkward.
> 
> ```rust
> trait IterExtension {
>     fn to_params(self) -> ParamsFromIter<Self>;
> }
> ```
> 
> And then impl it for all `I where I: IntoIterator, I::Item: ToSql`. This would let you do `slice.to_params()` or `vec.to_params()`. I don't really use extension traits myself very much but I wouldn't mind adding this.
> 
> Unfortunately, I don't see any other real way to do this honestly without running into coherence issues (most of the designs I came up with for this ended up failing to type check or being really annoying to use)

Just wait and see if anyone complains...

PR #834, opened 2020-11-04
Add `#[inline]` and `#[cold]` in far more places

### LINKED ISSUE #832: More functions should be marked as #[inline]
We have a lot of trivial one-liners that just call into SQLite, construct an object, or return a member — these should be marked as `#[inline]`

We also should mark functions where constant propagation and dead-code elimination could really improve their codegen.

We should be careful to try to avoid too bad of a compile-time hit from this, but admittedly #[inline] only is used in release builds anyway, and these are pretty small.

Anyway, I'll do this after #830 lands.

thomcc:
Fixes #832

I added inline to anything small enough or where the compiler would be able to DCE it. I added cold to functions hit in error paths (even if they're small). I also avoided inlining deprecated functions and std::fmt::Display impls, etc.

This also has a few unrelated changes that I noticed when I went through every file.

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -13,6 +13,7 @@ pub struct Rows<'stmt> {
 }
 
 impl<'stmt> Rows<'stmt> {
+    #[inline]
     fn reset(&mut self) {
         if let Some(stmt) = self.stmt.take() {
             stmt.reset();
@@ -31,6 +32,7 @@ impl<'stmt> Rows<'stmt> {
     /// consider using `query_map` or `query_and_then` instead, which
     /// return types that implement `Iterator`.
     #[allow(clippy::should_implement_trait)] // cannot implement Iterator
+    #[inline]
     pub fn next(&mut self) -> Result<Option<&Row<'stmt>>> {
         self.advance()?;
         Ok((*self).get())
@@ -47,6 +49,7 @@ impl<'stmt> Rows<'stmt> {
     /// }
     /// ```
     // FIXME Hide FallibleStreamingIterator::map
+    #[inline]
     pub fn map<F, B>(self, f: F) -> Map<'stmt, F>
     where
         F: FnMut(&Row<'_>) -> Result<B>,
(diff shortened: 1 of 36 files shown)


src/row.rs
```diff
@@ -127,21 +136,13 @@ pub struct MappedRows<'stmt, F> {
     map: F,
 }
 
-impl<'stmt, T, F> MappedRows<'stmt, F>
```
thomcc, line 130: MappedRows::new isn't needed if we just implement `query_map` in terms of `query(...).mapped`. This is unrelated but accidentally got rolled into this patch. It doesn't change behavior or the public API at all.

Ditto for AndThenRows::new

src/types/mod.rs
```diff
@@ -119,11 +119,11 @@ pub enum Type {
 impl fmt::Display for Type {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match *self {
-            Type::Null => write!(f, "Null"),
-            Type::Integer => write!(f, "Integer"),
-            Type::Real => write!(f, "Real"),
-            Type::Text => write!(f, "Text"),
-            Type::Blob => write!(f, "Blob"),
+            Type::Null => f.pad("Null"),
```
thomcc, line 122: This is unrelated but allows us to work in cases where a align/width/etc format args are provided. Admittedly, IDK how relevant that is here, but it's better to use Formatter::pad for this sort of thing and I noticed it.

thomcc:
This is probably more aggressive about #[inline] than it would be if written from scratch, but this allows a lot of rusqlite code to be much lower overhead compared to direct SQLite api usage when compiled in release mode (#[inline] does nothing in debug mode).

Without this, you could still have that benefit with `lto = true` but most people dont use that.

thomcc:
(Going to land this w/o review, but if you have an issue @gwenn I'm okay with reverting it and filing a PR that goes through review)

gwenn (commented):
LGTM

PR #837, opened 2020-11-05
Make tests return Result

gwenn:


thomcc:
Hmm, unwrap()s mean we see which line actually failed if they hit, which is pretty nice.

The `?` is cleaner though.

thomcc:
Actually, there are a few ways in which I'd like to rework `rusqlite::Error` which would remove that problem, so this is probably fine once it's green.

PR #838, opened 2020-11-06
Rename get_raw to get_ref_unwrap and get_raw_checked to get_ref

gwenn:
And test `From<FromSqlError> for Error`

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -350,8 +350,8 @@ impl<'stmt> Row<'stmt> {
     /// Renamed to [`get_ref_unwrap`](Row::get_ref_unwrap).
     #[deprecated = "Use [`get_ref_unwrap`](Row::get_ref_unwrap) instead."]
     #[inline]
-    pub fn get_raw<I: RowIndex>(&self, idx: I) -> Result<ValueRef<'_>> {
-        self.get_ref(idx)
+    pub fn get_raw<I: RowIndex>(&self, idx: I) -> ValueRef<'_> {
+        self.get_ref_unwrap(idx)
     }
 }
 
--- a/src/row.rs
+++ b/src/row.rs
@@ -303,7 +303,7 @@ impl<'stmt> Row<'stmt> {
     /// This `ValueRef` is valid only as long as this Row, which is enforced by
     /// it's lifetime. This means that while this method is completely safe,
     /// it can be somewhat difficult to use, and most callers will be better
-    /// served by `get` or `get`.
+    /// served by `get` or `get_unwrap`.
     ///
     /// ## Failure
     ///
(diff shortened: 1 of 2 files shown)


src/row.rs
```diff
@@ -312,7 +312,7 @@ impl<'stmt> Row<'stmt> {
     ///
     /// Returns an `Error::InvalidColumnName` if `idx` is not a valid column
     /// name for this row.
-    pub fn get_raw_checked<I: RowIndex>(&self, idx: I) -> Result<ValueRef<'_>> {
+    pub fn try_get_raw<I: RowIndex>(&self, idx: I) -> Result<ValueRef<'_>> {
```
thomcc, line 315: Hmm, for non-raw gets it's `Row::get` vs `Row::get_unwrap`. Maybe this should be `get_raw` and `get_raw_unwrap`?

gwenn: It means that `get_raw_checked` becomes `get_raw` and `get_raw` becomes `get_raw_unwrap` !

gwenn: Also:
https://docs.rs/postgres/0.18.1/postgres/row/struct.Row.html#method.try_get
https://docs.rs/sqlx/0.3.5/sqlx/row/trait.Row.html#method.try_get

thomcc: Not sure I follow.

And IDK if sqlx/postgres are relevant — I think being consistent with non-raw gets is more important than following the names other crates use.

dubiousjim:
To avoid the pain of changing the sig of `get_raw`, why don't you instead do something like:

* rename `get_raw_checked` to `get_ref` or some other new name
* rename `get_raw` to `get_ref_unwrap`
* deprecate `get_raw` and `get_raw_checked`

thomcc:
Hmm, I don't exactly remember my hesitation here. Well, it's not like the current release isn't a major breaking change anyway, we should probably just land it as-is.

thomcc (approved):
LGTM as-is. Land unless there's something holding you back.

gwenn:
> LGTM as-is. Land unless there's something holding you back.

For me, to ease migration of existing code, we should avoid doing both:
* renaming `old_method_name_1` to `new_method_name_1` (`get_raw` => `get_raw_unwrap`)
* renaming `old_method_name_2` to `old_method_name_1` (`get_raw_checked` => `get_raw`)
Because if you do first `s/\<get_raw_checked\>/\<get_raw/>/g` and then `s/\<get_raw\>/\<get_raw_unwrap/>/g` inadvertently, you are in trouble.

thomcc:
Fair enough. The suggestion https://github.com/rusqlite/rusqlite/pull/838#issuecomment-747147992 works too.

PR #845, opened 2020-11-16
add extra features for bundled sqlite

Dushistov:
This PR add some extra features that I need from bundled version.

I need fts4 feature, and because of:
> When this option is defined in the amalgamation, versions 3 and 4 of the full-text search engine is added to the build automatically
from https://www.sqlite.org/compile.html

this patch should not break any code.


thomcc (approved):
Hm, we already bundle the rest of FTS3 and FTS5 so I guess this is fine.

gwenn:
https://sqlite.org/mark/compile.html?Because*disabled#enable_fts3_tokenizer
> Because of security concerns, the two-argument fts3_tokenizer() feature was disabled
You are sure ?

thomcc:
Oof, nice catch!

thomcc (commented):
Sorry, given the security concerns I don't think we'll take this.

I think you can still use the `LIBSQLITE3_FLAGS` env var to configure this, instead.

PR #855, opened 2020-12-04
Use a weaker dependency on smallvec.

emilio:
There's no reason to use 1.4, 1.0.0 works just as well.

1.4 causes a big perf regression in Firefox, see
https://github.com/servo/rust-smallvec/issues/243, so while we figure
that out we'd like to keep using 1.3.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,7 +109,7 @@ fallible-iterator = "0.2"
 fallible-streaming-iterator = "0.1"
 memchr = "2.3"
 uuid = { version = "0.8", optional = true }
-smallvec = "1.4"
+smallvec = "1.0"
 
 [dev-dependencies]
 doc-comment = "0.3"


emilio:
r? @gwenn or someone else?

If a dot-release could be made with this change it'd be fairly amazing.

gwenn:
`version = ">=1.3.0, <1.4.0"` ?

thomcc:
This evening I can apply this and cut a release with it and no other changes from the prev.

(We have had a substantial breaking change so it couldn't be a patch release without that, and I don't want to cut the next major *quite* yet)

thomcc:
1.0 is fine if it compiles (or whichever got rid of the soundness holes by using MaybeUninit)

emilio:
> (We have had a substantial breaking change so it couldn't be a patch release without that, and I don't want to cut the next major _quite_ yet)

Yeah, that sounds great, thanks!

> 1.0 is fine if it compiles (or whichever got rid of the soundness holes by using MaybeUninit)

Yeah, it does, I checked out this repo, `cargo build`, `cargo update --precise 1.0.0 -p smallvec`, then `cargo build` again.

Keeping 1.0 in the dependency gives us flexibility to update when the bug is fixed as well, so I'd prefer this to `>=1.3.0, <1.4.0`.

Thanks.

thomcc:
@emilio This is in 0.24.2, which I just released.

PR #857, opened 2020-12-05
Upgrade chrono to 0.4.19

erictapen:
This fixes an annoying and hard to debug error `the trait is not implemented`, as `rusqlite` uses an 3 years old `chrono`, that used to expose a module `chrono::datetime` which is now private.

```rust
#[derive(Debug)]
struct Focus {
    datetime: DateTime<Local>,
    title: String,
}

...

    let mut query = db.prepare("SELECT datetime, title FROM titles").unwrap();
    let focus_iter = query
        .query_map(params![], |row| {
            Ok(Focus {
                datetime: row.get(0)?,
                title: row.get(1)?,
            })
        })
        .unwrap();
```

```
error[E0277]: the trait bound `chrono::datetime::DateTime<chrono::offset::local::Local>: rusqlite::types::from_sql::FromSql` is not satisfied
  --> src/client.rs:24:31
   |
24 |                 datetime: row.get(0)?,
   |                               ^^^ the trait `rusqlite::types::from_sql::FromSql` is not implemented for `chrono::datetime::DateTime<chrono::offset::local::Local>`
```

Didn't test this beyond that it builds and runs fine in my project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ bundled-full = [
 time = { version = "0.2", optional = true }
 bitflags = "1.2"
 hashlink = "0.6"
-chrono = { version = "0.4", optional = true }
+chrono = { version = "0.4.19", optional = true }
 serde_json = { version = "1.0", optional = true }
 csv = { version = "1.1", optional = true }
 url = { version = "2.1", optional = true }


gwenn:
I guess the issue is somewhere else because version`0.4.19` is compatible with version `0.4`.

erictapen:
Oops, just realised, that what fixed it for me was `features = [ "chrono" ]`, not the version bump.

But I guess it does no harm to have a more recent verson of chrono?

Maybe it could also be documented on [`FromSql`](https://docs.rs/rusqlite/0.24.2/rusqlite/types/trait.FromSql.html) that the feature needs to be enabled?

chansuke:
@erictapen Thanks for the PR and the comment above helped me a lot(I've wasted 2hours for that `not implemented` error).

> Maybe it could also be documented on FromSql that the feature needs to be enabled?

I agree that the documentation will be really useful.

gwenn:
> > Maybe it could also be documented on FromSql that the feature needs to be enabled?
> 
> I agree that the documentation will be really useful.

See #857

PR #860, opened 2020-12-08
Add bundle-sqlcipher and bundle-ssl

dubiousjim:
This reworks (and replaces) #771 on top of a more recent master. See that PR for some earlier discussion.

As mentioned there, we should confirm that Zetetic is ok with us re-distributing their amalgamated sources (which have to be built from sqlcipher/sqlcipher).

Would fix #765.

PR #861, opened 2020-12-08
Miscellaneous cleanup

dubiousjim:
Expands/cleans up documentation. Also renames (and documents) the so-far-undocumented LIBSQLITE3_FLAGS.

This PR lays groundwork for #860, but should be useful cleanup even without that.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -138,18 +138,18 @@ mod build_bundled {
         }
         println!("cargo:rerun-if-env-changed=SQLITE_MAX_EXPR_DEPTH");
 
-        if let Ok(extras) = env::var("LIBSQLITE3_FLAGS") {
+        if let Ok(extras) = env::var("SQLITE_FLAGS") {
             for extra in extras.split_whitespace() {
                 if extra.starts_with("-D") || extra.starts_with("-U") {
                     cfg.flag(extra);
                 } else if extra.starts_with("SQLITE_") {
                     cfg.flag(&format!("-D{}", extra));
                 } else {
-                    panic!("Don't understand {} in LIBSQLITE3_FLAGS", extra);
+                    panic!("Don't understand {} in SQLITE_FLAGS", extra);
                 }
             }
         }
-        println!("cargo:rerun-if-env-changed=LIBSQLITE3_FLAGS");
+        println!("cargo:rerun-if-env-changed=SQLITE_FLAGS");
 
         cfg.compile("libsqlite3.a");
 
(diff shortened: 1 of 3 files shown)


libsqlite3-sys/build.rs
```diff
@@ -138,18 +138,18 @@ mod build_bundled {
         }
         println!("cargo:rerun-if-env-changed=SQLITE_MAX_EXPR_DEPTH");
 
-        if let Ok(extras) = env::var("LIBSQLITE3_FLAGS") {
```
thomcc, line 141: This is breaking, and possibly silently so, so I'd prefer some justification for it.

If we're going to break that, we should error on the old variable rather than ignore it, I think

dubiousjim: Makes sense.

I pushed this env flag back in the spring but never got around to documenting it anywhere (save in comments in the source, and a bullet in the relnotes for 0.24.0). I wanted to document it now, and just thought the naming scheme I had chosen earlier was weird.

Summarizing the other variables honored by the build script: there are variables like (a) SQLITE3_LIB_DIR and (b) SQLCIPHER_LIB_DIR (also ..._INC_DIR and ..._STATIC) --- these are all used by our build/link script; there are (c) SQLITE_FOO compile-time options honored by sqlite/sqlcipher's own sources, and there is (d) the hackish LIBSQLITE3_SYS_BUNDLING used by our build/link script when we're generating the bundled bindings. I initially shied from using a name on pattern (c), because the new flag would be one step removed from the ones that SQLite's own sources see. I shied from using (a)/(b) because I wanted a single variable, not a different one for SQLITE3 and for SQLCIPHER. By elimination, I then settled on using pattern (d). But now that I was moving to write up some documentation for this flag, it struck me how unnatural this choice was. So I persuaded myself it was easiest to just go with pattern (c) after all. The `_FLAGS` suffix on the name is on the model of `CFLAGS` etc.

Happy to go with whatever naming scheme you like here. I'm not attached to anything. I agree with you if we change it it'd be a good idea to emit a warning or error when the old name is used.

I doubt many people are using this flag yet. Googling doesn't turn up any examples.

thomcc: I've definitely pointed people to it in the past when they've asked for additions.

In general it's kind of redundant with the `cc` crate's CFLAGS support https://github.com/alexcrichton/cc-rs#external-configuration-via-environment-variables, but I guess it's SQLite specific.

Anyway, I'd prefer either we don't change this (ideally), or we change it and error on the old value. A warning would be preferrable, but there isn't a mechanism for us to emit one that a dependent would see — `println!("cargo:warning=...")` is only for crates within the same workspace.

dubiousjim: Reverting name back to original, and adjusting the docs...

gwenn (approved):

Issue #643, opened 2020-02-28
Expose sqlite3_context_db_handle


phiresky:
Within a user-defined function, it's possible to get the handle of the DB using sqlite3_context_db_handle. This is not possible from rusqlite currently. This is useful when the function is not defined as a lambda, because then it doesn't have access to the closure where the db connection is probably defined. I need this specifically when creating a sqlite extension (#631), though I'm sure it's useful elsewhere.

Here's an implementation that seems to be working, though I'm not sure if it's sound (are multiple Connection objects for one *mut sqlite3_handle valid?)

https://github.com/jgallagher/rusqlite/blob/022266239233857faa7f0b415c1a3d5095d96a53/src/functions.rs#L124

```rust
    pub fn get_connection(&self) -> Result<Connection> {
        let handle = unsafe { ffi::sqlite3_context_db_handle(self.ctx) };
        unsafe { Connection::from_handle(handle) }
    }
```

thomcc:
> (are multiple Connection objects for one *mut sqlite3_handle valid?)

I don't think so, I'm pretty sure it has to be owning.

gwenn:
I guess it is OK only a `Connection` reference is returned bound to `Context` lifetime.
And `Connection::from_handle` allocates its own statement cache which may be useless.

phiresky:
A reference that only lives for the duration of the callback is fine. Though in my case I'm starting a transaction within the user-defined-function, and `conn.transaction()` requires a mutable reference to the connection..

gwenn:
Would you mind sending a PR ?

phiresky:
Sure, I have the code I'm currently using here: https://github.com/phiresky/rusqlite/commits/pr/loadable-extensions but

1. it's based on the loadable extensions pr (could be rebased I guess)
2. I'm not sure how to do it without using Connection::from_handle (because of the problem you mentioned above) and also how to return a mutable reference correctly (is that even possible without breaking mutable reference assumptions? Never really used ffi)

thomcc:
> I guess it is OK if only a Connection reference is returned, bound to Context lifetime.

I don't know for sure if this is valid or not. It's tricky. You mean something like:

```rust
pub struct ConnectionRef<'ctx> {
	// comes from Connection::from_handle(sqlite3_context_db_handle(...))
    // and is non-owning
    conn: Connection,
    phantom: PhantomData<&'ctx Context<'ctx>>,
}
```

[...]

PR #866, opened 2020-12-18
Pass context to aggregate init and finalize

### LINKED ISSUE #865: Aggregate Function API: Expose Context in both init and finish function
The rusqlite::functions::Aggregate<> interface adds a init() function that is not present in pure sqlite. This is nice because it means you can handle the initialisation of the user context separately from the step function. 

However, in practice this does not work that well since the init function is not passed the `sqlite3_context` so it can't create the state based on the function parameters or aux data. At least for me it just means I have to wrap stuff in more Option<>s and save a handle to the context myself: https://github.com/phiresky/sqlite-zstd/blob/aeb5466d97f23439e143dfb44b9af46859b7c2ff/src/dict_training.rs#L30

For the init function this is easy to fix:

https://github.com/phiresky/rusqlite/commit/5a2200e24d9ce6050235954ee87a7ee525c231e8

For the finish function it's somewhat unclear what the best move is, since the rusqlite context also has the args, which are not passed to the finish function. There's two possibilities here:

1. Set args to empty vec in the context passed to finish function and document that (or throw)
2. Separate the args from the context object (somewhat cleaner, but more complex and breaking)

I'll make a PR for both, but it would be good to get feedback first.

phiresky:
Fixes #865.

Passing to init is useful as is, passing to finalize will only be useful once #643 is fixed.

This is a breaking change, though fixing the breakage is trivial.

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -911,7 +911,7 @@ mod test {
     struct Count;
 
     impl Aggregate<i64, Option<i64>> for Sum {
-        fn init(&self) -> i64 {
+        fn init(&self, _: &mut Context<'_>) -> i64 {
             0
         }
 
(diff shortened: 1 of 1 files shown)


gwenn:
I search for an aggregate function in SQLite source which uses auxiliary data but found none !

phiresky:
For me the reason I want this is that I need to do an SQL query in the finish function - for which I need the context handle and #867 .
The reason I want it in init is that some of function parameters are pretty much assumed to be constants and needed to configure the aggregate function during initialization

gwenn (approved):

PR #867, opened 2020-12-18
add get_connection method to function context

### LINKED ISSUE #643: Expose sqlite3_context_db_handle

phiresky:
This method is marked unsafe because it internally creates a new Connection object which may have a different state than the originally created one (especially when that one is sent to another thread).

More detail by @thomcc here: https://github.com/rusqlite/rusqlite/issues/643#issuecomment-640181213

I'm not sure if it's possible to fix that and make this fully safe. One way I could think of is to have a global list of existing connections, and in the get_connection method look for the existing connection with the same internal sqlite object, then set a flag in that connection to disallow use in some way. Or possibly in the from_handle function throw if a handle for the given sqlite ffi already exists.

This fixes #643

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -58,6 +58,8 @@ use std::panic::{catch_unwind, RefUnwindSafe, UnwindSafe};
 use std::ptr;
 use std::slice;
 use std::sync::Arc;
+use std::marker::PhantomData;
+use std::ops::Deref;
 
 use crate::ffi;
 use crate::ffi::sqlite3_context;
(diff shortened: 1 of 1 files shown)


thomcc (commented):
LGTM with lints fixed

thomcc:
I'm more comfortable with this as unsafe -- I think its fine to just make the last paragraph the `# Safety` section.

phiresky:
Lints are fixed

thomcc (approved):

PR #887, opened 2021-01-30
Implement AsRef<Statement> for Row(s)

gwenn:
I couldn't use `Statement::expanded_sql` while debugging:
```rust
statenent.query_map([test_date], |row| {
     // Impossible to see expanded sql here
 });
```
Breaking changes: `Rows::column*` and `Row::column*` removed now that you have access to underlying statement.


thomcc:
This seems fine if you want to merge it, but I vaguely remember something was only sound because we didn't expose this, so will want to do a quick pass over this code. I have a lot of free time next week to do this though.

gwenn:
> This seems fine if you want to merge it, but I vaguely remember something was only sound because we didn't expose this, so will want to do a quick pass over this code between then and releasing. I have a lot of free time next week to do this though.

#636 ?

thomcc:
I think the issue was more along the lines of: what happens if you call step/reset or something else that causes that (like triggering another query) inside the closure?

gwenn:
> I think the issue was more along the lines of: what happens if you call step/reset or something else that causes that (like triggering another query) inside the closure?

You need a mutable reference to call step/reset, no ?

thomcc:
Not internally, but it's possible that in the public APIs we do https://github.com/rusqlite/rusqlite/blob/master/src/statement.rs#L897

thomcc (approved):

PR #888, opened 2021-01-30
Fix unstable tests when machine is slow

gwenn:
I only guess...
See https://ci.appveyor.com/project/thomcc/rusqlite/builds/37525246


thomcc (commented):
Nice catch.

PR #891, opened 2021-01-31
Fix some warnings from nightly clippy

thomcc:
Happened to be using nightly and noticed them.


thomcc:
I guess this lint isn't known outside of nightly, so I'll close for now...

PR #896, opened 2021-02-02
Fix smallvec version

gwenn:
https://rustsec.org/advisories/RUSTSEC-2021-0003.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,7 +109,7 @@ fallible-iterator = "0.2"
 fallible-streaming-iterator = "0.1"
 memchr = "2.3"
 uuid = { version = "0.8", optional = true }
-smallvec = "1.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -109,7 +109,7 @@ fallible-iterator = "0.2"
 fallible-streaming-iterator = "0.1"
 memchr = "2.3"
 uuid = { version = "0.8", optional = true }
-smallvec = "1.0"
+smallvec = ">=1.6.1"
```
vilgotf, line 112: This should just be written as "1.6.1" which implies >=1.6.1 <2.0.0, see [doc](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

gwenn: You are right.

PR #897, opened 2021-02-02
[WIP] Initial proposal for the preupdate hook.

MidasLamb:
I created an initial draft for the preupdate hook that seems to work ok after preliminary testing.

Some questions arose and I'm looking for some input/feedback:

1. Does it needs it own feature flag instead of being bundled with all the other hooks (it needs a special feature flag on the `libsqlite3-sys` crate).
    * If it's behind a separate one, I'll probably have to extract some of the `hooks` code to be included if one or the other is active.
2. How close do we want to map the original API:
    * According to [SQLite3 docs](https://www.sqlite.org/c3ref/preupdate_count.html), the 6th and 7th parameter of the original API are dependent (and can be undefined) on the type of operation. I think it would be more idiomatic Rust if this was exposed through an enum instead of closely mapping the API.
1. During the preupdate_hook, you have access to some special functions ([see docs again](https://www.sqlite.org/c3ref/preupdate_count.html)), but it depends on which type of operation is done.
    - Here I was thinking to use an enum for which the variants expose the appropriate functions, to prevent UB from occurring and provide a safe interface.

If anyone can help me and give some feedback on what the best way forward would be, that would be much appreciated.


gwenn:
May I suggest that `hooks` is kept untouched and a new feature `pre_update_hook = ["libsqlite3-sys/pre_update_hook", "hooks"]` is used instead (`libsqlite3-sys/pre_update_hook` should be used to activate the `SQLITE_ENABLE_PREUPDATE_HOOK` compile-time option).

MidasLamb:
@gwenn , of course you can suggest that, that is also the first point I touch in my initial message. However the "problem" with that is that it might be confusing that `hooks` does not encompass all hooks. Of course changing this (i.e. rename `hooks` to something more specific so it doesn't encompass the preupdate hook) would be a breaking change.

This is my first time writing something that has an interface with C-code so I'm being very cautious and trying to make as little change as possible at a time.

gwenn:
> However the "problem" with that is that it might be confusing that `hooks` does not encompass all hooks.

By default, SQLite does not encompass all hooks so it seems "normal" that rusqlite does the same, no ?

[go-sqlite3](https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_preupdate_hook.go#L7)
[tclsqlite](https://github.com/sqlite/sqlite/blob/master/src/tclsqlite.c#L3492)

MidasLamb:
> By default, SQLite does not encompass all hooks so it seems "normal" that rusqlite does the same, no ?

The difference (unless I'm mistaken about SQLite itself), is that SQLite will have some hooks by default and you can add a flag for the preupdate_hook. But in rusqlite you have to specify that you want hooks (by using the `hooks` feature), but then there are hooks that are not in there (i.e. the preupdate hook).

Just to clarify, I am completely on board with having a separate flag for the preupdate hook since it requires a special compilation of SQLite, but I'm concerned that it might not be clear if or how these different features interact.

src/hooks.rs
```diff
@@ -36,365 +31,555 @@ impl From<i32> for Action {
...
+            db.execute_batch("INSERT INTO foo VALUES ('lisa')")?;
+            assert!(called);
+            Ok(())
         }
-        self.free_commit_hook = free_commit_hook;
     }
+}
+
+#[cfg(feature = "hooks")]
+mod datachanged_and_friends {
```
MidasLamb, line 214: This is just a preliminary name, I couldn't come up with a succint naming for all the hooks, excluding the preupdate_hook.

src/hooks.rs
```diff
@@ -36,365 +31,555 @@ impl From<i32> for Action {
...
-    /// `handler`. If `num_ops` is less than one then the progress handler
-    /// is disabled.
-    ///
-    /// If the progress callback returns `true`, the operation is interrupted.
-    pub fn progress_handler<F>(&self, num_ops: c_int, handler: Option<F>)
-    where
-        F: FnMut() -> bool + Send + RefUnwindSafe + 'static,
-    {
-        self.db.borrow_mut().progress_handler(num_ops, handler);
+    impl PreUpdateHookFunctions {
```
MidasLamb, line 57: This is where I'm not 100% sure how to proceed. These functions are accessible based on the type of action (insert, delete, or update), during the preupdate_hook callback. There are multiple ways to resolve this:

- Pass Option<Fn> to the callback depending on the function
- Pass an Enum for which the value is based on the Action, and the object inside that Enum implements the relevant functions (similar to what is happening now only with an added layer and removal of the unsafe on the function itself.
- Keep it like this and hold the end user responsible for correctly using the functions.

MidasLamb:
@gwenn , I just tried to run this hook with a lot of inserts/updates/deletes, and during update after a while it suddenly runs into a SIGSEGV, with gdb this is the top of the stack trace:
image

Do you have any idea what could cause this?

Issue #312, opened 2017-11-23
Execute statement with variable number of parameters?


arthurmilliken:
I'm a rust newbie, so please forgive my ignorance if this is a stupid question...

How do I execute statements with a variable number of parameters? Let's say that the statement is:

`"INSERT INTO TableName(ColumnA, ColumnB, ColumnC) VALUES (?1, ?2, ?3)"`

But `TableName`, Columns, and number of columns are not known at compile time. Is there a way of coercing an iterator over a collection of strings into a `&[&ToSql]` type?

SolraBizna:
Greetings, fellow Rust newbie. The following should work:

```rust
let params: Vec<&ToSql> = my_iterator.collect();
my_statement.execute(params.as_slice());
```

Hope it's still useful months later.

arthurmilliken:
Hi @SolraBizna , this does not seem to work:
`
error[E0308]: mismatched types
  --> src/load/mod.rs:61:34
   |
61 |     conn.execute(&schema.insert, fields.as_slice())?; // THIS LINE FAILS
   |                                  ^^^^^^^^^^^^^^^^^ expected trait load::rusqlite::types::ToSql, found str
   |
   = note: expected type '&[&load::rusqlite::types::ToSql]'
              found type '&[&str]'
`

Here is my source:
`
  for record in rdr.records() {
    let row = record?;
    println!("{}", schema.insert);
    let fields: Vec<&str> = row.iter().collect();
    println!("            {:?}", fields);
    conn.execute(&schema.insert, fields.as_slice())?; // THIS LINE FAILS
  }
`

gwenn:
Not related to rusqlite but to Rust :
```
use std::fmt::Debug;
fn test(values: &[&Debug]) {
    println!("{:?}", values);
}
fn main() {
    let fields = vec!["field1", "field2"];
    test(fields.as_slice()); // THIS LINE FAILS
}
```
And I don't know the correct/idiomatic way to cast a `&[&str]` into a `&[&Debug]`.

arthurmilliken:
I suspect it is impossible, since the Rust compiler cannot determine the correct memory allocation at compile time.

gwenn:
This works:
```rust
use std::fmt::Debug;
fn test<T: Debug + ?Sized>(values: &[&T]) {
    println!("{:?}", values);
}
fn main() {
    let fields = vec!["field1", "field2"];
    test(fields.as_slice()); // THIS LINE FAILS
}
```
[...]

PR #908, opened 2021-03-03
Improve flexibility of named params

### LINKED ISSUE #907: How do I make a (variable length) set of named parameters at runtime?
I need to generate a query based on user input and I can't figure out how to build an object that will be accepted as named parameters. For unnamed parameters I was able to pass a `&Vec<Box<dyn ToSql>>`, but for the functions that take named parameters, `&Vec<(&str, Box<dyn ToSql>)>` gives a type mismatch error. I'm not sure where to go from here.

thomcc:
Fixes #907... but I guess maybe they don't need it.

My main concern with this change is that it might make type inference work less well for users not using `named_params!`. It also might make the docs less clear...

What do you think @gwenn. The use case for this is pretty niche, is it worth supporting?

### DIFF
--- a/src/params.rs
+++ b/src/params.rs
@@ -188,8 +188,8 @@ impl Params for &[&dyn ToSql] {
     }
 }
 
-impl Sealed for &[(&str, &dyn ToSql)] {}
-impl Params for &[(&str, &dyn ToSql)] {
+impl<T: ToSql> Sealed for &[(&str, T)] {}
+impl<T: ToSql> Params for &[(&str, T)] {
     #[inline]
     fn __bind_in(self, stmt: &mut Statement<'_>) -> Result<()> {
         stmt.bind_parameters_named(self)
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -539,11 +539,8 @@ impl Statement<'_> {
     }
 
     #[inline]
-    pub(crate) fn bind_parameters_named<T: ?Sized + ToSql>(
-        &mut self,
-        params: &[(&str, &T)],
-    ) -> Result<()> {
-        for &(name, value) in params {
+    pub(crate) fn bind_parameters_named<T: ToSql>(&mut self, params: &[(&str, T)]) -> Result<()> {
+        for &(name, ref value) in params {
             if let Some(i) = self.parameter_index(name)? {
                 let ts: &dyn ToSql = &value;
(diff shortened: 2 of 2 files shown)


gwenn:
Is it the same issue as described here:
https://github.com/rusqlite/rusqlite/issues/312#issuecomment-386396277
?

thomcc:
> Is it the same issue as described here:
> [#312 (comment)](https://github.com/rusqlite/rusqlite/issues/312#issuecomment-386396277)
> ?

It's similar, but with named params.

gwenn:
> Maybe handling it the way iterables are handled would be better...

You mean like [ParamsFromIter](https://github.com/rusqlite/rusqlite/blob/master/src/params.rs#L329) ?

```rust
pub struct NamedParamsFromIter<I>(I);

pub fn named_params_from_iter<I>(iter: I) -> ParamsFromIter<I>
where
    I: IntoIterator,
    I::Item: (&str, ToSql), // or T: ToSql
{
    NamedParamsFromIter(iter)
}
```

thomcc:
Basically. I think we could solve the &str vs String case too like:

```rust
pub struct NamedParamsFromIter<I>(I);

pub fn named_params_from_iter<I, S, T>(iter: I) -> ParamsFromIter<I>
where
    I: IntoIterator<Item = (S, T)>,
    S: AsRef<str>,
    T: ToSql,
{
    NamedParamsFromIter(iter)
}
```

(Haven't actually tried this code)

thomcc:
The use case here seems pretty niche still though, but it's not like there's a good way to do it otherwise. Hrm.

gwenn:
Do you know if something like [BorrowToSql](https://docs.rs/postgres/0.19.0/postgres/types/trait.BorrowToSql.html) would help here ?
> A trait used by clients to abstract over `&dyn ToSql` and `T: ToSql`

hutchisr:
@thomcc Less that I don't "need" it as much as it is possible to work around by using numbered parameters. The ability to do this would still be very much appreciated if it doesn't end up breaking other functionality.

mzr:
@thomcc are you thinking of finishing this PR? This looks exactly like something I was looking for.

thomcc:
I'll look into finishing it this weekend.

gwenn:
See #1652 with BindIndex

PR #911, opened 2021-03-07
Fix example code in README

galeone:
Currently, the code in the README doesn't even compile.

```
error[E0277]: `[_; 0]` is not an iterator
  --> src/main.rs:19:9
   |
19 |         [],
   |         ^^ borrow the array with `&` or call `.iter()` on it to iterate over it
   |
   = help: the trait `Iterator` is not implemented for `[_; 0]`
   = note: arrays are not iterators, but slices like the following are: `&[1, 2, 3]`
   = note: required because of the requirements on the impl of `IntoIterator` for `[_; 0]`

error[E0277]: `[_; 0]` is not an iterator
  --> src/main.rs:32:38
   |
32 |     let person_iter = stmt.query_map([], |row| {
   |                                      ^^ borrow the array with `&` or call `.iter()` on it to iterate over it
   |
   = help: the trait `Iterator` is not implemented for `[_; 0]`
   = note: arrays are not iterators, but slices like the following are: `&[1, 2, 3]`
   = note: required because of the requirements on the impl of `IntoIterator` for `[_; 0]`

error: aborting due to 2 previous errors
```

This merge request update the code for making it compile with the latest version

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Rusqlite is an ergonomic wrapper for using SQLite from Rust. It attempts to expo
 an interface similar to [rust-postgres](https://github.com/sfackler/rust-postgres).
 
 ```rust
-use rusqlite::{params, Connection, Result};
+use rusqlite::{params, Connection, Result, NO_PARAMS};
 
 #[derive(Debug)]
 struct Person {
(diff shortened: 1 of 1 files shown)


galeone:
Closing since I just saw that's the expected behavior

oconnor663:
I just ran into the same thing. Where did you see that this is expected? I must be missing a comment or something.

thomcc:
The latest version on github master is not the same as the latest version on crates.io. It includes changes that make what's in the readme work.

PR #912, opened 2021-03-08
silence clippy::needless-update warning in vtab module creation functions

jrandall:
When the ffi bindings specify exactly the set of sqlite3_module fields we explicitly list in the vtab module creation functions, clippy warns on a `needless_update` lint (https://rust-lang.github.io/rust-clippy/master/#needless_update). 

This fix is just to allow that lint since we don't have a good way to tell what version of sqlite has been built against.


gwenn:
See rustfmt check

jrandall:
> See rustfmt check

removed the comment to fix the rustfmt issue

PR #918, opened 2021-03-15
Doctest column name reference

gwenn:


src/column.rs
```diff
@@ -42,6 +42,38 @@ impl Statement<'_> {
...
+    /// ```
+    /// use rusqlite::{Connection, Result};
+    /// fn main() -> Result<()> {
+    ///     let db = Connection::open_in_memory()?;
+    ///     db.execute_batch("CREATE TABLE y (x);")?;
+    ///     let mut stmt = db.prepare("SELECT x FROM y;")?;
+    ///     let column_name = stmt.column_name(0)?;
+    ///     assert_eq!("x", column_name);
+    ///     db.execute_batch("ALTER TABLE y RENAME COLUMN x TO z;")?;
+    ///     // column name is not refreshed until statement is re-prepared
```
weiznich, line 71: I think this should at least include a comment stating that this is testing that the internal behaviour of the SQLite implementation did not change, as the documentation states that the returned pointer is only valid till the next call to `sqlite3_column_name()` for the same column.
For diesel I've decided to not relay on the SQLite internals but to follow the documented behaviour by just caching the necessary column name pointers, so that `sqlite3_column_name()` is called maximally once. See https://github.com/diesel-rs/diesel/pull/2676 for details.

gwenn: columns name / count cannot be cached because SQLite will automatically re-prepare the statement in case of schema is altered.
https://github.com/rusqlite/rusqlite/blob/master/src/raw_statement.rs#L64
https://github.com/mattn/go-sqlite3/issues/240

weiznich: Maybe I misread your argument, but for me this sounds like this would have really bad implications. It basically means no rust interface can return a `&str` based on the original column name from sqlite without copying it to some internal data structure as there is no lifetime to describe the returned pointer in a static way.
For this test case this would mean calling `column_name()`/`column_names()` before /after such a modification will result in a pointer invalidation as SQLite could automatically re-prepare the statement at any time. This would also imply that the documentation on the underlying `sqlite3_column_name()` function is not accurate, as this clearly states the returned pointer is valid till one of the following things happen:
* The prepared statement is destroyed
* Someone called `sqlite3_step()` for the first time on the prepared statement after creating it or after resting it
* Till someone calls `sqlite3_column_name()` again on the same column.

So if that's true that SQLite will silently re-prepare the statement behind the scenes if the schema is altered that would be another point on this list.
If you just mean that it is not possible to cache the result for multiple runs of the statement, then that's no new news for me as this is already covered by the existing documentation and diesel already handles this by different types for prepared statements and statements that are currently executed. Only the second kind does the column name caching for a single execution of the statement.

gwenn: rusqlite `Stmt` cannot be re-prepared without using a mutable reference.
So we are safe.

gwenn: https://github.com/sqlite/sqlite/blob/master/src/legacy.c#L65-L117
After the first call to `sqlite3_step`, a read lock prevents any `ALTER` so reference to `sqlite3_column_name` (acquired **after** `sqlite3_step`) can be safely reused until `sqlite3_reset` / `sqlite3_finalize`.

gwenn: diesel `StatementUse` is always created after an `sqlite3_step` unlike rusqlite `Rows`?

weiznich: Its created before the first call to step, but it internally tracks if step was already called an just does not return a column name before that. Additionally its not exposed as public API. The row type that is exposed can only be created by calling step, so there is basically no way that someone can request a column name before calling step.

PR #924, opened 2021-03-25
Silent some clippy warnings

gwenn:
* allow(clippy::upper_case_acronyms) for rust enum entries that match
  SQLite constants.
* allow(clippy::needless_return) for collation_needed_callback until we
  find a way to propagate the error.


thomcc (approved):

gwenn:
I don't know why but I could not reproduced some clippy warnings locally... Hence the several commits.
And it seems to me that `allow(non_snake_case, non_camel_case_types)` should imply `allow(clippy::upper_case_acronyms)`...

thomcc:
No worries.

And I agree that allowing some warnings should imply allowing other warnings more of the time...

PR #925, opened 2021-03-26
Add missing params! in README's example

Canop:
Right now the README's example doesn't compile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ fn main() -> Result<()> {
                   name            TEXT NOT NULL,
                   data            BLOB
                   )",
-        [],
+        params![],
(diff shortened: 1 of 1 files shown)


thomcc:
It does on nightly.

thomcc:
Sorry, we should be releasing something soon. I've been busy, unfortunately.

You should use the readme text here, which is copied from when the current version of the crate was released: https://crates.io/crates/rusqlite

If you're going to use stuff from the readme from the git repository, you need to use a git dependency, e.g. `rusqlite = { git = "https://github.com/rusqlite/rusqlite" }` but I don't recommend this.

PR #927, opened 2021-03-30
allow setting static linkage through feature flag rather than env v…

avnerbarr:
…ariable since difficult to configure env through toml

we need to staticlly link. it seems that it is only possible by setting an environment variable:

`export SQLCIPHER_STATIC=1`

I wanted to configure through `toml`, but wasn't able to add the env variable. Tried through buid.rs, but it ran after the compilation of rustql. 

It wasn't feasible to modify many build scripts to have the export, so this seems the best option (unless it will be possible to add the env variable through toml)


gwenn:
TODO try to find a popular FFI crate with static link activated by a feature.

thomcc:
> TODO try to find a popular FFI crate with static link activated by a feature.

https://github.com/Rust-SDL2/rust-sdl2/blob/master/sdl2-sys/Cargo.toml#L60 is an example. I have mixed feelings about it, but am not really negative about providing this.

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
thomcc, line 1: Please don't add these files to the repository.

thomcc (changes requested):
Please don't add the .idea/* files into this repository.

Dushistov:
The [related issue in cargo](https://github.com/rust-lang/cargo/issues/4121) (Passing environment variables from down-stream to up-stream library) has been resolved by this https://github.com/rust-lang/cargo/pull/9175 . 

So is `[env]` `.cargo/config.toml` is not enough for this case?

thomcc:
> So is [env] .cargo/config.toml is not enough for this case?

It's not, because .cargo/config.toml isn't used when building dependencies.

That said, perhaps it's dodgy to set a feature like that if you aren't the root crate. Hmm.

avnerbarr:
> The [related issue in cargo](https://github.com/rust-lang/cargo/issues/4121) (Passing environment variables from down-stream to up-stream library) has been resolved by this [rust-lang/cargo#9175](https://github.com/rust-lang/cargo/pull/9175) .
> 
> So is `[env]` `.cargo/config.toml` is not enough for this case?

I believe that `env` support is a very new feature (i.e. within the last few weeks) and in any case not yet supported widely (perhaps in latest version only - I couldn't get it to work properly)

So what is the consensus? I'll remove the `.idea` files, but also found an issue in the `build.rs` script where when the feature flag is set there is an issue with the env prefix line:
`println!("cargo:rerun-if-env-changed={}_LIB_DIR", env_prefix());`

need to debug this (not sure how to debug `build.rs` 🤦 )

I'll try to update today.

PR #928, opened 2021-04-01
Leniently parse rfc3339 timezones

nhynes:
https://github.com/rusqlite/rusqlite/pull/886 introduced a breaking change to databases that store RFC3339 datetimes manually without the full timezone (i.e. not strictly through this library, which includes the full timezone).  This PR relaxes the timezone requirement during parsing.

An example of when this might be problematic is when doing something like the following:
```sql
CREATE TABLE table (
  created_at  DATETIME NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ','now')),
  -- as opposed to datetime('now'), which is not RFC3339 compliant
);
```

### DIFF
--- a/src/types/chrono.rs
+++ b/src/types/chrono.rs
@@ -101,9 +101,9 @@ impl FromSql for DateTime<Utc> {
             let s = value.as_str()?;
 
             let fmt = if s.len() >= 11 && s.as_bytes()[10] == b'T' {
-                "%FT%T%.f%:z"
+                "%FT%T%.f%#z"
             } else {
-                "%F %T%.f%:z"
+                "%F %T%.f%#z"
             };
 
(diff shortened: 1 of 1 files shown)


thomcc:
> This PR relaxes the timezone requirement during parsing.

So, just to check, this is in all cases less strict than the previous code?

nhynes:
> So, just to check, this is in all cases less strict than the previous code?

Per [chrono](https://docs.rs/chrono/0.4.19/chrono/format/strftime/index.html)

%#z | +09 | Parsing only: Same as %z but allows minutes to be missing or present.
-- | -- | --

thomcc (approved):
Thanks, looks good to me.

thomcc:
Please fix the clippy and rustfmt complaints.

gwenn (approved):

PR #929, opened 2021-04-03
Upgrade SQLite bundled version to 3.35.4

gwenn:


thomcc:
Sorry I've been busy. If you're interested in cutting a release you should go for it — unfortunately I'm likely to continue to be busy for a bit, and the backlog is getting crazy. No rush though, and it's fine if you don't want to.

gwenn:
I've already prepared the release notes but they are still some pending PRs like #908.
I am going to send a PR to update Cargo files and README anyway...

thomcc:
Thanks. I think we probably shouldn't do #908 for now. We can do it later backwards compatibly anyway.

gwenn:
Fine. See #930

PR #935, opened 2021-04-09
feat: add wasm32-unknown-unknown support and memvfs feature

passchaos:
# Content:

1. Add a memory virtual file system isolated by filename, different from sqlite3's ':memory:' mode
2. Add the ability to make SQLite link to external memory allocator
3. Reimplement the functions used in libc (such as strcspn, qsort)

With these three features: linking sqlite3 to Rust's memory allocator, using an in-memory virtual file system, linking to a separate implementation of string functions and qsort functions, along with fine-tuned WASI headers, our implementation works under `wasm32-unknown-unknown`

# Remind:
Time zone function temporarily missing

### DIFF
--- a/libsqlite3-sys/Cargo.toml
+++ b/libsqlite3-sys/Cargo.toml
@@ -35,6 +35,13 @@ with-asan = []
 wasm32-wasi-vfs = []
 # lowest version shipped with Windows 10.0.10586 was 3.8.8.3
 winsqlite3 = ["min_sqlite_version_3_7_16"]
+sqlite-memvfs = ["lazy_static", "parking_lot", "log", "anyhow"]
+
+[dependencies]
+lazy_static = { version = "1.4.0", optional = true }
+parking_lot = { version = "0.11.1", optional = true }
+log = { version = "0.4", optional = true }
+anyhow = { version = "1.0.40", optional = true }
 
 [build-dependencies]
 bindgen = { version = "0.58", optional = true, default-features = false, features = ["runtime"] }
--- a/libsqlite3-sys/sqlite3/wasm-ext.c
+++ b/libsqlite3-sys/sqlite3/wasm-ext.c
@@ -1,0 +1,207 @@
+// import implementations from android bionic libc
+
+#include <stdio.h>
+#include <string.h>
+
+size_t strcspn(const char* s1, const char* s2) {
+    const char *p, *spanp;
+    char c, sc;
+    /*
+     * Stop as soon as we find any character from s2.  Note that there
+     * must be a NUL in s2; it suffices to stop when we find that, too.
+     */
+    for (p = s1;;) {
+        c = *p++;
+        spanp = s2;
+        do {
+            if ((sc = *spanp++) == c)
+                return (p - 1 - s1);
+        } while (sc != 0);
+    }
+    /* NOTREACHED */
+}
+
+int strcmp(const char *s1, const char *s2)
+{
+    while (*s1 == *s2++)
+		if (*s1++ == 0)
+			return (0);
+	return (*(unsigned char *)s1 - *(unsigned char *)--s2);
+}
+
+size_t
+strlen(const char *str)
+{
+    const char *s;
+    for (s = str; *s; ++s)
+        ;
+    return (s - str);
+}
+
+int
+strncmp(const char *s1, const char *s2, register size_t n)
+{
+    if (n == 0)
+        return (0);
+    do {
+        if (*s1 != *s2++)
+            return (*(unsigned char *)s1 - *(unsigned char *)--s2);
+        if (*s1++ == 0)
+            break;
+    } while (--n != 0);
+    return (0);
+}
+
+// import from glibc
+char *
+strrchr (const char *s, int c)
+{
+    const char *found, *p;
+    c = (unsigned char) c;
+    /* Since strchr is fast, we use it rather than the obvious loop.  */
+    if (c == '\0')
+        return strchr (s, '\0');
+    found = NULL;
+    while ((p = strchr (s, c)) != NULL)
+    {
+        found = p;
+        s = p + 1;
+    }
+    return (char *) found;
+}
+
+char *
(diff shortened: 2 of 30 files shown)


gwenn:
1. libsqlite3-sys/sqlite3/sqlite3.c cannot be touched (pristine from SQLite3 official almagation)
2. I know nothing about wasm platform so I may ask a dumb question: why headers in libsqlite3-sys/wasm-include are not provided by the compiler/platform ?

libsqlite3-sys/sqlite3/wasm-ext.c
```diff
@@ -0,0 +1,207 @@
+// import implementations from android bionic libc
```
thomcc, line 1: What's the license on these?

libsqlite3-sys/Cargo.toml
```diff
@@ -35,6 +35,13 @@ with-asan = []
 wasm32-wasi-vfs = []
 # lowest version shipped with Windows 10.0.10586 was 3.8.8.3
 winsqlite3 = ["min_sqlite_version_3_7_16"]
+sqlite-memvfs = ["lazy_static", "parking_lot", "log", "anyhow"]
+
+[dependencies]
+lazy_static = { version = "1.4.0", optional = true }
+parking_lot = { version = "0.11.1", optional = true }
+log = { version = "0.4", optional = true }
+anyhow = { version = "1.0.40", optional = true }
```
thomcc, line 44: Why do we need anyhow — doesn't seem to be used?

libsqlite3-sys/src/lib.rs
```diff
@@ -71,3 +71,64 @@ impl Default for sqlite3_vtab_cursor {
         unsafe { mem::zeroed() }
     }
 }
+
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+mod allocator {
```
thomcc, line 76: This allocator seems broken. sqlite_free assumes that the first 8 bytes of the pointer are going to be the size, but sqlite_malloc doesn't allocate space for that, nor store it, etc. I'm surprised this works at all (it seems like if it does, it's a coincidence based on an implementation detail of SQLite allocations).

You need something more along the lines of https://shift.click/blog/on-dealloc/#appendix-how-to-wrap-allocators-which-need-layout-info-in-dealloc, where the information about the allocation is stored before the pointer. Note that that code also handles the full layout, e.g. alignment too, whereas for this we can assume the needed alignment is always the same, and only handle the size.

libsqlite3-sys/src/lib.rs
```diff
@@ -71,3 +71,64 @@ impl Default for sqlite3_vtab_cursor {
...
+    use std::alloc::{alloc, dealloc, realloc as rs_realloc, Layout};
+
+    #[no_mangle]
+    pub unsafe fn sqlite_malloc(len: usize) -> *mut u8 {
```
thomcc, line 80: These must be `extern "C"`.

I feel like we should either be overriding malloc, or should be using the normal config-based mechanism for overriding the allocator globally, or we should have a big comment explaining why we're doing it this way.

thomcc:
Wow, nice. I wonder if this would make us one of the -sys crates that has a story on wasm32-unknown-unknown.

> libsqlite3-sys/sqlite3/sqlite3.c cannot be touched (pristine from SQLite3 official almagation)

Agreed 100%.

> I know nothing about wasm platform so I may ask a dumb question: why headers in libsqlite3-sys/wasm-include are not provided by the compiler/platform ?

No, sadly. It's because there's no libc on the web. This has historically been a huge challenge for C code used by rust libraries on the web. Typically, C code uses the wasm32-unknown-emscripten target, which 

---

Is there anywhere in the SQLite api where structures are passed or returned by value? Unfortunately, clang and rustc disagree about wasm32-unknown-unknown's ABI, and so it would be incompatible. Stuff like this is on the path to solving it, but that path is just starting: https://github.com/rust-lang/rust/pull/83763

---

A lot of the declarations in the header don't have definitions. What's the deal there?

---

We also should at least test building for this target in CI.

thomcc (changes requested):

trevyn:
@passchaos This is very cool, do you intend to continue working on this PR to get it merged? If not, do you mind if others pick up the work?

Also, is the memvfs part critical, or can one just be restricted to the `:memory:` db? What is the advantage of having a separate memvfs, since one can have multiple distinct `:memory:` databases?

libsqlite3-sys/sqlite3/wasm-ext.c
trevyn: @thomcc It appears to be BSD 3-clause, is this acceptable? https://en.wikipedia.org/wiki/Bionic_%28software%29

passchaos:
@trevyn Recently I am not dealing with WASM related things, welcome to continue this work; The reason for using a separate memvfs is that SQLite's ':memory:' is connnection-local, data can't be shared between database connections, my implementation allows different db connections to share data by db name, even if used from different threads (However, it seems that the latest SQLite version solves this problem, without confirming whether it can).

libsqlite3-sys/sqlite3/wasm-ext.c
thomcc: Yes, it would need to include the license somewhere nearby though, but that's not hard to arrange.

trevyn:
@passchaos Thank you!!

For reference, I have a very rough WIP of using a very small subset of the OpenBSD libc with clean licensing here: https://github.com/trevyn/rusqlite/tree/wasm32-unknown-unknown

Will submit a separate PR if I can get it working and address the rest of @thomcc 's reviews.

libsqlite3-sys/src/lib.rs
trevyn: https://sqlite.org/malloc.html#the_default_memory_allocator :

>The default allocator implements memsize() by always allocating 8 extra bytes on each malloc() request and storing the size of the allocation in that 8-byte header.

Issue #941, opened 2021-04-20
Seperate Types submodule into own crate


Progdrasil:
Similarly to [`postgres-types`](https://docs.rs/postgres-types/0.2.1/postgres_types/) which is part of the [`rust-postgres`](https://github.com/sfackler/rust-postgres) crate bundle. This would allow to just have the types as a dependency, namely the `ToSql` Trait. This would for example help in making a query builder crate such as [`sea-query`](https://github.com/SeaQL/sea-query). Adding all of rusqlite would be too much. See https://github.com/SeaQL/sea-query/issues/15.

thomcc:
Hmm, not really sure I follow. In what situation would someone want to depend on rusqlite-types when they don't want to depend on rusqlite itself.

That is, if they do want to depend on rusqlite, then the whole dependency makes sense.

If they don't, then even the rusqlite-types dependency is pointless and should be behind a cargo feature.

Putting things in a separate crate is somewhat limiting, as it means we have to worry about the orphan rule/coherence more.

Progdrasil:
The idea is to not require the entirety of rusqlite with all the batteries included and need to deal with all the other features that are not type specific that it provides. With the old cargo feature resolver this is less of an issue, but with the [new feature resolver](https://github.com/rust-lang/rfcs/pull/2957) this can become quite the head ache and lead to binary bloat.

Not only that but having the types module into its own crate can speed up incremental compilation since it will be its own compilation unit.

While I agree that they will eventually need to depend on the full rusqlite crate at one point in the chain, this reduces the size of dependencies that only really need the `ToSql` trait.

Regarding the orphan rule, I completely understand, however from #942 it seems already really well isolated.

thomcc:
Hmm, my experience with the new feature resolver is that it leads to *less* bloat, so I'm confused by your assertion there.

It may speed up incremental compilation slightly, but total compilation time scales with the size of your dependency graph more than the size of the code in it, so I'm not sure I consider that especially convincing.

Progdrasil:
I would have though having duplicates of the same dependency with different features would lead to more bloat.

thomcc:
I missed this comment.

> I would have though having duplicates of the same dependency with different features would lead to more bloat.

Features are still unified. You won't have two copies of the same crate with different features in your crate graph (unless they're different semver versions, which doesn't really apply to rusqlite because of linking with sqlite).

[...]

PR #942, opened 2021-04-20
[WIP] Refactor types module to subcrate

Progdrasil:
Start of the work on #941 to give an idea to the maintainers if they like it. 
Marking this as a draft for that reason.

This does not compile, it still needs some work on the features side and
transfers of those features from said crate.

There are other issues that has cropped up since the refactoring, namely
of non-exhaustive matches, but those are most likely due to the lack of
feature management.


gwenn:
I may be wrong but it seems that [git-mv](https://git-scm.com/docs/git-mv) has not been used.
Consequently, all history for files in [rusqlite-types/src](https://github.com/Progdrasil/rusqlite/tree/seperate-types-into-subcrate/rusqlite-types/src) is lost !

Progdrasil:
Hm my bad, I tried fixing that but I seem to have ammended the wrong commit. I'll close this.

Issue #913, opened 2021-03-08
Potential use after free in `Statement::column_names()` and similar functions


weiznich:
While investigating the consequences of https://github.com/diesel-rs/diesel/pull/2663 I had a look at the corresponding rusqlite code to gather some information how the underlying problem of the "strange" lifetime of the column name pointer is handled there. While doing that I noticed that at least one case is also not handled here.

> The returned string pointer is valid until either the prepared statement is destroyed by sqlite3_finalize() or until the statement is automatically reprepared by the first call to sqlite3_step() for a particular run **or until the next call to sqlite3_column_name() or sqlite3_column_name16() on the same column.**

[SQLite Documentation for `sqlite3_column_name`](https://www.sqlite.org/c3ref/column_name.html) (emphasis by me)

That sounds for me that calling [this](https://github.com/rusqlite/rusqlite/blob/e68d752cf42e6b0ac40b05eb56ff8b6f065a452d/src/column.rs#L28) function from a user application at least twice and using the first result after the second call will be a potential use after free. The core issue here is that [this](https://github.com/rusqlite/rusqlite/blob/e68d752cf42e6b0ac40b05eb56ff8b6f065a452d/src/raw_statement.rs#L87) function just forwards every call to the corresponding `sqlite3_column_name` function. (This likely applies to any other function in `column.rs` which internally calls that function in `raw_statement.rs`).

gwenn:
I guess you need a mutable reference to `Statement` to reprepared / step, no ?
So you should not be able to keep a reference to column names and step.

And if you call `column_names` twice, they should share the same references.
But to be sure, we will add a unit test...

thomcc:
We should probably just returned an owned copy of the names in either case, but it would be good to know if it actually causes a problem in practice to know how big of a deal this is.

weiznich:
@gwenn @thomcc The sqlite documentation explicitly states that the return pointer is only valid till the next call to `sqlite3_column_name` for the same column. Its hard to show how this could cause problems in real code as this requires triggering something that actually overrides the freed pointer.

thomcc:
Yeah, I definitely agree we should fix this soon.

I'd like to figure out if it's actually a UAF¹ (and if so how likely it is), which makes a difference for how urgent the fix is² and whether or not it's worth backporting to the last stable release of rusqlite (the current release has a decent amount of breakage, although honestly we should be getting the release out soon anyway — it's been long enough, probably). 

[...]

PR #943, opened 2021-04-25
Add more column info getters

nhynes:
This PR adds the following new methods `struct Column`:
* `origin_name` to get the de-aliased column name
* `table_name` to get the originating table name


src/column.rs
```diff
@@ -133,15 +147,27 @@ impl Statement<'_> {
     /// calling this method.
     #[cfg(feature = "column_decltype")]
     pub fn columns(&self) -> Vec<Column> {
+        macro_rules! expect_optional_str {
```
thomcc, line 150: I think this could just be a function no?

thomcc:
Hrm... the column APIs in Rusqlite might need reworking due to #913. At the moment they don't appear to have any issues, but it seems a bit... worrisome...

PR #944, opened 2021-04-25
Add Statement::parameter_name

nhynes:
This PR does what it says on the tin: it adds the inverse method of `parameter_index`. Although this method isn't used by this library (e.g., for parameter binding), it's useful for external consumers who want to massage a heap of parameters into the right shape for a statement.


thomcc (approved):
looks good to me, thanks

PR #946, opened 2021-04-30
Add support for authorizer hook

nhynes:
This PR adds support for an authorizer hook, which is the best way to get information about what the parser is doing. There shouldn't be any surprises here (except for maybe using a new definition of `Action` and not including docs that can be generated by a macro).

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -624,26 +624,26 @@ impl InnerConnection {
             use std::ffi::CStr;
             use std::str;
 
-            let optional_str = |p_str: *const c_char| {
+            let expect_optional_utf8 = |p_str: *const c_char, description: &'static str| {
                 if p_str.is_null() {
-                    None
-                } else {
-                    let c_slice = CStr::from_ptr(p_str).to_bytes();
-                    str::from_utf8(c_slice).ok()
+                    return None;
                 }
-            };
-            let action = AuthAction::from_raw(
-                action_code,
-                optional_str(action_arg1_str),
-                optional_str(action_arg2_str),
-            );
-            let auth_ctx = AuthContext {
-                action,
-                database_name: optional_str(db_str),
-                accessor: optional_str(accessor_str),
+                str::from_utf8(CStr::from_ptr(p_str).to_bytes())
+                    .unwrap_or_else(|_| panic!("received non-utf8 string as {}", description))
+                    .into()
             };
 
             let r = catch_unwind(|| {
+                let action = AuthAction::from_raw(
+                    action_code,
+                    expect_optional_utf8(action_arg1_str, "authorizer param 1"),
+                    expect_optional_utf8(action_arg2_str, "authorizer param 2"),
+                );
+                let auth_ctx = AuthContext {
+                    action,
+                    database_name: expect_optional_utf8(db_str, "database name"),
+                    accessor: expect_optional_utf8(accessor_str, "inner-most trigger or view"),
+                };
                 let boxed_hook: *mut F = p_arg as *mut F;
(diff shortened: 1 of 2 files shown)


thomcc:
This looks fine to me at a glance, but @gwenn knows much more about sqlite parsing than I do.

src/hooks.rs
```diff
@@ -297,6 +605,84 @@ impl InnerConnection {
...
+        {
+            use std::ffi::CStr;
+            use std::str;
+
+            let optional_str = |p_str: *const c_char| {
+                if p_str.is_null() {
+                    None
+                } else {
+                    let c_slice = CStr::from_ptr(p_str).to_bytes();
+                    str::from_utf8(c_slice).ok()
```
gwenn, line 632: If any string parameter is not utf-8, `None` value makes `AuthAction::from_raw` return `Unknown`.

src/hooks.rs
```diff
@@ -297,6 +605,84 @@ impl InnerConnection {
...
+    where
+        F: for<'r> FnMut(AuthContext<'r>) -> Authorization + Send + RefUnwindSafe + 'static,
+    {
+        unsafe extern "C" fn call_boxed_closure<'c, F>(
+            p_arg: *mut c_void,
+            action_code: c_int,
+            action_arg1_str: *const c_char,
+            action_arg2_str: *const c_char,
+            db_str: *const c_char,
+            accessor_str: *const c_char,
```
gwenn, line 619:
```
param1,
param2,
db_name,
trigger_name,
```

src/hooks.rs
nhynes: want it to `expect` instead?

src/hooks.rs
nhynes: I agree with these names, but they don't match the pre-existing style. For instance `db_str` came from `update_hook`, and `tbl_str` (from the same fn) suggests that the parameters are suffixed with their types. If you want the names changed here, I'd also like to go change the rest of them to keep the style consistent. What would you prefer?

src/hooks.rs
```diff
@@ -37,6 +37,304 @@ impl From<i32> for Action {
     }
 }
 
+/// `feature = "hooks"` The context recieved by an authorizer hook.
```
thomcc, line 40: Please link to https://www.sqlite.org/c3ref/set_authorizer.html

src/hooks.rs
```diff
@@ -37,6 +37,304 @@ impl From<i32> for Action {
...
+    // The inner-most trigger or view responsible for the access attempt.
+    // `None` if the access attempt was made by top-level SQL code.
+    pub accessor: Option<&'c str>,
+}
+
+/// `feature = "hooks"` Actions and arguments
+/// found within a statement during preparation.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[non_exhaustive]
+#[allow(missing_docs)] // This is self-documenting.
```
thomcc, line 58: We should link to https://sqlite.org/c3ref/c_alter_table.html (weird that this is the URL...)

src/hooks.rs
```diff
@@ -37,6 +37,304 @@ impl From<i32> for Action {
...
+    Select,
+    Transaction {
+        operation: TransactionOperation,
+    },
+    Update {
+        table_name: &'c str,
+        column_name: &'c str,
+    },
+    Attach {
+        filename: &'c std::path::Path,
```
thomcc, line 150: This should be a `&str`, since getting a `&str` out of a `&Path` is a pain, and the reverse is trivial.

thomcc (approved):

PR #947, opened 2021-05-02
fix several typos

striezel:
Fix typos in comments / code documentation.

### DIFF
--- a/src/params.rs
+++ b/src/params.rs
@@ -198,7 +198,7 @@ impl Params for &[(&str, &dyn ToSql)] {
 
 macro_rules! impl_for_array_ref {
     ($($N:literal)+) => {$(
-        // These are already generic, and theres a shitload of them, so lets
+        // These are already generic, and there's a shedload of them, so lets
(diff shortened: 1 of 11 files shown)


src/params.rs
```diff
@@ -198,7 +198,7 @@ impl Params for &[(&str, &dyn ToSql)] {
 
 macro_rules! impl_for_array_ref {
     ($($N:literal)+) => {$(
-        // These are already generic, and theres a shitload of them, so lets
+        // These are already generic, and there's a shedload of them, so lets
```
thomcc, line 201: *Fine*

striezel, line 201: Yes, shedload is a word in the Cambridge English Dictionary: https://dictionary.cambridge.org/dictionary/english/shedload
And it's nicer than the previously used one, so I see this as an absolute win. :)

thomcc (approved):
Err, yeah... I can't type worth crap. I don't think *all* of these are mine, but a lot of them look familiar 😅. Thanks.

gwenn (approved):

Issue #566, opened 2019-08-26
Consider supporting an older rustc? ("MSRV")


FauxFaux:
As a popular library that is used by other libraries, it would be convenient if `rusqlite` specified that it would work on some compiler for some "semver release"; a specification which moved slower than just "it works on `stable`". If not, writing that down (even if it's just here) would be cool too.

I'm not going to write a big essay on the advantages and disadvantages of this, as I'm not convinced either way myself.

As an example breakage, https://github.com/jgallagher/rusqlite/commit/208f3c084baada626eca4ce6ad4111b09f159251 (Jul 10th), released Jul 27th, broke rustc <1.36, released Jul 4th. Based on the commit message, this wasn't (super) intentional.

`1.34.2` might be a good thing to support initially, as it's currently recent (but not recent enough to build rusqlite), and it's what's in Debian (i.e. what I hit!). I would also be happy with just whatever's `stable` now, though.

gwenn:
@thomcc I agree with you.
Maybe we can edit release notes by checking Travis CI history and adding the stable version used ?

FauxFaux:
Okay, all good!

Perhaps stating that `0.20.x` will continue to support whatever was stable at `0.20.0`? Not that there has ever really been a point release since `0.10.x`.

thomcc:
> Also related, breaking this MSRV should be considered a breaking semver change. Otherwise you can break other crates depending on this crate but having an older Rust version.

For a library, if you care about an MSRV it's only important that there exist a set of packages that satisfy the MSRV under your requirements, rather than have every possible set of packages under your satisfy the MSRV. The distinction means that you can enforce MSRV via a lockfile, at the point -- It's generally impossible to enforce this sort of thing without one. Concretely, can do something like `cargo update -p rusqlite --precise $specific_version_here` to ensure (if possible) that your Cargo.lock contains `$specific_version_here` for rusqlite, for example.

In practice rusqlite is not that aggressive about using new features anyway, and where reasonable we have a history of taking patches to make things work in older Rust.

See also https://github.com/rust-lang/libs-team/issues/72 for more discussions on this topic. MSRV is considerably less simple than most people make it out to be.

nyurik:
[...]

PR #949, opened 2021-05-04
Fix build failure on Debian GNU/Linux 10

striezel:
**Short summary:**
rusqlite 0.25.1 does not build with the default Rust compiler that comes with Debian GNU/Linux 10, but version 0.23.1 did build just fine on that system. This pull request makes some adjustments so that it builds again and Debian users can enjoy the newer version, too.

**Longer explanation:**
[Debian](https://www.debian.org/) is a Linux distribution, and among the top ten most-popular distributions according to [DistroWatch](https://distrowatch.com/table.php?distribution=debian), and a lot of other distributions are based on Debian. Like other Linux distributions it comes with various prepackaged software, and among that software also is the Rust compiler. However, the current stable release of Debian comes with Rust 1.41.1, which is [a bit over a year old](https://blog.rust-lang.org/2020/02/27/Rust-1.41.1.html). Therefore, it does not support some of the newer Rust features from the current stable branch of Rust. The three main "problematic" newer features are:

* the [`matches!` macro](https://doc.rust-lang.org/std/macro.matches.html), introduced in Rust 1.42.0, and
* pointer's offset_from() that uses a library feature that was still unstable until Rust 1.47 was released, and
* `MIN` and `MAX` constants for types like `u32` and `i64`, those have been added in Rust 1.43.0.

The pull request replaces all of them with compatible versions that also work with Rust 1.41. I tried to use the corresponding code snippets that were used in those places in earlier rusqlite versions as replacements where possible, so that the functionality should not be impacted. The tests still pass.

Furthermore, I took the liberty to add a Minimum Supported Rust Version (MSRV) to the configuration file for Clippy, `clippy.toml`, and set it to 1.41.0. That way Clippy does not warn about using the `matches!` macro anymore. I guess that was the reason why someone replaced the match expressions with the macro in the first place. On top of that a new workflow is added to GitHub actions, using that older Rust version, to ensure it builds and tests pass.

_I am aware that these changes may be a bit intrusive,_ so if you decide not to merge these changes, I'm fine with that, too. Because what this pull request also does as a side effect is setting the MSRV of rusqlite to 1.41.0. I was unable to find an explicit statement what rusqlite's current MSRV is, so I am not sure whether this is OK or not.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -151,3 +151,26 @@ jobs:
 
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v1
+
+  # Checks that the library builds properly when using the MSRV as compiler.
+  msrv:
+    name: Minimum Supported Rust Version
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: 1.41
+        profile: minimal
+        override: true
+    - name: Rust and Cargo version
+      run: |
+        rustc --version --verbose
+        cargo --version --verbose
+    - name: Basic build
+      run: |
+        cargo build --features bundled
+    - name: Basic tests
+      run: |
+        cargo test --features bundled
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,2 @@
 doc-valid-idents = ["SQLite", "lang_transaction"]
+msrv = "1.41.0"
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -377,7 +377,10 @@ mod bindings {
     fn generating_bundled_bindings() -> bool {
         // Hacky way to know if we're generating the bundled bindings
         println!("cargo:rerun-if-env-changed=LIBSQLITE3_SYS_BUNDLING");
-        matches!(std::env::var("LIBSQLITE3_SYS_BUNDLING"), Ok(v) if v != "0")
+        match std::env::var("LIBSQLITE3_SYS_BUNDLING") {
+            Ok(v) if v != "0" => true,
+            _ => false,
+        }
     }
 
     pub fn write_to_out_dir(header: HeaderLocation, out_path: &Path) {
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -261,7 +261,8 @@ impl InnerConnection {
         let tail = if c_tail.is_null() {
             0
         } else {
-            let n = unsafe { c_tail.offset_from(c_sql) };
+            // note: feature ptr_offset_from (#41079) only available in Rust 1.47 or later
+            let n = (c_tail as isize) - (c_sql as isize);
             if n <= 0 || n >= len as isize {
                 0
             } else {
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -136,7 +136,10 @@ impl FromSql for f64 {
 impl FromSql for bool {
     #[inline]
     fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
-        i64::column_result(value).map(|i| !matches!(i, 0))
+        i64::column_result(value).map(|i| match i {
+            0 => false,
+            _ => true,
+        })
     }
 }
 
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -428,34 +428,40 @@ mod test {
         // Basic non-converting test.
         test_conversion!(db_etc, 0u8, u8, expect 0u8);
 
+        // Constants like u32::MAX are only available in Rust 1.43+.
+        const U32_MAX: u32 = 4294967295;
+        const I64_MIN: i64 = -9223372036854775808;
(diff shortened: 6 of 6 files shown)


gwenn:
See https://github.com/rusqlite/rusqlite/issues/566

striezel:
So if I understand issue 566 correctly, then there is no intention to explicitly decide on any MSRV other than latest stable? That's sad, because it basically means one cannot depend on rusqlite outside of development, i. e. on production systems. :crying_cat_face:

Considering that rusqlite is also listed as the crate for using [SQLite dabases in the Rust Cookbook](https://rust-lang-nursery.github.io/rust-cookbook/database/sqlite.html), one can expect that more people will hit this very issue when trying to work with Rust and SQLite. (Doesn't need to be with Debian, could also be a derived distro.) And trying a library that does not compile, because one needs the latest version of a compiler is not exactly a uplifting experience. In fact, it is one of the quickest way to turn people away from that library.

But let's get back to the single argument that has been made against any MSRV in #566: _"supporting older versions is additional work"_.
Well, it is, but in this case the work has already been done in this pull request, and it has been shown to work, as can be seen in the newly added check. _The necessary work has already been done by someone else so you do not need to do it._ That's the magic of open source collaboration. :) So the only amount of work that is left to do at the moment is to merge these changes.

thomcc:
> That's sad, because it basically means one cannot depend on rusqlite outside of development, i. e. on production systems

You're still free to use an older version of rusqlite. If you aren't using the latest version of Rust, it seems reasonable that you won't use the latest version of every Rust library.

Also note that rusqlite *is* used in a number of production systems. I don't think that evaluation is particularly fair.

---

Anyway, my feelings here are: I don't really mind taking simple changes like these to allow compiling on  an older Rust, but I don't think I'm willing add a promise that rusqlite releases have, for example, a MSRV no newer than debian stable.

To a large extent, that can turn out as quite a quagmire if a dependency breaks it (for example), and Rust 1.41 is well over a year old at this point...

thomcc:
More generally: sing Rust in production and development tends to involve using rustup, rather than using the version of Rust that comes with your Linux distro.

The way I see it — the version of Rust in the Debian package repo (or, in any distro's repo) is basically just for use with the rust packages in that repo. Once you're pulling from crates.io, if you're pulling the latest versions of libraries, you should not expect the latest version of most packages to work with that version of Rust.

Regardless of your feelings on this, this is basically the policy that the Rust ecosystem has settled on at the moment. New users will hit similar issues pretty quickly with the vast majority of the crates in the ecosystem as well...

Almost nothing supports a Rust that old unless it has a reason to or hasn't been updated recently, and so I'm not terribly worried about a bad experience there (although I do wish the error messages in this case were better).

---

So yeah, while my feelings on MSRV have mellowed a bit since I said that stuff, I'm not really comfortable with committing to one so old at the very least...

That's not to say there'd be no benefit for rusqlite either having an MSRV, or even working on the Rust shipped in Debian Stable. In particular, it seems reasonable to want to allow programs using rusqlite to be packaged in Debian, or at least not be a particular thorn in the side of that process... But at the moment, new packages aren't being added to Debian 10, so I doubt that matters.

thomcc:
All that said, and with the caveat that this doesn't constitute rusqlite committing to a MSRV:

If you need them, I'd don't really see a problem with accepting the code changes (e.g. without the CI/clippy.toml changes), and putting them in a 0.25.2 patch release, since they are small, unobtrusive, don't impact any public APIs, and the work has already been done.

To be clear, even if we do that, you should expect that rusqlite 0.26 will very likely need a newer Rust — Possibly even requiring 1.51, since I think there are APIs that can be improved using const generics.

striezel:
> Almost nothing supports a Rust that old unless it has a reason to or hasn't been updated recently, and so I'm not terribly worried about a bad experience for new users there (although I do wish the error messages in this case were better).

While this evaluation certainly holds for some Rust crates out there, that is not the case for some of the most popular crates. Those tend to support older Rust versions than the current stable. Just to name a few: [regex](https://crates.io/crates/regex) supports rustc 1.41.1+ (updated 3 days ago), [serde](https://crates.io/crates/serde) even supports rustc 1.36+ (updated a month ago), [syn](https://crates.io/crates/syn) even supports rustc 1.31+ (updated one day ago), and both of the latter ones are among the top ten most downloaded crates on crates.io. In fact, all of the top ten most downloaded crates on crates.io support rustc 1.41.1+ or even some older versions. So (in my experience) while there are some obstacles when using older rustc, most of the relevant crates still work with the rustc in Debian 10, even when using the newest version of those crates.

> If you need this, I don't really see a problem with accepting the code changes (e.g. without the CI/clippy.toml/etc changes), and putting them in a 0.25.2 patch release, since they are small, unobtrusive, don't impact any public APIs, and the work has already been done.

Very well, that is good enough for me, :+1: even without the commitment to any MSRV. I'll send you another PR without the CI and clippy.toml changes.

thomcc:
> In fact, all of the top ten most downloaded crates on crates.io support rustc 1.41.1+ or even some older versions

Sure, but the vast majority of the crates on crates.io don't, even among crates people use, it's fairly rare.

> I'll send you another PR without the CI and clippy.toml changes.

Superseded by #952

PR #951, opened 2021-05-04
Clarify when Error::InvalidColumnType is returned

### LINKED ISSUE #950: Documentation for column type errors mentions column type rather than value type
`InvalidColumnType` is described as an error when the type requested doesn't match the column type, but in my testing it seems to only depend on the SQLite value (not the column type). For example, getting a string column as an Option<i32> is valid as long as the data is NULL. The way the docs are currently worded, a nullable column should be gotten as `Option<T>` and a non-nullable column should be gotten as `T`.

I assume "column type" was meant to refer to the column of a row (the value type rather than the type in the table schema). But that interpretation still doesn't match current behavior.

I'll make a PR to reword it.

lefth:
It is due to mismatch in the SQLite value in the row, not the column type.

Fixes #950.

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -245,7 +245,7 @@ impl<'stmt> Row<'stmt> {
     /// Panics if calling [`row.get(idx)`](Row::get) would return an error,
     /// including:
     ///
-    /// * If the underlying SQLite column type is not a valid type as a source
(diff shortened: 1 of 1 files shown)


gwenn:
See https://github.com/rusqlite/rusqlite/issues/950#issuecomment-832824604

PR #952, opened 2021-05-05
Fix build failure on Debian GNU/Linux 10 (without changes of CI and clippy.toml)

striezel:
@thomcc:

> f you need this, I don't really see a problem with accepting the code changes (e.g. without the CI/clippy.toml/etc changes), and putting them in a 0.25.2 patch release, since they are small, unobtrusive, don't impact any public APIs, and the work has already been done.

As mentioned in #949, here's the new PR. It is essentially the same as #949, but without setting the MSRV in clippy.toml and without adding a new CI job. I'd be happy if these changes could be merged and released as a new patch release.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -377,7 +377,10 @@ mod bindings {
     fn generating_bundled_bindings() -> bool {
         // Hacky way to know if we're generating the bundled bindings
         println!("cargo:rerun-if-env-changed=LIBSQLITE3_SYS_BUNDLING");
-        matches!(std::env::var("LIBSQLITE3_SYS_BUNDLING"), Ok(v) if v != "0")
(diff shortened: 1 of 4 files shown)


striezel:
As expected, the clippy CI job does now warn about not using the `matches!` macro and fails for that reason. You might want to [cherry pick](https://git-scm.com/docs/git-cherry-pick) commit 2b694e8 from PR #949 to fix that. I left it out intentionally, as requested in #949.

thomcc:
> As expected, the clippy CI job does now warn about not using the matches! macro and fails for that reason

It's fine just to allow the lint, I think.

gwenn:
You intend to merge these changes in a dedicated branch (and not in trunk), right ?
Because I guess there will be no PR that rollbacks theses changes when rustc is upgraded in Debian.

thomcc:
Lets take #954 instead.

PR #953, opened 2021-05-05
update year in license file

striezel:


gwenn:
[We](https://github.com/rust-lang/rust/blob/master/LICENSE-MIT) don't care ?

striezel:
> We don't care ?

Well, that is a question I should not decide. (I guess I could, but then who cares what a random guy is saying on the internet?)

The [MIT template](https://choosealicense.com/licenses/mit/) usually includes a line with the year(s), and that is the case within rusqlite at the moment. Therefore I assumed that an update to the current year would be in order. And if you are pointing to rust-lang/rust repo: Even they did [update years in copyright notices](https://github.com/rust-lang/rust/commit/f979f91ae20b2da9b2414033460985d5e023fbc5#diff-421f8eecf75a1acdd7907667d6c14e9f8870cb90a7cc499459c8418a93cdc8b0) before the line was removed from the license.

So there are (at least) two possibilities on how to proceed:

* 0: Decide that you don't care and remove the line altogether.
* 1: Decide that you do care and update it.
* 2: Decide that you want to keep the line but do not care to keep it up to date.

Choose one.

thomcc:
It [doesn't really matter](https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year), but I don't see a reason not to accept it.

PR #954, opened 2021-05-05
Fix build on some older versions of Rust

thomcc:
This is a less obtrusive version of #952, that also handles non-default features. It doesn't commit to any MSRV, but works back to 1.41 since some users care about that, and it's only a couple tiny fixes.

This doesn't include changes to make the tests pass under older rust, since that's not needed for normal usage.

### DIFF
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -261,7 +261,7 @@ impl InnerConnection {
         let tail = if c_tail.is_null() {
             0
         } else {
-            let n = unsafe { c_tail.offset_from(c_sql) };
+            let n = (c_tail as isize) - (c_sql as isize);
             if n <= 0 || n >= len as isize {
                 0
             } else {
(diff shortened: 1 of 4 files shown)


src/inner_connection.rs
```diff
@@ -261,7 +261,7 @@ impl InnerConnection {
         let tail = if c_tail.is_null() {
             0
         } else {
-            let n = unsafe { c_tail.offset_from(c_sql) };
+            let n = (c_tail as isize) - (c_sql as isize);
```
thomcc, line 264: IMO this is actually better, since it doesn't require unsafe.

striezel:
Note that this commit is missing the [change in `src/types/mod.rs`](https://github.com/rusqlite/rusqlite/pull/952/commits/a3bb1610f7674cea597c31aba96b9a2cb1c1c21c#diff-44bee847315d06d50bad561b0284381c81b52274429ff45ccec047d25f8ce45d) from #952, so it won't build with rustc 1.41.1 yet.

thomcc:
@striezel Yeah that's intentional. There's no reason to think if the tests pass on recent rust they wouldn't on old rust, and I'm not willing to live with those changes, and don't care enough about this to backport it to non-master.

gwenn (approved):

PR #958, opened 2021-05-09
`bundled-full` should not be a default feature

thomcc:
Undo adding `bundled-full` to default features as that's a huge accidental breaking change just made in `0.25.2`.

We can talk about re-adding it later, I think probably not, but I could possibly be convinced.


gwenn:
Thanks.

PR #961, opened 2021-05-24
Use `env::var` instead of `cfg` in some cases in build.rs

daladim:
This MR is properly using `std::env::var("CARGO_CFG_whatever")` instead of `cfg!(whatever)`

This is required to handle cross-compilation, see https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts \
Note that this requires to pull `vcpkg` as a dev-dependency whatever cargo features are used. But this does not cuase any trouble on native Windows, native Linux or cross Linux-to-Windows.

Without these changes, using msvc-wine to build rusqlite would try to compile libsqlite with `-DHAVE_LOCALTIME_R` (which fails).

(Note also that there is an undocumented feature (`bundled-windows`), that you may want to either document or drop completely)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - fix_cross_compilation
   pull_request:
     branches:
       - master
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - fix_cross_compilation
   pull_request:
     branches:
       - master
--- a/libsqlite3-sys/Cargo.toml
+++ b/libsqlite3-sys/Cargo.toml
@@ -13,8 +13,8 @@ categories = ["external-ffi-bindings"]
 
 [features]
 default = ["min_sqlite_version_3_6_8"]
-bundled = ["cc", "bundled_bindings"]
-bundled-windows = ["cc", "bundled_bindings"]
+bundled = ["bundled_bindings"]
+bundled-windows = ["bundled_bindings"]
 buildtime_bindgen = ["bindgen", "pkg-config", "vcpkg"]
 sqlcipher = []
 min_sqlite_version_3_6_8 = ["pkg-config", "vcpkg"]
@@ -39,7 +39,5 @@ winsqlite3 = ["min_sqlite_version_3_7_16"]
 [build-dependencies]
 bindgen = { version = "0.58", optional = true, default-features = false, features = ["runtime"] }
 pkg-config = { version = "0.3.19", optional = true }
-cc = { version = "1.0", optional = true }
-
-[target.'cfg(target_env = "msvc")'.build-dependencies]
+cc = { version = "1.0" }
 vcpkg = { version = "0.2", optional = true }
--- a/libsqlite3-sys/Cargo.toml
+++ b/libsqlite3-sys/Cargo.toml
@@ -13,8 +13,8 @@ categories = ["external-ffi-bindings"]
 
 [features]
 default = ["min_sqlite_version_3_6_8"]
-bundled = ["bundled_bindings"]
-bundled-windows = ["bundled_bindings"]
+bundled = ["cc", "bundled_bindings"]
+bundled-windows = ["cc", "bundled_bindings"]
 buildtime_bindgen = ["bindgen", "pkg-config", "vcpkg"]
 sqlcipher = []
 min_sqlite_version_3_6_8 = ["pkg-config", "vcpkg"]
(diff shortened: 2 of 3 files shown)


daladim:
OK it actually doesn't build... I'll have a look at it

thomcc:
I think you should probably stick to moving the cfg!(windows) to the runtime check. Moving everything will be a problem, as you're seeing.

That said, part of the reason I never bothered with this is cargo seems to set this properly in my experience. I did cross-compilation builds of rusqlite many times for https://github.com/mozilla/application-services, and its even part of one of the CI tasks there.

daladim:
This version should now build.
I have also promoted `cc` to a mandatory dev-dependency, I hope that's something you're fine with (reading your comment, it looks like you preferred avoiding pulling `cc` as a dependency, but I'm not sure there is an alternative).

I'm interested about your last message. You're talking about CI tests that try to cross-compile rusqlite in `application-services`, but I couldn't find them there :-( Could you tell me where they are?
I'd like to see whether this current PR is mandatory, or if there are some other ways to tell cargo to properly cross-compile. (Yet, looking at the cargo book, it looks like we won't be able to properly cross-compile unless we get rid of these `cfg!` macros).

.github/workflows/main.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - fix_cross_compilation
```
thomcc, line 7: This isn't needed. There's no way to get your CI to run without our manual approval, annoyingly.

thomcc:
In here somewhere, https://github.com/mozilla/application-services/blob/main/taskcluster/ci/module-build/kind.yml. Note that it goes through https://github.com/mozilla/rust-android-gradle/.

I guess https://github.com/mozilla/application-services/issues/3917 indicates that this no longer works in the CI, but that seems unrelated to the use of these macros in rusqlite, which have been there for quite a while.

That said, I'm not strictly speaking opposed to these changes, as long as they're minimally disruptive. I don't know that I think requiring vcpkg and cc everywhere is minimally disruptive, though.

thomcc:
Anyway, I think the `cfg!(windows)` makes sense that it wouldn't work (my experience is that it *does*, although perhaps that branch of the build script never ran). So if you fix the clippy/rustfmt complaints I'll land this.

daladim:
I've pushed a few review commits.

Thanks for the link, I would not have found it myself!

daladim:
Note that running a cross compilation (from Linux to Windows) displays a warning, but does not fail the build:
```
~/src/rusqlite $ cargo build --target x86_64-pc-windows-msvc --features bundled
warning: sqlite3/sqlite3.c(22525,8): warning: implicit declaration of function 'localtime_r' is invalid in C99 [-Wimplicit-function-declaration]
warning:   rc = localtime_r(t, pTm)==0;
warning:        ^
warning: sqlite3/sqlite3.c(27292,3): warning: '_ReadWriteBarrier' is deprecated: use other intrinsics or C++11 atomics instead [-Wdeprecated-declarations]
warning:   _ReadWriteBarrier();
warning:   ^
warning: /usr/lib/llvm-10/lib/clang/10.0.0/include/intrin.h(176,16): note: '_ReadWriteBarrier' has been explicitly marked deprecated here
warning: __attribute__((__deprecated__("use other intrinsics or C++11 atomics instead")))
warning:                ^
warning: 2 warnings generated.
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
```

But this turns into a hard error when I try to use rusqlite as a dependency from another rust crate (e.g., in my case where I'm using lld: `lld-link: error: undefined symbol: localtime_r`)

thomcc (approved):
lgtm

thomcc:
Any hesitation @gwenn?

libsqlite3-sys/Cargo.toml
```diff
@@ -39,7 +39,5 @@ winsqlite3 = ["min_sqlite_version_3_7_16"]
 [build-dependencies]
 bindgen = { version = "0.58", optional = true, default-features = false, features = ["runtime"] }
 pkg-config = { version = "0.3.19", optional = true }
-cc = { version = "1.0", optional = true }
-
-[target.'cfg(target_env = "msvc")'.build-dependencies]
+cc = { version = "1.0" }
```
gwenn, line 42: We should not depend on `cc` if no 'bundled*' feature is enabled, no ?

daladim: You're right, I could actually avoid it. I've pushed a fix

thomcc:
Hm, I'm not sure why CI didn't trigger here...

PR #969, opened 2021-06-11
Add time back to modern-full feature

gwenn:
See https://github.com/time-rs/time/issues/292


thomcc:
Can you remove it from the places its explicitly added in https://github.com/rusqlite/rusqlite/blob/master/.github/workflows/main.yml ?

gwenn:
> Can you remove it from the places its explicitly added in https://github.com/rusqlite/rusqlite/blob/master/.github/workflows/main.yml ?

Indeed !

thomcc (approved):
Assuming everything is green (and I don't see why it wouldn't be) :shipit:!

Issue #968, opened 2021-06-11
Missing datetime format support 


epilys:
According to https://sqlite.org/lang_datefunc.html the following string formats are expected by sqlite time functions:

1.  YYYY-MM-DD
2.  YYYY-MM-DD HH:MM
3.   YYYY-MM-DD HH:MM:SS
4.   YYYY-MM-DD HH:MM:SS.SSS
5.   YYYY-MM-DDTHH:MM
6.   YYYY-MM-DDTHH:MM:SS
7.   YYYY-MM-DDTHH:MM:SS.SSS
8.   HH:MM
9.    HH:MM:SS
10.   HH:MM:SS.SSS
11.    now
12.   DDDDDDDDDD 

rusqlite doesn't cover all of them, for example the fractional second part. That's `%Y-%m-%d %H:%M:%S %f` but `%f` is not supported by the `time` crate. It only has `%N` which is 9 digits. You can test if the string is 23 bytes long, and pad it with zeros:

```rust
if s.len() == 23 {
  let mut s = format!("{}000000", s);
}
```

Additionally, 

> Formats 2 through 10 may be optionally followed by a timezone indicator of the form "[+-]HH:MM" or just "Z". The date and time functions use UTC or "zulu" time internally, and so the "Z" suffix is a no-op

If the `Z` is missing, the string also doesn't parse correctly.

gwenn:
Could you please be more specific ?
For example, you cannot reasonably transform an `HH:MM` into a `OffsetDateTime` !
And I don't expect `now` to be stored in an SQLite database.

epilys:
Of course, as I said in the original post, a string like `2021-06-06 15:01:11.697` is valid for sqlite3 but cannot be parsed with rusqlite since the format strings here https://github.com/rusqlite/rusqlite/blob/840a6f3789f91d71599fa8fe909cc0d51b13f862/src/types/time.rs#L6-L8
don't recognize it.

Here are the formatting specifiers for the `time` crate: https://docs.rs/time/0.2.26/time/#formatting

The only subsecond specifier is

> %N | Subsecond nanoseconds. Always 9 digits

[...]

PR #970, opened 2021-06-11
Fix FromSql impl for OffsetDateTime

gwenn:
See 'Missing datetime format support' #968

- [X] wait until version 0.3 of `time` is released
- [ ] support more input formats ?


epilys:
Please see #971, I think it's more succinct and complete (?).

src/types/time.rs
```diff
@@ -1,30 +1,64 @@
...
+            .map_err(|err| Error::ToSqlConversionFailure(err.into()))?;
         Ok(ToSqlOutput::from(time_string))
     }
 }
 
 impl FromSql for OffsetDateTime {
     fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
         value.as_str().and_then(|s| {
+            let s = s.strip_suffix('Z').unwrap_or(s);
             match s.len() {
```
thomcc, line 41: I think this would be clearer as if/else now, but I also don't really care that much, so if you don't feel like changing it, we can keep it.

thomcc (approved):

PR #973, opened 2021-06-13
Use #[doc(cfg)]

### LINKED ISSUE #835: Use #[doc(cfg)] instead of manually indicating `feature = "blah"` in the comments
Apparently there's a way to do this. Not exactly sure on the details, though... It's unstable, so maybe we'd need to work around that somehow too...

gwenn:
Fix #835


src/load_extension_guard.rs
```diff
@@ -14,6 +14,7 @@ use crate::{Connection, Result};
 ///     conn.load_extension(Path::new("my_sqlite_extension"), None)
 /// }
 /// ```
+#[cfg_attr(docsrs, doc(cfg(feature = "load_extension")))]
```
gwenn, line 17: Works only when put here (not in lib.rs)...

src/hooks.rs
```diff
@@ -9,11 +9,12 @@ use crate::ffi;
 
 use crate::{Connection, InnerConnection};
 
-/// `feature = "hooks"` Action Codes
+/// Action Codes
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[repr(i32)]
 #[non_exhaustive]
 #[allow(clippy::upper_case_acronyms)]
+#[cfg_attr(docsrs, doc(cfg(feature = "hooks")))]
```
gwenn, line 17: Works only when put here (not in lib.rs)...

src/types/to_sql.rs
```diff
@@ -70,9 +72,11 @@ from_value!(Vec<u8>);
 // `i128` needs in `Into<Value>`, but it's probably fine for the moment, and not
 // worth adding another case to Value.
 #[cfg(feature = "i128_blob")]
+#[cfg_attr(docsrs, doc(cfg(feature = "i128_blob")))]
```
gwenn, line 75: Doesn't work (due to macro ?)

src/types/to_sql.rs
```diff
@@ -70,9 +72,11 @@ from_value!(Vec<u8>);
...
 from_value!(i128);
 
 #[cfg(feature = "uuid")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid")))]
```
gwenn, line 79: Doesn't work (due to macro ?)

thomcc (approved):
Sorry for the delay. This seems fine.

PR #974, opened 2021-06-13
Fix InnerConnection decode_result / changes

### LINKED ISSUE #931: Why do some non-mutating methods on `InnerConnection` take `&self` and others `&mut self`?
Hi there. I am trying to create lower-level, application-specific bindings to SQLite myself, and for that I'm studying the source of Rusqlite.

I am aware that it is highly non-trivial to wrap the SQLite C library in a way that respects thread safety and Rust's lifetime and aliasing rules. For this reason, I'm trying to understand when exactly one needs mutable exclusive or immutable shared access to the `sqlite3 *` connection handle.

The relevant type in Rusqlite is `InnerConnection`. Unfortunately, I do not understand why some seemingly read-only methods take `&mut self` while others simply take `&self`. The SQLite C API itself is not const-correct, so `const` arguments don't help answer this question.

As a concrete example, let's take:

```
pub fn last_insert_rowid(&self) -> i64 { … }
pub fn changes(&mut self) -> usize { … }
pub fn decode_result(&mut self, code: c_int) -> Result<()> { … }
```

Compare this to:
* `last_insert_rowid` takes `&self` and does not mutate anything, as `sqlite3_last_insert_rowid` just returns a member of the DB handle directly. So far so good.
* `changes` takes `&mut self`. However, it does not mutate anything, as `sqlite3_changes` simply returns a member of the `sqlite3` structure.
* `decode_result` takes a `&mut self` as well. It calls `sqlite3_errstr()` internally, which can indeed mutate the DB handle. It can change the text encoding of the backing `sqlite3_value` in-place. However, this function is protected by a mutex. So it looks like it might as well take `&self`.

All in all, what am I missing?

gwenn:
Fix #931

### DIFF
--- a/src/blob/mod.rs
+++ b/src/blob/mod.rs
@@ -223,7 +223,7 @@ impl Connection {
         row_id: i64,
         read_only: bool,
     ) -> Result<Blob<'a>> {
-        let mut c = self.db.borrow_mut();
+        let c = self.db.borrow_mut();
         let mut blob = ptr::null_mut();
         let db = db.as_cstring()?;
         let table = super::str_to_cstring(table)?;
--- a/src/busy.rs
+++ b/src/busy.rs
@@ -64,7 +64,7 @@ impl Connection {
                 0
             }
         }
-        let mut c = self.db.borrow_mut();
+        let c = self.db.borrow_mut();
         let r = match callback {
             Some(f) => unsafe {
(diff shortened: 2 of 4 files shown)


thomcc (approved):
Hmm, I'm not 100% sure I think this matters at all, but ok.

PR #975, opened 2021-06-13
Fix AuthContext / Authorization visibility

### LINKED ISSUE #972: authorizer is public but AuthContext / Authorization are not
https://github.com/rusqlite/rusqlite/blob/master/src/hooks.rs#L403
```rust
    pub fn authorizer&lt;'c, F&gt;(&amp;self, hook: Option&lt;F&gt;)    where        F: for&lt;'r&gt; FnMut(AuthContext&lt;'r&gt;) -&gt; Authorization + Send + RefUnwindSafe + 'static,<br class="Apple-interchange-newline">
```

gwenn:
Fix #972
Breaking change: Action is not visible from root anymore

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -42,14 +42,14 @@ impl From<i32> for Action {
 /// See <https://sqlite.org/c3ref/set_authorizer.html> for more info.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct AuthContext<'c> {
-    // The action to be authorized.
+    /// The action to be authorized.
     pub action: AuthAction<'c>,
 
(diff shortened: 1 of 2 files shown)


thomcc (approved):
Is there anything else in hooks we don't want to expose? (As that module is older, and this might expose other stuff from it).

If not this is fine.

gwenn:
> Is there anything else in hooks we don't want to expose? (As that module is older, and this might expose other stuff from it).
> 
> If not this is fine.

Except https://github.com/rusqlite/rusqlite/blob/master/src/hooks.rs#L600 (minor), all seems good to me.

PR #982, opened 2021-07-02
Include the session FFI in the bundled bindings

thomcc:
This has essentially no downside (it will not cause link errors unless you actually call these functions), but allows the session feature to work properly without requiring buildtime_bindgen (which doesn't generally work on windows).

It also fixes a number of errors in my editor that I see whenever I work on this project.

That said, @gwenn, you've been resistant to this in the past, but I don't understand why. I'd really like to land this (the errors are annoying, and I don't know of a downside to it), so if you object to it I'd like to understand why.

### DIFF
--- a/libsqlite3-sys/upgrade.sh
+++ b/libsqlite3-sys/upgrade.sh
@@ -24,7 +24,7 @@ rm -f "$SQLITE3_LIB_DIR/bindgen_bundled_version.rs"
 cargo update
 # Just to make sure there is only one bindgen.rs file in target dir
 find "$SCRIPT_DIR/../target" -type f -name bindgen.rs -exec rm {} \;
-env LIBSQLITE3_SYS_BUNDLING=1 cargo build --features "buildtime_bindgen" --no-default-features
+env LIBSQLITE3_SYS_BUNDLING=1 cargo build --features "buildtime_bindgen session" --no-default-features
 find "$SCRIPT_DIR/../target" -type f -name bindgen.rs -exec mv {} "$SQLITE3_LIB_DIR/bindgen_bundled_version.rs" \;
 
 SQLCIPHER_VERSION="4.4.3"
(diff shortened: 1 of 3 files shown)


libsqlite3-sys/upgrade.sh
```diff
@@ -41,9 +41,10 @@ rm -rf "v${SQLCIPHER_VERSION}.tar.gz" sqlcipher.src
 
 # Regenerate bindgen file for sqlcipher
 rm -f "$SQLCIPHER_LIB_DIR/bindgen_bundled_version.rs"
+cargo clean
```
thomcc, line 44: This missing was a bug, and would cause some us to find the wrong bundled version some of the time.

gwenn: `find "$SCRIPT_DIR/../target" -type f -name bindgen.rs -exec rm {} \;` doesn't work ?

gwenn:
> allows the session feature to work properly without requiring buildtime_bindgen

Should we update `Cargo.toml` accordingly ?

And I am fine with your PR because you don't touch [default compilation options](https://github.com/rusqlite/rusqlite/blob/master/libsqlite3-sys/build.rs#L110-L129) directly.
You just activate [these](https://github.com/rusqlite/rusqlite/blob/master/libsqlite3-sys/build.rs#L258-L263) lines.

thomcc:
> Should we update Cargo.toml accordingly ?

Session didn't actually depend on buildtime_bindgen, if that's what you mean.

Unless you mean adding it to modern-full/bundled-full, which I don't care about really. I guess it would make stuff easier to test manually.

libsqlite3-sys/upgrade.sh
gwenn: I guess `cargo clean` at line 44 should be removed. And `# find "$SCRIPT_DIR/../target" -type f -name bindgen.rs -exec rm {} \;` at line 46 should be uncommented.

thomcc: oh, that makes sense.

PR #996, opened 2021-07-22
[WIP] VM must run to completion for constraint violations to become visible

### LINKED ISSUE #995: VM must run to completion for constraint violations to become visible
This issue is about a strange interaction between `RETURNING`, rusqlite, and the way SQLite raises errors about constraint violations.

* constraints are only violated during inserts, updates, and deletes
* before `RETURNING` existed, inserts and updates would only go through `execute` and not trigger this issue
* when `RETURNING`, constraint violations are only thrown by SQLite after the VM is fully stepped
* rusqlite methods like `query_row` lead to violations being silently reverted

I'm not sure if this is actually a bug, but it might be worth noting somewhere (e.g., in this issue).

## MWE

```toml
# Cargo.toml
rusqlite = { version = "*", features = ["bundled"] }
```

```rust
// main.rs
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let db = rusqlite::Connection::open_in_memory()?;

    db.execute_batch(
        "
        CREATE TABLE items (id INTEGER PRIMARY KEY);
        CREATE TABLE trades (item INTEGER REFERENCES items(id) NOT NULL);
        ",
    )?;

    let mut stmt = db.prepare_cached("INSERT INTO trades (item) VALUES (42) RETURNING *")?;

    // Works.
    {
        let mut rows = stmt.query([])?;
        assert!(rows.next().is_ok());
        assert!(rows.next().is_err());
    }

    // Doesn't work because the VM hasn't stepped to completion.
    {
        let item_id: u8 = stmt.query_row([], |row| row.get("item"))?;
        assert_eq!(item_id, 42, "huh");
    }
    // `sqlite3_reset` via `<Rows as Drop>::drop` doesn't help either.

    // However, SQLite is still enforcing the constraint.
    assert_eq!(
        db.query_row("SELECT * FROM trades", [], |_| Ok(())),
        Err(rusqlite::Error::QueryReturnedNoRows)
    );

    Ok(())
}
```

gwenn:
Should fix #995 

- [X] Check SQLite version and disable test accordingly
- [ ] Check with multiple VALUES
- [ ] Check if [onecolumn](https://sqlite.org/tclsqlite.html#onecolumn) of TCL interface is also impacted

If multiple values are inserted (`INSERT INTO ... VALUES (...), (...) RETURNING ...;`  or `INSERT INTO ... SELECT ... RETURNING ...;`), this PR doesn't work.
Either we make `query_row` fails if more than one row is retrieved or we must loop until done (or failure) !


thomcc:
Probably worth just putting the test behind `#[cfg(feature = "modern_sqlite")]` or something. I don't *think* those others would have it set.

thomcc:
> Either we make query_row fails if more than one row is retrieved or we must loop until done (or failure) !

Both of these seem really, really unfortunate. The first would break code in the wild, and the 2nd would seriously regress its performance (including queries I know exist in real world programs — specifically firefox).

I wonder if we should consider reporting this as a bug upstream to SQLite?

gwenn:
TCL interface is also impacted:
```sh
% load /usr/lib/sqlite3.36.0/libtclsqlite3.so
% sqlite3 db1
% db1 eval {CREATE TABLE items (id INTEGER PRIMARY KEY);}
% db1 eval {CREATE TABLE trades (item INTEGER REFERENCES items(id) NOT NULL);}
% db1 onecolumn {INSERT INTO trades (item) VALUES (42) RETURNING *}
42
```

gwenn:
@nhynes, @thomcc there is still no reply to https://sqlite.org/forum/forumpost/1921e60ce4
What should we do ?

thomcc:
I think just going one iteration here like the patch does is probably enough. I tried this at some point (to try and explain the issue in detail on the forums) and couldn't 100% reproduce (or didn't fully understand) what you described around multiple values -- it seemed to work.

gwenn:
> I think just going one iteration here like the patch does is probably enough. I tried this at some point (to try and explain the issue in detail on the forums) and couldn't 100% reproduce (or didn't fully understand) what you described around multiple values -- it seemed to work.

So I guess something has been fixed in SQLite3.
Because it failed here https://github.com/rusqlite/rusqlite/pull/996/commits/95794a25b102381970a2b033d8c10c3adc65bb2c
I will try to investigate

gwenn:
```tcl
% load /usr/lib/sqlite3.38.2/libtclsqlite3.so
% sqlite3 db1
% db1 eval {PRAGMA foreign_keys=1}
% db1 eval {CREATE TABLE items (id INTEGER PRIMARY KEY);}
% db1 eval {CREATE TABLE trades (item INTEGER REFERENCES items(id) NOT NULL);}
% db1 onecolumn {INSERT INTO trades (item) VALUES (42) RETURNING *}
FOREIGN KEY constraint failed
```
I guess we can close #995 / reject this PR (or keep only its tests)...

thomcc:
That's good news. If you'd like to keep its tests, that's fine. I'm fine closing too.

PR #998, opened 2021-07-24
Add feature for page checksums via cksumvfs.

andrewchambers:
See https://www.sqlite.org/cksumvfs.html for the official documentation on this extension.


andrewchambers:
I worked out how to do this from my own package.

andrewchambers:
Reopening in case someone else finds this useful.

thomcc:
> I worked out how to do this from my own package.

What did you end up doing? How bad is helps motivate including something like this in the core rusqlite (or a related crate, e.g. a hypothetical `rusqlite-checksum-vfs` or w/e).

andrewchambers:
I just had to duplicate the sqlite3 header (because I do a bundled build), then i could build the extension in static mode.from my own build.rs. Linking then worked fine.

PR #999, opened 2021-07-24
Enable LTO support for clang-like compiler when using feature "bundled" in libsqlite3-sys

NobodyXu:
This PR enables LTO support in libsqlite3-sys when the compiler is clang-like.

I've tested that it all tests inside libsqlite3-sys passed by adding:

```
[dev-dependencies]
libsqlite3-sys = { path = ".", features = ["bundled"] }
```

to `libsqlite3-sys/Cargo.toml` and run `cargo test --release`.

I've also run the test inside `rusqlite` with `features = ["bundled"]` for `libsqlite3-sys`.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -129,6 +129,10 @@ mod build_bundled {
             .flag("-D_POSIX_THREAD_SAFE_FUNCTIONS") // cross compile with MinGW
             .warnings(false);
 
+        if cfg.get_compiler().is_like_clang() {
+            cfg.flag_if_supported("-flto");
+        }
+
         if cfg!(feature = "bundled-sqlcipher") {
             cfg.flag("-DSQLITE_HAS_CODEC").flag("-DSQLITE_TEMP_STORE=2");
 


thomcc:
There's more to this that's likely needed for it to actually LTO the Rust with the C. https://doc.rust-lang.org/rustc/linker-plugin-lto.html

But either way, IDT we want to do this — it will generally come with the downside that clang and rustc be compiled with the same version of LLVM, which sometimes is easy, and sometimes massively inconvenient.

NobodyXu:
Thanks for the feedback!

I think I can add the `-C linker-plugin-lto` flags to `.cargo/config` and check for [llvm version used by cargo](https://users.rust-lang.org/t/print-llvm-version-used-by-rust-toolchain/43519) and the llvm version used by clang by running the compiler with the `—version` flag.

I will have a try and update the PR if succeeded.

thomcc:
.cargo/config isn't replicated for users of the crate via crates.io, so it wouldn't help outside this crate.

I think LTO is too finicky to be worth trying to integrate with the bundled feature. Again, it requires matching versions of LLVM used by clang and rustc, which can be quite tricky. I think if you want LTO, you probably should manually build and link.

thomcc:
I also don't expect it to help in this case — I'm guessing you're hoping for it to devirtualize the stuff in vtable.rs, but I don't think in practice you'll see that, given the way that virtual tables happen internally for sqlite.

NobodyXu:
> .cargo/config isn't replicated for users of the crate via crates.io, so it wouldn't help outside this crate.
> 
> I think LTO is too finicky to be worth trying to integrate with the bundled feature. Again, it requires matching versions of LLVM used by clang and rustc, which can be quite tricky. I think if you want LTO, you probably should manually build and link.

It looks like it is better to just compile sqlite3 with LTO manually and then use `PKG_CONFIG_PATH` to inject the manually writen `.pc` file though.

I think I can do this :D

NobodyXu:
> I also don't expect it to help in this case — I'm guessing you're hoping for it to devirtualize the stuff in vtable.rs, but I don't think in practice you'll see that, given the way that virtual tables happen internally for sqlite.

Wait, doesn’t LTO mean that it is possible to perform optimization accross the object files?

I thought the compiler is smart enough to devirtualize C-style function pointers, but I can be wrong.

Last time I heard, devirtualizing C style function pointers isn’t as simple as vtable in C++ and rust due to the mutability, lack of type information and the fact that it is coded by the programmer.

thomcc:
> Wait, doesn’t LTO mean that it is possible to perform optimization accross the object files?

Yes.

> I thought the compiler is smart enough to devirtualize C-style function pointers, but I can be wrong.

It depends on the case. It usually requires the compiler to be able to conclusively identify the thing thats being called. What Sqlite does ends up with the virtual table being put into a data structure, and then looked up later when preparing the statement, and then the functions are called when stepping the statement.

This is too indirect to be reliable for compilers. I'd be very surprised if it could manage something like that. Maybe in the future, but it seems a ways off...

Also note that the function call overhead, even an indirect function call like this, is probably pretty small compared to a lot of what SQLite does.

NobodyXu:
> Also note that the function call overhead, even an indirect function call like this, is probably pretty small compared to a lot of what SQLite does.

Indeed, SQLite does a lot than `vtable.rs`, which is simply a random generator, the overhead of calling into it might as well be shadowed by the interpreter in SQLite and memory allocation.

Though I do think linking with SQLite using `LTO` can throw away many unused function and reduce the overhead of setting `Values`, which might not make a difference at all xD.

Speaking of which, the `vtab` isn’t easy to implement and took me some time to look at the SQLite doc to understand the API.

`rusqlite::vtab::VTab::best_index` is the most confusing method, with `IndexInfo` mixing the input parameters and the required output parameters and the `idx_num` not being super clear at first.

Maybe I can submit a PR later on that part, which I think do need improvements.

PR #1000, opened 2021-07-25
Fix confusing typename used in doc for VTab and VTabCursor

NobodyXu:

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -198,7 +198,7 @@ impl VTabConnection {
 /// #[repr(C)]
 /// struct MyTab {
 ///    /// Base class. Must be first
-///    base: ffi::sqlite3_vtab,
+///    base: rusqlite::vtab::sqlite3_vtab,
 ///    /* Virtual table implementations will typically add additional fields */
 /// }
 /// ```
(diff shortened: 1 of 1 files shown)


thomcc:
I'm not an expert on rusqlite's vtable infrastructure, so @gwenn should probably be the one to review these when he has time.

NobodyXu:
> I'm not an expert on rusqlite's vtable infrastructure, so @gwenn should probably be the one to review these when he has time.

This PR mainly fixes the document for VTab and VTabCursor.

Their doc uses the namespace ffi, however in rusqlite it is private.

So I changed to the actual path where user of this crate can actually access.

NobodyXu:
@thomcc If you take a look at the [vtab doc](https://docs.rs/rusqlite/0.25.3/rusqlite/vtab/index.html), you will find sqlite3_vtab and sqlite3_vtab_cursor to be present in it, however the doc for VTab and VTabCursor contradicts this and used internal namespace.

thomcc (approved):
Oh, right. My comment was actually more about the other PRs. This is fine.

thomcc:
Ah, actually, it should be `rusqlite::ffi::sqlite_vtab` and such. I'll fix it though.

thomcc:
Oh, they're `pub use`d. This is fine, then.

PR #1001, opened 2021-07-25
 Improve API VTab::best_index: Separate mutable and immutable data 

NobodyXu:
This PR depends on the PR #1000 .

Mixing mutable and immutable data has several problems, one is that you cannot loop over `IndexConstraintIter` while setting respective `IndexConstraintUsage`.

This makes it hard to implement `VTab::best_index`.

Another problem is that it is confusing, as programmer do not know which part is for reading and which part is required to be set for the `VTab::best_index` implementation to works.

This commit separates mutable from the immutable, by passing additional param `constraint_usages` and makes `info` to be immutable, and requires the implementation to returns a `BestIndex` so that these fields will always be explicitly initialized.

~~param `constraint_usages` also automatically sets each `IndexConstraintUsage` to some sane values before passing it to `VTab::best_index` implementation.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,6 +136,7 @@ name = "deny_single_threaded_sqlite_config"
 
 [[test]]
 name = "vtab"
+required-features = ["vtab"]
 
 [[bench]]
 name = "cache"
--- a/src/vtab/array.rs
+++ b/src/vtab/array.rs
@@ -34,7 +34,9 @@ use std::rc::Rc;
 use crate::ffi;
 use crate::types::{ToSql, ToSqlOutput, Value};
 use crate::vtab::{
-    eponymous_only_module, Context, IndexConstraintOp, IndexInfo, VTab, VTabConnection, VTabCursor,
+    eponymous_only_module, Context, IndexConstraintOp,
+    IndexInfo, IndexConstraintUsages, BestIndex,
+    VTab, VTabConnection, VTabCursor,
     Values,
 };
 use crate::{Connection, Result};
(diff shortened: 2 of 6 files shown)


gwenn:
> Mixing mutable and immutable data has several problems, one is that you cannot loop over `IndexConstraintIter` while setting respective `IndexConstraintUsage`.

Fixed by #1006 

> Another problem is that it is confusing, as programmer do not know which part is for reading and which part is required to be set for the `VTab::best_index` implementation to works.

I don't see how you can be confused by: https://docs.rs/rusqlite/0.25.3/rusqlite/vtab/struct.IndexInfo.html
 
> This commit separates mutable from the immutable, by passing additional param `constraint_usages` and makes `info` to be immutable, and requires the implementation to returns a `BestIndex` so that these fields will always be explicitly initialized.

And now you will be (when rusqlite is on par with SQLite) forced to specify all of these:
```
    pub idxNum: ::std::os::raw::c_int,
    pub idxStr: *mut ::std::os::raw::c_char,
    pub needToFreeIdxStr: ::std::os::raw::c_int,
    pub orderByConsumed: ::std::os::raw::c_int,
    pub estimatedCost: f64,
    pub estimatedRows: sqlite3_int64,
    pub idxFlags: ::std::os::raw::c_int,
    pub colUsed: sqlite3_uint64,
```

gwenn:
For reference:
https://rogerbinns.github.io/apsw/vtable.html#apsw.VTTable.BestIndex
https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_vtable.go#L641
https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_vtable.go#L368

NobodyXu:
> For reference:
> https://rogerbinns.github.io/apsw/vtable.html#apsw.VTTable.BestIndex
> https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_vtable.go#L641
> https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_vtable.go#L368

I realized that the problem with `VTab` is actually a documentation problem.

After I looked at the document above, I immediately realize what `VTable` is doing, however the document of `VTab` in rusqlite3 is just not clear enough that I am often confused at what to do.

NobodyXu:
Closed this in favor of #1006

gwenn:
Version 0.26.0 released

PR #1002, opened 2021-07-25
[HELP WANTED] Add `VTabSafe` and `CreateVTabSafe` so that `unsafe impl` is no longer required

NobodyXu:
This PR depends on #1001 .

This PR adds `VTabSafe` and `CreateVTabSafe` which are trait without `unsafe` and modules can be created using `read_only_module_safe` and `eponymous_only_module_safe` respectively.

This would make it much easier to implement without the hassle of `sqlite3_vtab` and `#[repr(C)]`, which user of this crate might forget to add to their own vtable implementation.

**HELP WANTED**

After updating `csvtab` to use `CreateVTabSafe`, it failed to open `test.csv`, which is extremely confusing for me and I failed to fixed it.

I would appreciate if anyone can help me fix the error!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,6 +136,7 @@ name = "deny_single_threaded_sqlite_config"
 
 [[test]]
 name = "vtab"
+required-features = ["vtab"]
 
 [[bench]]
 name = "cache"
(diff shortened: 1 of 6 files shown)


NobodyXu:
Closed since #1001 is closed

PR #1012, opened 2021-08-17
Add as_type_or_null to ValueRef

peku33:
In most cases a column will contain just single type, like `Text` or `Real`.
However very often it will contain mixed `Nulls` with non null values.

Typical use case for non-null column from `test_get_raw`:
```
        while let Some(row) = rows.next()? {
            let i: i64 = row.get_ref(0)?.as_i64()?;
        }
```

Current approach if column can be either i64 or null:
```
        while let Some(row) = rows.next()? {
            let i: Option<i64> = match row.get_ref(0)? {
                ValueRef::Null => None,
                ValueRef::Integer(i) => Ok(Some(i)),
                _ => return Err(FromSqlError::InvalidType),
            }
        }
```

New approach allows:
```
        while let Some(row) = rows.next()? {
            let i: Option<i64> = row.get_ref(0)?.as_i64_or_null()?;
        }
```

### DIFF
--- a/src/types/value_ref.rs
+++ b/src/types/value_ref.rs
@@ -45,6 +45,19 @@ impl<'a> ValueRef<'a> {
         }
     }
 
+    /// If `self` is case `Null` returns None.
+    /// If `self` is case `Integer`, returns the integral value.
+    /// Otherwise returns [`Err(Error::InvalidColumnType)`](crate::Error::
+    /// InvalidColumnType).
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Hmm, this is a good addition for now, but eventually we'll want a more general solution (that solution is likely blocked until GATs stabilize though)

Issue #914, opened 2021-03-08
Segfault on cross compiling to musl


enaut:
### Versions

- **Rust: x86_64-unknown-linux-musl**
- **Diesel: 1.4**
- **Database: SQLITE**

### Feature Flags

- **diesel: features = ["sqlite"]**

## Problem Description

When cross compiling to `x86_64-unknown-linux-musl` target diesel segfaults immediately in line: https://github.com/diesel-rs/diesel/blob/bec389a1ea329c161fc7a8fb8df8bea29a3432bf/diesel/src/sqlite/connection/raw.rs#L23

### Workaround

In #musl IRC I was given the workaround to add:
```
libsqlite3-sys = { version = "*", features = ["bundled"] }
```
to `Cargo.toml` solves this issue.

So somehow even when cross compiling the libsqlite3-sys is not bundled but dynamically linked which fails.

### What are you trying to accomplish?

I want to use musl builds to have a standalone binary of my actix-web app.

### Steps to reproduce

I created a minimal example: https://github.com/enaut/musl-segfault-sample

compile and run on linux with: `cargo run --target x86_64-unknown-linux-musl`

avkonst:
I confirm the problem. I have got the same.

jbsiddall:
Im having same issue, but i'm using postgres instead of sqlite. I've not tried the workaround yet.

thomcc:
I think this comes from a dynamic/static C runtime mismatch, but am not 100% certain.

enaut:
It may be helpful to automagically add the "bundled" feature when compiling for musl target - or a similar condition?

hellerbarde:
Potentially related: https://github.com/tonarino/innernet/issues/228#issuecomment-1198990309

rolandjitsu:
I've also stumbled into a segfault: https://github.com/rusqlite/rusqlite/issues/1290#issuecomment-2292664271.

PR #1021, opened 2021-09-27
MSVC: Allow static linking against the Visual C++ Runtime 

Christopher22:
This small contributution enables static linking against the Microsoft Visual C++ Runtime [if it is explicitely requested by the user](https://github.com/rust-lang/rfcs/blob/master/text/1721-crt-static.md). Without the flag for MSVC, the linking may fail given the implicit dependency on _vcruntime140.dll_ when using "bundled-" features.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -221,6 +221,11 @@ mod build_bundled {
             cfg.flag("-fsanitize=address");
         }
 
+        // If explicitly requested: enable static linking against the Microsoft Visual C++ Runtime to avoid dependencies on vcruntime140.dll and similar libraries.
+        if cfg!(target_feature = "crt-static") && is_compiler("msvc") {
+            cfg.static_crt(true);
+        }
+
         // Older versions of visual studio don't support c99 (including isnan), which
         // causes a build failure when the linker fails to find the `isnan`
         // function. `sqlite` provides its own implementation, using the fact


gwenn (approved):
I am not sure that we really need `is_compiler("msvc")`.

thomcc:
I think this should get the data from the `CARGO_CFG_TARGET_FEATURE` env var. As it is, it checks the target feature of the build.rs, not at runtime.

See https://github.com/BurntSushi/memchr/blob/master/build.rs#L70-L74

Also, yeah, I think the `is_compiler("msvc")` check is not needed.

Christopher22:
Thanks for your review! We do not need the `is_compiler("msvc")` as it is checked in cc internally. Just added it for the sake of clarity. I'm happy to remove it :)

Regarding the `CARGO_CFG_TARGET_FEATURE`, I'm a bit unsure. [By default, cc already checks for this env var](https://github.com/alexcrichton/cc-rs/blob/114437e178739d4c4c195b551524bd560f54fc34/src/lib.rs#L1449). However, it does not appear to be enough in my local test case which compiles and links well only when checking for the target feature of the build.rs additionally:

```
[...]
= note:    Creating library C:\Users\Admin\Documents\Projekte\matryoshka-sqlite\target\release\deps\matryoshka.dll.lib and object C:\Users\Admin\Documents\Projekte\matryoshka-sqlite\target\release\deps\matryoshka.dll.exp
          LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
          LINK : warning LNK4217: symbol 'free' defined in 'libucrt.lib(free.obj)' is imported by 'liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o)' in function 'sqlite3MemFree'
          LINK : warning LNK4217: symbol 'malloc' defined in 'libucrt.lib(malloc.obj)' is imported by 'liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o)' in function 'sqlite3MemMalloc'
          LINK : warning LNK4217: symbol '_msize' defined in 'libucrt.lib(msize.obj)' is imported by 'liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o)' in function 'sqlite3MemSize'
          LINK : warning LNK4217: symbol 'qsort' defined in 'libucrt.lib(qsort.obj)' is imported by 'liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o)' in function 'sqlite3Fts3SegReaderPending'
          LINK : warning LNK4217: symbol 'strncmp' defined in 'libucrt.lib(strncmp.obj)' is imported by 'liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o)' in function 'execSql'
          liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o) : error LNK2019: unresolved external symbol __imp_realloc referenced in function sqlite3MemRealloc
          liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o) : error LNK2019: unresolved external symbol __imp_strcspn referenced in function patternCompare
          liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o) : error LNK2019: unresolved external symbol __imp__localtime64_s referenced in function localtimeOffset
          liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o) : error LNK2019: unresolved external symbol __imp__beginthreadex referenced in function sqlite3ThreadCreate
          liblibsqlite3_sys-e960bef6b47b98ed.rlib(sqlite3.o) : error LNK2019: unresolved external symbol __imp__endthreadex referenced in function sqlite3ThreadProc
          C:\Users\Admin\Documents\Projekte\matryoshka-sqlite\target\release\deps\matryoshka.dll : fatal error LNK1120: 5 unresolved externals
```

thomcc:
Hm, that's very strange. I wonder if it's a `cc` bug.

thomcc (approved):
This is probably fine.

I think we can leave the `is_compiler("msvc")` here for now. I had both forgotten that the option for gcc/clang is different (`cc::Build::static_flag`), but also, in theory this shouldn't be needed anywhere — the env var should be enough, and the `cfg!` should be unrelated...

Since this is apparently untrue for unknown reasons... we can keep sort of thing limited targets where it turns out to actually be required, like (so far) the `msvc` windows targets, as the code currently does.

---

Anyway, mostly as a note to myself, I think this there's some weirdness with static targets and libsqlite3-sys, but it could be `cc` or the toolchain itself.

That is, something is probably broken somewhere in `libsqlite3-sys`'s build.rs, a library it uses (`cc`, one of the package locators, ...), or even cargo/rustc (or how it invokes the linker — static-crt is rarely used enough for it to seem possible to have some bugs still)... 

I think whatever it is is responsible for bugs like this, but also weird static/dynamic runtime issues like #914 (or maybe I'm wrong and these are coincidences).... Or maybe I just don't understand how this is supposed to work, and it just seems buggy. Could be!

I think trying to figure out if things behave as expected in cases like:
- gcc/clang targets with `-static` flag (e.g. `cc::Build::static_flag`) and +static-crt...
- mingw windows targets (also `+static-crt`, also `-static`)...
- `*-linux-musl` with/without `static-crt`. (I think "without" is still not fully supported by rustc, though, just planned)...

and others. some testing on targets might root out the bug... Or at least, help point out what I've misunderstood.

---

None of this really makes a difference in the short term, so I think we can land this.

Issue #1025, opened 2021-10-05
`bundled-sqlcipher-vendored-openssl` doesn't compile on windows


douloiha:
When using the `bundled-sqlcipher-vendored-openssl` feature, it produces the following error:

```
could not find native static library `crypto`, perhaps an -L flag is missing?
```

I guess this is because in the `libsqlite3-sys/build.rs` file, line 184 doesn't distinguish windows and linux crypto library. On windows, the library is named libcrypto (and not crypto).

thomcc:
Yeah, I think it's called something else on windows (libeay32 i think). Anyway, it's known to be unsupported (the README calls bundled-sqlcipher "more or less Unix only at the moment"). But honestly I had kinda intended to get back to it at some point and didnt, 

I'll see if I can figure something out, since I don't really think surprise breakage on windows is ok, but... hrm.

PR #1027, opened 2021-10-06
Fix OpenSSL linking on Windows

BlackHoleFox:
fwiw, this is a breaking change moving from `libeay32` to `libcrypto` but no one should be using an openssl version from <=2016.

Also took the chance to clean up a bit of the path handling building the library path on Windows when linking against an already pre-built OpenSSL somewhere on the system.

Fixes https://github.com/rusqlite/rusqlite/issues/1025


thomcc:
In case there's confusion, the extern crate does make sense.

If we don't use anything from openssl_sys, it will not be guaranteed to get linked in, and thus the libs *it* depends on wont get brought in either — we don't link to it transitively, so we don't need it.

In practice, this only shows up on some targets. In practice, windows works without it, but not macOS or Linux.

PR #1028, opened 2021-10-06
Update readme, prep for a cutting a libsqlite3-sys@v0.23.1

thomcc:
I think everything is compatible, so we only need a patch release 0.23.1, and only for the -sys. This just fixes the bundled-sqlcipher and -vendored-openssl logic so that it 

Thanks to @douloiha for (even unintentionally) reminding me about the fact I this should get into a followup https://github.com/rusqlite/rusqlite/pull/860#issuecomment-853304210, and to @BlackHoleFox for actually doing the work after I whined about nonesense elsewhere online.

(Also thanks to @dubiousjim for doing the initial version of the feature, and sorry its so long before it became usable)


thomcc:
Published `libsqlite3-sys` version v0.23.1 https://crates.io/crates/libsqlite3-sys/0.23.1

PR #1041, opened 2021-11-15
libsqlite3-sys: darwin: fix crypto link framework

ur0:
SecurityFoundation.framework is an unrelated macOS framework which does not directly provide cryptography functions. This framework also doesn't exist on iOS, which breaks linking This commit changes to link against Security.framework instead, as noted in the SQLCipher [docs](https://www.zetetic.net/blog/2013/6/27/sqlcipher-220-release.html).


thomcc (approved):

thomcc:
Nice catch.

PR #1042, opened 2021-11-22
Include varying positional param types in docs

### LINKED ISSUE #1029: [Docs] "Positional params of varying types" doesn't have varying types?
I think this is an example of there being room for more helpful docs: https://docs.rs/rusqlite/0.26.0/rusqlite/struct.Connection.html#with-positional-params-of-varying-types

image

This only has a single parameter, so does not demonstrate varying types (as far as I understand). I'm open to submitting a PR with a more elaborate example if this is the case.

daanschutte:
Resolves #1029.

Currently the documentation for `lib::execute` has the same examples for both `With positional params` and `With positional params of varying types`.

In this PR I updated the documentation with a more helpful example.

Disclaimer: I am new to both Rust and `rusqlite` so apologies in advance if this is not the best method to go about this 🙃

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -513,7 +513,7 @@ impl Connection {
     /// ### With positional params of varying types
     ///
     /// ```rust,no_run
-    /// # use rusqlite::{Connection};
+    /// # use rusqlite::{params, Connection};
     /// fn update_rows(conn: &Connection) {
     ///     match conn.execute("UPDATE foo SET bar = 'baz' WHERE qux = ?1 AND quux = ?2", params![1i32, 1.5f64]) {
     ///         Ok(updated) => println!("{} rows were updated", updated),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -515,7 +515,7 @@ impl Connection {
     /// ```rust,no_run
     /// # use rusqlite::{Connection};
     /// fn update_rows(conn: &Connection) {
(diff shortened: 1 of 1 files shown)


thomcc (approved):

daanschutte:
I see the builds are failing due to a missing import (they all passed locally on Linux for me). 

I'll add the import and we can try again.

daanschutte:
I see the particular command that failed in the CI build was `cargo test --features bundled --workspace --doc --verbose `, which is not in the [contributing checklist](https://github.com/rusqlite/rusqlite/blob/master/README.md#checklist) in the README. 

Would you like me to open a PR to add this or is this not a common enough problem?

thomcc:
> Would you like me to open a PR to add this or is this not a common enough problem?

A PR is welcome. I actually think this probably gets tested if we remove `--all-targets` from the line there, which perhaps we should suggest.

Does `cargo test --workspace --features bundled` cover it?

daanschutte:
> Does `cargo test --workspace --features bundled` cover it?

It does indeed. I'll open a PR with the suggestion 👍🏻

thomcc:
Yeah, if you could remove `--all-targets` from those commands. It does have an impact (in CI it will check the benchmark files too), but these are so rarely modified compared to docs that we should remove it from contribution guide.

PR #1046, opened 2021-11-28
Split upgrade script

gwenn:
Such as we can upgrade SQLite / SQLCipher without regenerating SQLCipher / SQLite files


thomcc (approved):
LGTM

Issue #1048, opened 2021-11-29
screwy update_hook lifetime raises SIGSEGV


sqwishy:
The following compiles (in release mode) but will segfault trying to unlock the mutex after the mutex is destroyed. The example runs fine if `move` is added to the closure because the mutex isn't dropped after `update_hook()`.

The `Arc` & `Mutex` isn't important, it could be a `Vec` or anything but it's allowed to be dropped after `update_hook()`.  I guess ideally the lifetimes should be a bit more precise so this can't compile?

``` rust
use rusqlite::{hooks::Action, Connection};
use std::sync::{Arc, Mutex};

fn main() {
    let db = Connection::open_in_memory().unwrap();

    {
        let locked: Arc<Mutex<()>> = Arc::new(Mutex::new(()));

        let hook = |_: Action, _: &str, _: &str, _: i64| {
            if let Ok(ref mut mutex) = locked.try_lock() {
                **mutex = ();
            }
        };

        db.update_hook(Some(hook));
    };

    db.execute("CREATE TABLE tbl(i integer)", []).unwrap();
    db.execute("INSERT INTO tbl (i) VALUES (1)", []).unwrap();
}
```

```
rusqlite = { version = "0.26", features = ["hooks"] }
```

Also tested on master (currently at d70fbac2314333ecf2e67db43dbb43b0a47a4797).

gwenn:
There is already a lifetime constraint:
https://github.com/rusqlite/rusqlite/blob/dcaa67a6172571f4e7d26a879d77da21bcacaf71/src/hooks.rs#L378-L380
but it doesn't work!?

gwenn:
Bug introduced by https://github.com/rusqlite/rusqlite/commit/926977846fd0c4ec8e77fcaa9a6c39e9021a4663 ?

jgallagher:
That lifetime constraint doesn't work as intended. Here's a playground example with a similar lifetime that ends up storing a raw pointer to an address on the stack that is no longer valid: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0c702334cac58af1536eca86114e81b3

IIUC, the `'c` lifetime on `update_hook()` just means the compiler has to find some lifetime that can be satisfied where both `&self` and `F` are alive; it does _not_ mean "the full lifetime of self". In this example code at the top, the lifetime it finds is the lifetime of the reference to locked: both the closure and `&self` live at least that long, so it compiles.

[...]

PR #1049, opened 2021-11-30
Fix callbacks lifetime

gwenn:
See #1048

Also:
- [X] https://github.com/rusqlite/rusqlite/blob/master/src/functions.rs#L414
- [X] https://github.com/rusqlite/rusqlite/blob/master/src/functions.rs#L440

?


thomcc (approved):
Oof, we got this wrong everywhere, huh... We should get this into a release and also file an advisory. I'll do this later today, unless you have hesitations.

PR #1052, opened 2021-12-06
Check callbacks lifetime

gwenn:


thomcc:
I actually think this is a bad change. I don't think UnwindSafe works well for FFI code — we only have `catch_unwind` because it's currently UB to panic over the `extern "C"` boundary. That's being fixed, and when it gets fixed, we'll remove the catch_unwind.

gwenn:
> I actually think this is a bad change. I don't think UnwindSafe works well for FFI code — we only have `catch_unwind` because it's currently UB to panic over the `extern "C"` boundary. That's being fixed, and when it gets fixed, we'll remove the catch_unwind.

Ok, so we should remove all `UnwindSafe` / `RefUnwindSafe` constraints (`progress_handler` already has this contraint) ?
While adding these doc tests, `rustc` raised a warning that a `mut bool` is not `UnwindSafe` (if I remember correctly) which  seems useful.

gwenn:
std library seems to use both (`RefUnwindSafe` and `catch_unwind`) here:
https://github.com/rust-lang/rust/blob/master/library/std/src/rt.rs#L107
https://github.com/rust-lang/rust/blob/master/library/std/src/rt.rs#L128

Same here:
https://github.com/Michael-F-Bryan/ffi_helpers/blob/da298c854ad3843f82b82c71900643fbb64dd7dc/src/panic.rs#L43-L45

PR #1054, opened 2021-12-08
Backport fix of closure lifetime issue to v0.25.x

thomcc:
This is the code for #1049 — had to apply the patch by hand due to formatting changes. This shouldn't be merged, but I *think* this will let me run CI on it to make sure nothing got messed up.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
+      - hotfixes-*
   pull_request:
     branches:
       - master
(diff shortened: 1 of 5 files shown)


thomcc:
Hm, I guess that I can't run CI this way.

thomcc:
Well, that didn't work. Oh well, the patch passes all tests for me locally, and not having the CI set up right doesn't seem like a reason to block a release of this, so I'll figure this out some other day, but just release that patch as it is (if it turns out there are issues, I'll push out another fix, ofc)

Issue #1059, opened 2021-12-10
Add support for Decimal type


mamcx:
Similar to [rust postgres](https://github.com/paupino/rust-decimal/blob/master/src/postgres.rs), it will be great if this lib add support for the[ Decimal type](https://docs.rs/rust_decimal/latest/rust_decimal/).

My case is that I'm building my own mini-orm, and differently to rust postgres, I can't implement ToSql/FromSql for it:

```rust
impl FromSql for Decimal {
    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
        unimplemented!()
    }
}

```

```bash
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
   --> db/src/sqlite.rs:220:1
    |
220 | impl FromSql for Decimal {
    | ^^^^^^^^^^^^^^^^^-------
    | |                |
    | |                `rust_decimal::Decimal` is not defined in the current crate
    | impl doesn't use only types from inside the current crate
    |
    = note: define and implement a trait or new type instead
```

gwenn:
The `orphan rule` means that only `rust_decimal` or `rusqlite` crates can implement `FromSql`/`ToSql` for `Decimal` directly.

mamcx:
Oh, I see. I ask there. Thanks.

mamcx:
Ok, so can I put a pull request for this here?

gwenn:
Yep, something like https://github.com/rusqlite/rusqlite/pull/506

mamcx:
Making the test I found is not possible to round-trip [Decimal::Max](https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#associatedconstant.MAX), and it fails with `InvalidColumnType` instead of report is an overflow.

In [get<I: RowIndex, T: FromSql>(https://github.com/rusqlite/rusqlite/blob/master/src/row.rs#L280)  it returns this error.

Can I return the overflow instead or let it be as-is and you look at it later?

gwenn:
You should be able to use [FromSqlError::Other](https://docs.rs/rusqlite/latest/rusqlite/types/enum.FromSqlError.html#variant.Other), no ?

mamcx:
Yeah, but is still returned as `InvalidType`, no matter the original error...

mamcx:
Ok, I send the pull request, tell me if I need to improve it anyway!

PR #1061, opened 2021-12-14
Adding support for rust_decimal

mamcx:
Adding initial support [for decimal type ](https://github.com/rusqlite/rusqlite/issues/1059).

NOTE: Round-trip from [Decimal::MAX](https://docs.rs/rust_decimal/1.18.0/rust_decimal/struct.Decimal.html#associatedconstant.MAX) currently fail.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ modern-full = [
     "collation",
     "column_decltype",
     "csvtab",
+    "rust_decimal",
     "extra_check",
     "functions",
     "hooks",
@@ -110,6 +111,7 @@ fallible-iterator = "0.2"
 fallible-streaming-iterator = "0.1"
 memchr = "2.3"
 uuid = { version = "0.8", optional = true }
+rust_decimal = { version = "1.18.0", optional = true }
 smallvec = "1.6.1"
 
 [dev-dependencies]
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-s
 * [`csvtab`](https://sqlite.org/csv.html), CSV virtual table written in Rust. (Implies `vtab`.)
 * [`array`](https://sqlite.org/carray.html), The `rarray()` Table-Valued Function. (Implies `vtab`.)
 * `i128_blob` allows storing values of type `i128` type in SQLite databases. Internally, the data is stored as a 16 byte big-endian blob, with the most significant bit flipped, which allows ordering and comparison between different blobs storing i128s to work as expected.
+* `rust_decimal` allows storing and retrieving `Decimal` values from the [`rust_decimal`](https://docs.rs/rust_decimal/) crate.
 * `uuid` allows storing and retrieving `Uuid` values from the [`uuid`](https://docs.rs/uuid/) crate using blobs.
 * [`session`](https://sqlite.org/sessionintro.html), Session module extension. Requires `buildtime_bindgen` feature. (Implies `hooks`.)
 * `extra_check` fail when a query passed to execute is readonly or has a column count > 0.
--- a/src/row.rs
+++ b/src/row.rs
@@ -302,6 +302,12 @@ impl<'stmt> Row<'stmt> {
                 self.stmt.column_name_unwrap(idx).into(),
                 value.data_type(),
             ),
+            #[cfg(feature = "rust_decimal")]
+            FromSqlError::InvalidDecimal(_) => Error::InvalidColumnType(
+                idx,
+                self.stmt.column_name_unwrap(idx).into(),
+                value.data_type(),
+            ),
         })
     }
 
--- a/src/types/decimal.rs
+++ b/src/types/decimal.rs
@@ -1,0 +1,79 @@
+//! [`ToSql`] and [`FromSql`] implementation for [`rust_decimal::Decimal`].
+use crate::types::{FromSql, FromSqlError, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
+use crate::Result;
+use rust_decimal::Decimal;
+use std::str::FromStr;
+
+fn parse_from_str(s: &str) -> Result<Decimal, FromSqlError> {
+    Decimal::from_str(s).map_err(FromSqlError::InvalidDecimal)
+}
+
+/// Serialize `Decimal` to text.
+impl ToSql for Decimal {
+    #[inline]
+    fn to_sql(&self) -> Result<ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::from(self.to_string()))
+    }
+}
+
+/// Deserialize to `Decimal`.
+impl FromSql for Decimal {
+    #[inline]
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        match value {
+            ValueRef::Integer(x) => Ok(Decimal::from(x)),
+            //Note: Parsing straight from f64 could cause loss of precision, when parsing from text works fine
+            // like for example f64::MAX
+            ValueRef::Real(x) => parse_from_str(&x.to_string()),
+            ValueRef::Text(_) => value.as_str().and_then(parse_from_str),
+            _ => Err(FromSqlError::InvalidType),
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::types::Type;
+    use crate::{params, Connection, Error, Result};
+
+    fn checked_memory_handle() -> Result<Connection> {
+        let db = Connection::open_in_memory()?;
+        db.execute_batch("CREATE TABLE Decimals (i INTEGER, v DECIMAL)")?;
+        Ok(db)
+    }
+
+    fn get_decimal(db: &Connection, id: i64) -> Result<Decimal> {
+        db.query_row("SELECT v FROM Decimals WHERE i = ?", [id], |r| r.get(0))
+    }
+
+    #[test]
+    fn test_sql_decimal() -> Result<()> {
+        let db = &checked_memory_handle()?;
+
+        let zero = Decimal::from(0);
+        let max_float = Decimal::from_str(&f64::MAX.to_string()).unwrap();
+        let max_decimal = Decimal::MAX;
+
+        db.execute(
+            "INSERT INTO Decimals (i, v) VALUES (0, ?), (1, ?), (2, ?), (3, ?)",
+            // also insert invalid data that will fail to decode to decimal
+            params![0, f64::MAX, max_decimal, "illegal"],
+        )?;
+
+        assert_eq!(get_decimal(db, 0)?, zero);
+        assert_eq!(get_decimal(db, 1)?, max_float);
+        // Currently the round-trip from Decimal::MAX can fail... The real error is a overflow
+        assert_eq!(
+            get_decimal(db, 2),
+            Err(Error::InvalidColumnType(0, "v".into(), Type::Real))
+        );
+        //This is in fact a parsing error...
+        assert_eq!(
+            get_decimal(db, 3),
+            Err(Error::InvalidColumnType(0, "v".into(), Type::Text))
+        );
+
+        Ok(())
+    }
+}
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -28,6 +28,12 @@ pub enum FromSqlError {
     #[cfg_attr(docsrs, doc(cfg(feature = "uuid")))]
     InvalidUuidSize(usize),
 
+    /// Error returned when reading a `Decimal`. Only available when the `rust_decimal`
+    /// feature is enabled.
+    #[cfg(feature = "rust_decimal")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "rust_decimal")))]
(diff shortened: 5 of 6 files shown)


gwenn:
Maybe we should try https://docs.rs/rust_decimal/1.18.0/rust_decimal/struct.Decimal.html#method.serialize and https://docs.rs/rust_decimal/1.18.0/rust_decimal/struct.Decimal.html#method.deserialize

mamcx:
I don't know how to use those methods, it needs to go deeper into the FFI?

gwenn:
See https://github.com/mamcx/rusqlite/pull/1

mamcx:
I need to do something else?

gwenn:
> I need to do something else?

See https://github.com/rusqlite/rusqlite/runs/4564693209?check_suite_focus=true

thomcc:
Hmm, I'm worried about this, since people might assume it's compatible with the official SQLite decimal extension (e.g. <https://github.com/sqlite/sqlite/blob/master/ext/misc/decimal.c>), and it wouldn't be...

mamcx:
Hi, do I need to do anything else to see this merged, or just wait :)
P.D: This my first pull request ever so I unsure what are the full steps...

gwenn:
> Hi, do I need to do anything else to see this merged, or just wait :)
> P.D: This my first pull request ever so I unsure what are the full steps...

You/we need to discuss the concern raised by @thomcc !

mamcx:
> You/we need to discuss the concern raised by @thomcc !

Ah, great. 

---

I think is hard to find this problem as stated by  @thomcc, in the sense that SQLite and its functions/libraries are "in-built", running inside.  I don't think others crates (and in other languages!) exist the expectation that Json, Date, Regex, etc are exactly as the spec in the RDBMS...

mamcx:
Any blocker left? Or is now wait? Can I help?

thomcc:
Right, but we generally do try for compatibility with things stored in the database. Regexes are not stored in the DB (and SQLite does not document their format), but JSON and dates are, and we are compatible with them.

thomcc:
I suppose because the feature name is `rust_decimal`, it's probably fine. I still have misgivings about it, though.

mamcx:
What else can be done to unblock this?

jannschu:
`From<f64>` is not implemented for Decimal for a reason. Bypassing that by casting to a string is incorrect.

gwenn:
> `From<f64>` is not implemented for Decimal for a reason. Bypassing that by casting to a string is incorrect.

You mean from SQLite3 REAL I guess. And it is:
https://github.com/rusqlite/rusqlite/pull/1061/files#diff-dca35f2dce05c5640f8c7f1acc59fdd46c6f34c881f4464029ac8922851770faR20

jannschu:
I meant `From<f64>` for `rust_decimal::Decimal`, which is [not implemented](https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html). One reason is, it is not possible due to `±Infinity` and `NaN` which have no equivalent in `Decimal`, but other reasons are:

- A floating point number has no precision in the sense that it does not store how many digits are significant whereas `rust_decimal::Decimal` does.
- Not _even every finite_ floating point number is representable in Decimal, because one is a base two number while the other is base 10.

`rust_decimal::Decimal` has two functions for converting from `f64` with some further explanations, `from_f64` and `from_f64_retain`. I would argue that a `FromSql` implementation should be lossless, which is not the case with

```rust
ValueRef::Real(x) => parse_from_str(&x.to_string()),
```

If you really want that you should cast to string in SQL in my opinion. With other words, the `ValueRef::Real(x)` case should fail. `f64` and `Decimal` are conceptually different things and if the user wants that the string conversion variant can be either achieved with explicit casts in SQL or you use `from_f64` or `from_f64_retain` in Rust.

You could say why not convert those `f64` that _are_ exactly representable, but that makes the implementation unpredictable in my opinion.

mamcx:
`parse_from_str` is from `rust_decimal::Decimal` so it follows its
rules for what is valid, I don't skip it.

The use of string here is orthogonal to your remarks: is about not
losing precision.

Note how the macros actually do the same:

https://github.com/paupino/rust-decimal/blob/master/macros/src/lib.rs

and the docs show the use of data in strings is pervasive. Taking form
f64 is riskier.

2022-09-06 11:55 GMT-05:00, jannschu ***@***.***>:
> I meant `From<f64>` for `rust_decimal::Decimal`, which is [not
> implemented](https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html).
> One reason is, it is not possible due to `±Infinity` and `NaN` which have no
> equivalent in `Decimal`, but other reasons are:

>

> - A floating point number has no precision in the sense that it does not
> store how many digits are significant whereas `rust_decimal::Decimal`
> does.

> - Not _even every finite_ floating point number is representable in Decimal,
> because one is a base two number while the other is base 10.

>

> `rust_decimal::Decimal` has two functions for converting from `f64` with
> some further explanations, `from_f64` and `from_f64_retain`. I would argue
> that a `FromSql` implementation should be lossless, which is not the case
> with

>

> ```rust

> ValueRef::Real(x) => parse_from_str(&x.to_string()),

> ```

>

> If you really want that you should cast to string in SQL in my opinion. With
> other words, the `ValueRef::Real(x)` case should fail. `f64` and `Decimal`
> are conceptually different things and if the user wants that the string
> conversion variant can be either achieved with explicit casts in SQL or you
> use `from_f64` or `from_f64_retain` in Rust.

>

> You could say why not convert those `f64` that _are_ exactly representable,
> but that makes the implementation unpredictable in my opinion.

>

> --

> Reply to this email directly or view it on GitHub:

> https://github.com/rusqlite/rusqlite/pull/1061#issuecomment-1238414204

> You are receiving this because you authored the thread.

>

> Message ID: ***@***.***>


-- 
Mario A.Montoya
http://www.elmalabarista.com

jannschu:
Comparing the macro `to_string` call with converting f64 to to_string makes no sense. The string conversion in the macro operates on language tokens, this has nothing to do with converting f64 to string.

I also never said `parse_from_str` is problematic. The problem is converting f64 to string.

jannschu:
Additional remark: People use something like `rust_decimal::Decimal` if they are specifically interested in the differences with f64. So converting f64 behind the scenes is problematic, even more so by doing via a string conversion.

jannschu:
I had a look into how the SQLite decimal extension stores data
 There a straight-forward string representation is used. So if instead of `serialize`/`deserialize` just `to_string` and `from_str` is there should be compatibility. The SQLite decimal is unbound however.

I think being able to use the SQLite decimal extensions would be a valuable tradeoff. The string representation also better fits the archiving goals of SQLite compared to the binary representation of `rust_decimal`.

Just to point this out once more: What I said about `f64` is unaffected.

Since @paupino offered some help in paupino/rust-decimal#449 for this task, maybe you can give us your opinion on these two points?

thomcc:
I guess would rather not add this. Here are the issues:

1. It's not really compatible with what people usually expect from decimal in databases (SQLite's and Postgres's DECIMAL are both arbitrary precision).

    Moreover, storing the result of `Decimal::serialize` is a huge breakage risk, because it's a format specific to the implementation[^1] (and it's current version). That is, if `rust-decimal` changes it (I'd hope this would be considered a breaking change so lets say in 2.0), then to update to `rust-decimal` to 2.0, `rusqlite` needs to add code that maintains compatibility with `rust-decimal@1`'s serialization, otherwise users of `rusqlite` who used it to store `Decimal`s in their DB can no longer read them. This could be fixed by using text, but that has performance costs, for example.
    
    Adding support to `rusqlite` means the user who wants to store Decimals in their DB is in control with how the data gets stored which is far better. For example, if they choose one format but need to update or change their decimal library, they can change it far more easily by running a migration -- `rusqlite` absolutely can't do that.

2. Honestly, the existing libraries we've supported here have caused a lot of trouble (`time` from breakage, `chrono` from it's maintenance gap, and `url` a long time ago also had semi-painful breakage, ...). We can't remove them without causing huge problems for users, but I definitely don't want to add more unless it's *really* worthwhile.

3. The crate is not really widely used enough. Of the existing crates we support here (`time`, `chrono`, `url`, `serde_json`, and `uuid`) even the least widely used has >40M downloads. (And even for these, I regret adding some of them, but don't want to leave people unable to migrate).

I do want to apologize for how I've been in this thread though. Honestly, I had a lot of trouble articulating why the first issue was such a problem, and so I kind of just ignored the issue. This was probably quite frustrating, and yeah, it's not good behavior on my part -- I'm sorry.

That said, I think it's a lot easier for users to just handle themselves. Usually it's the difference between something like `r.get::<_, Decimal>(0)` (if we had native support) and `Decimal::try_from(r.get::<_, Something>(0)?)` (e.g. an explicit conversion that fits on the same line).

[^1]: You'll note that all the existing crates parse/write data in well-understood formats you could easily parse with another library, or another language -- while the implementation of `rust-decimal`'s serialization is not unreasonable complex, it's still something I'd be very annoyed if I had to maintain in the future.

mamcx:
Using `String` alleviate all the issues with internal changes of
formats, and allows to transparently migrate to the new version (and
even support for things that are not rust decimal: I use sqlite across
many languages with different decimal libraries).

> I think it's a lot easier for users to just handle themselves.

I wish. Is not. My current workaround for creating new types and
reimplementing all the ops is one of the largest plumbing sections of
my code.

In most simple use cases, where at best you have a small number of
tables and all your code is hand-made could be, but quickly you add
serde, procedural macros from other things, etc and that is not that
simple (and I make an ERP/eCommerce and that means I need everything
under the sun!).

I understand the orphan rules are a hurdle, and rusqlite is not the
only crate I need to work around because of it. But being a numeric
type that is critical for correct business operation, the amount of
traits that need to support is large.

> The crate is not really widely used enough

Among the ones that could handle money values is the most popular and
complete, IMHO, so looking at the use case ("what deals with
decimal/currency") vs to the large ecosystem makes a different
calculation. If exist a better option for dealing with it I am open to
hearing it!

P.D: I have not a problem if is implemented here or in rust decimal,
so if the other side wanna it I can push there!





2022-09-07 3:02 GMT-05:00, Thom Chiovoloni ***@***.***>:
> I guess would rather not add this. Here are the issues:
>
> 1. It's not really compatible with what people usually expect from decimal
> in databases (SQLite's and Postgres's DECIMAL are both arbitrary
> precision).
>
>     Moreover, storing the result of `Decimal::serialize` is a huge breakage
> risk, because it's a format specific to the implementation[^1] (and it's
> current version). That is, if `rust-decimal` changes it (I'd hope this would
> be considered a breaking change so lets say in 2.0), then to update to
> `rust-decimal` to 2.0, `rusqlite` needs to add code that maintains
> compatibility with ***@***.***`'s serialization, otherwise users of
> `rusqlite` who used it to store `Decimal`s in their DB can no longer read
> them. This could be fixed by using text, but that has performance costs, for
> example.
>
>     Adding support to `rusqlite` means the user who wants to store Decimals
> in their DB is in control with how the data gets stored which is far better.
> For example, if they choose one format but need to update or change their
> decimal library, they can change it far more easily by running a migration
> -- `rusqlite` absolutely can't do that.
>
> 2. Honestly, the existing libraries we've supported here have caused a lot
> of trouble (`time` from breakage, `chrono` from it's maintenance gap, and
> `url` a long time ago also had semi-painful breakage, ...). We can't remove
> them without causing huge problems for users, but I definitely don't want to
> add more unless it's *really* worthwhile.
>
> 3. The crate is not really widely used enough. Of the existing crates we
> support here (`time`, `chrono`, `url`, `serde_json`, and `uuid`) even the
> least widely used has >40M downloads. (And even for these, I regret adding
> some of them, but don't want to leave people unable to migrate).
>
> I do want to apologize for how I've been in this thread though. Honestly, I
> had a lot of trouble articulating why the first issue was such a problem,
> and so I kind of just ignored the issue. This was probably quite
> frustrating, and yeah, it's not good behavior on my part -- I'm sorry.
>
> That said, I think it's a lot easier for users to just handle themselves.
> Usually it's the difference between something like `r.get::<_, Decimal>(0)`
> (if we had native support) and `Decimal::try_from(r.get::<_,
> Something>(0)?)` (e.g. an explicit conversion that fits on the same line).
>
> [^1]: You'll note that all the existing crates parse/write data in
> well-understood formats you could easily parse with another library, or
> another language -- while the implementation of `rust-decimal`'s
> serialization is not unreasonable complex, it's still something I'd be very
> annoyed if I had to maintain in the future.
>
> --
> Reply to this email directly or view it on GitHub:
> https://github.com/rusqlite/rusqlite/pull/1061#issuecomment-1239047459
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>


-- 
Mario A.Montoya
http://www.elmalabarista.com

LogicFan:
> I had a look into how the SQLite decimal extension stores data There a straight-forward string representation is used. So if instead of `serialize`/`deserialize` just `to_string` and `from_str` is there should be compatibility. The SQLite decimal is unbound however.
> 
> I think being able to use the SQLite decimal extensions would be a valuable tradeoff. The string representation also better fits the archiving goals of SQLite compared to the binary representation of `rust_decimal`.

If we use string representation under the hood, how should we deal with the number comparison in where clause? If I recall correctly, the lexical order does not 100% agree with the number order, right?

PR #1067, opened 2021-12-30
Fix markdown rendering of Connection::transaction doc

nicholasbishop:
This fixes `DropBehavior::Commit` being rendered as `DropBehavior: :Commit`.


gwenn:
Thanks.

PR #1069, opened 2022-01-02
Add ToSql/FromSql for [u8; N]

### LINKED ISSUE #536: `params!` does not work with `[u8]`
Neither of following invocation of `params!` compiles:

```rust
extern crate rusqlite;

fn main() {
    let slice = [0u8; 0];
    rusqlite::params![slice];
    rusqlite::params![&slice];
}
```

```
error[E0277]: the trait bound `[u8; 0]: rusqlite::ToSql` is not satisfied
   --> src/main.rs:
    |
    |     rusqlite::params![slice];
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `rusqlite::ToSql` is not implemented for `[u8; 0]`
    |
    = help: the following implementations were found:
              <[u8] as rusqlite::ToSql>
    = note: required for the cast to the object type `dyn rusqlite::ToSql`
```

From reading the documentation (specifically of `ToSql`) it seems like it should work because a `u8` slice should be usable as an sqlite type for blobs.

I guess this is known to not work and is a limitation of the type system or the macro somehow but in this case it could be documented somewhere.

I have worked around the problem with `slice.to_vec()`.

trevyn:
Fixes #536

### DIFF
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -16,21 +16,13 @@ pub enum FromSqlError {
     OutOfRange(i64),
 
     /// Error when the blob result returned by SQLite cannot be stored into the
-    /// requested type due to a size mismatch. (type size, blob size)
-    InvalidSize(usize, usize),
-
-    /// Error returned when reading an `i128` from a
-    /// blob with a size other than 16. Only available when the `i128_blob`
-    /// feature is enabled.
-    #[cfg(feature = "i128_blob")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "i128_blob")))]
-    InvalidI128Size(usize),
-
-    /// Error returned when reading a `uuid` from a blob with
-    /// a size other than 16. Only available when the `uuid` feature is enabled.
-    #[cfg(feature = "uuid")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "uuid")))]
-    InvalidUuidSize(usize),
+    /// requested type due to a size mismatch.
+    InvalidBlobSize {
+        /// The expected size of the blob.
+        expected_size: usize,
+        /// The actual size of the blob that was returned.
+        blob_size: usize,
+    },
 
     /// An error case available for implementors of the [`FromSql`] trait.
     Other(Box<dyn Error + Send + Sync + 'static>),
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - const-generics-tosql
```
gwenn, line 7: Please, revert this change.

trevyn: Of course. Am I missing a trick to get Actions to run on my branch without that?

gwenn:
https://caniuse.rs/features/min_const_generics => MSRV = 1.51, should be ok !

gwenn:
Check if `i128_blob` and `uuid` can reuse this.
https://doc.rust-lang.org/std/primitive.i128.html#method.from_be_bytes
https://doc.rust-lang.org/std/primitive.i128.html#method.to_be_bytes

gwenn:
Just for reference, `rust-postgres` implementations:
https://github.com/sfackler/rust-postgres/blob/master/postgres-types/src/lib.rs#L885-L896
https://github.com/sfackler/rust-postgres/blob/master/postgres-types/src/lib.rs#L535-L566

trevyn:
> Check if `i128_blob` and `uuid` can reuse this.

Yep, it cleans up nicely:

```rust
#[cfg(feature = "i128_blob")]
#[cfg_attr(docsrs, doc(cfg(feature = "i128_blob")))]
impl FromSql for i128 {
    #[inline]
    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
        let bytes = <[u8; 16]>::column_result(value)?;
        Ok(i128::from_be_bytes(bytes) ^ (1i128 << 127))
    }
}

#[cfg(feature = "uuid")]
#[cfg_attr(docsrs, doc(cfg(feature = "uuid")))]
impl FromSql for uuid::Uuid {
    #[inline]
    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
        let bytes = <[u8; 16]>::column_result(value)?;
        Ok(uuid::Uuid::from_u128(u128::from_be_bytes(bytes)))
    }
}
```

This would technically be a breaking change if they now all collapse into the same error type though, let me know if that's OK.

gwenn:
> This would technically be a breaking change if they now all collapse into the same error type though, let me know if that's OK.

I am OK with a breaking change but @thomcc may not.

src/types/to_sql.rs
```diff
@@ -224,6 +224,13 @@ impl ToSql for Vec<u8> {
     }
 }
 
+impl<const N: usize> ToSql for [u8; N] {
```
thomcc, line 227: Huh, we have a `ToSql for [u8]`, and I have expected the `&[u8; N]` to unsize into `&[u8]` and use that one. I guess I can see why the cases in your test might be too hard for the compiler to figure out.

thomcc:
> This would technically be a breaking change if they now all collapse into the same error type though, let me know if that's OK.

I'd be thrilled for us to replace `InvalidI128Size` and `InvalidUuidSize` with the more general `InvalidSize`. Only nit is a minor bikeshed, perhaps we should name it `InvalidBlobSize`?

src/types/from_sql.rs
```diff
@@ -15,6 +15,10 @@ pub enum FromSqlError {
     /// requested type.
     OutOfRange(i64),
 
+    /// Error when the blob result returned by SQLite cannot be stored into the
+    /// requested type due to a size mismatch. (type size, blob size)
+    InvalidSize(usize, usize),
```
thomcc, line 20: As mentioned, I think this should be something like `InvalidBlobSize` (perhaps even naming the fields? They're kind of opaque).

thomcc (commented):
Thrilled to have this! I'd just like the error name to be a little more specific.

If you have the energy, I'd love for you to add a new commit that uses it for the i128 and UUID stuff too, and removes the existing `InvalidI128Size` and `InvalidUuidSize` error variants.

thomcc:
Oh, also, can you squash some of your commits? I'd rather not have the reverted stuff and such that mucks with the CI.

trevyn:
Cool, will do all of this!

trevyn:
Done!

thomcc (approved):
Looks great!

PR #1077, opened 2022-01-05
Clean up unlock_notify code a bit

thomcc:
I was in this file and a couple things bugged me a bit.

- Avoid panicking if something poisons something in the UnlockNotification. This is mostly because we'd ignore the panic anyway — we wrap it in a `catch_unwind` (also, our `catch_unwind` use is a slightly wrong for FFI code, but fixing this is less of a pain).

    This is the main reason I did this, but I noticed the other stuff too while I was in the file.

- Move `#[cfg(feature = "unlock_notify")]` outside `mod unlock_notify`. This means we can get rid of the panicking implementations for when the feature is disabled.

    Also, it makes sure that if we ever need to add any more uses of the code from `mod unlock_notify` we are certain to handle the disabled feature case (admittedly, this is probably a theoretical concern.

- Restructure loop a bit in `RawStatement::step` a little to avoid calling into libsqlite3 more than we need to for the common/success path where we aren't locked.

    The calls this avoids are cheap, but this way enabling `unlock_notify` should have no cost at all unless you have a locking issue.

- Also, I guess nightly clippy got a new lint, so I fixed that too even though it's unrelated.

### DIFF
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -13,7 +13,6 @@ use super::{Connection, InterruptHandle, OpenFlags, Result};
 use crate::error::{error_from_handle, error_from_sqlite_code, Error};
 use crate::raw_statement::RawStatement;
 use crate::statement::Statement;
-use crate::unlock_notify;
 use crate::version::version_number;
 
 pub struct InnerConnection {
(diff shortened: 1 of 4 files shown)


thomcc:
@gwenn: Going to merge this now, but am willing to come back and fix things if you disagree with it.

Issue #738, opened 2020-05-16
Rusqlite need a community.


Voodlaz:
Rusqlite need an Riot(IRC)/Discord/gitter/etc. community where people can ask questions and discuss rusqlite.

thomcc:
I'm not opposed to it. I'm active on some of the rust discord channels, and mozilla riot (altho the latter mostly for work, and on the weekends I generally don't sign in there).

Gitter is probably the easiest for others though.

@gwenn do you have any thoughts here?

gwenn:
Sorry, I have only restricted access to internet where I work.
And I am not active on any of them at home.
So I don't have any opinion.

thomcc:
I've made https://gitter.im/rusqlite/community. I chose gitter just because it doesn't require anybody create another account for a thing, e.g. it uses github oauth.

Maybe it can help coordinate things, or something. Or at least reduce the number of issues we get that are actually about how to use rust or sqlite.

I'll add it to the readme later tonight.

Voodlaz:
I made a pull request adding gitter chat badge to readme.

PR #1079, opened 2022-01-05
Replace Gitter with Discord

thomcc:
So, I don't really like using Gitter, and only use it for Rusqlite. There's a much larger Rust community on Discord, and it's a chat service that I use regularly (and in practice I get more questions about `rusqlite` in various Rust servers than I do on Gitter).

So, I think we should switch away from Gitter, and have made a Discord server. I'm assuming that @gwenn doesn't care very much one way or another (given the comments in https://github.com/rusqlite/rusqlite/issues/738), so this is mostly because it will make my life easier[^1]. Of course, you're more than welcome to join if I've misjudged.

[^1]: I also hope that it will be better for users too — Discord is far more widely used, especially within Rust.

The invite link for the server is here: https://discord.gg/G6Se6Yzw. It's not *fully* set up to do everything I'd like yet[^2], but it's basically there. In the next week or so I'll figure out what to do with the Gitter server (how to retire it, since I don't think mirroring it would be nearly worth the trouble), but for now this is a start.

Also, the badges list had grown quite stale and somewhat misleading, so I cleaned it up while I'm here.

[^2]: Thankfully @khionu (discord wizard and member of the official Rust moderation team) has helped me set it up. They'll also be helping me manage the server, but this doesn't impact this repository in any way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Build Status](https://github.com/rusqlite/rusqlite/workflows/CI/badge.svg)](https://github.com/rusqlite/rusqlite/actions)
 [![dependency status](https://deps.rs/repo/github/rusqlite/rusqlite/status.svg)](https://deps.rs/repo/github/rusqlite/rusqlite)
 [![Latest Version](https://img.shields.io/crates/v/rusqlite.svg)](https://crates.io/crates/rusqlite)
-[![Gitter](https://badges.gitter.im/rusqlite.svg)](https://gitter.im/rusqlite/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
+[![Discord Chat](https://img.shields.io/discord/927966344266256434.svg?logo=discord)](https://discord.gg/G6Se6Yzw)
 [![Docs](https://docs.rs/rusqlite/badge.svg)](https://docs.rs/rusqlite)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,11 @@
 # Rusqlite
 
-[![Travis Build Status](https://api.travis-ci.org/rusqlite/rusqlite.svg?branch=master)](https://travis-ci.org/rusqlite/rusqlite)
-[![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/github/rusqlite/rusqlite?branch=master&svg=true)](https://ci.appveyor.com/project/rusqlite/rusqlite)
-[![Build Status](https://github.com/rusqlite/rusqlite/workflows/CI/badge.svg)](https://github.com/rusqlite/rusqlite/actions)
-[![dependency status](https://deps.rs/repo/github/rusqlite/rusqlite/status.svg)](https://deps.rs/repo/github/rusqlite/rusqlite)
```
gwenn, line 6: Could you please keep the deps status ?

gwenn (changes requested):

README.md
thomcc: Sure. It's inaccurate (we aren't insecure — even if you turn on the `chrono` dep, we disable the feature that is vulnerable...), and I'd prefer that we use something like a GHA check for `cargo audit` or something anyway, but I'll try and make that case another time.

gwenn (approved):

PR #1081, opened 2022-01-05
clippy::let_underscore_drop

gwenn:


thomcc (approved):
Not sure I think this is any better, but I also don't feel that strongly.

PR #1085, opened 2022-01-05
clippy::must_use_candidate

gwenn:


thomcc (approved):
I wonder where else we should be doing this... Seems fine, though.

PR #1089, opened 2022-01-05
clippy::unseparated_literal_suffix

gwenn:


thomcc (commented):
I don't really think this is better, tbh. But I don't really care much.

PR #1090, opened 2022-01-06
Enable buildtime_bindgen tests on windows CI

trevyn:
Noticed that this was a TODO, and I happened to have the incantation to fix it!


thomcc (approved):
Ah! That's very useful!

PR #1096, opened 2022-01-09
Move Limits enum from libsqlite3-sys into rusqlite

thomcc:
This is really not something that belongs in the -sys crate.


gwenn (approved):
Breaking change

thomcc:
> Breaking change

Yes. I have a few other breaking changes for libsqlite3-sys in progress. I'll hold off merging this until they're all ready to avoid doing more major versions bumps than needed.

I'm hoping we can get to a point where we can avoid breaking changes in libsqlite3-sys anymore, unless sqlite itself has a breaking change.

PR #1112, opened 2022-01-15
Fix shell scripts

gwenn:
Replace `$_` by `$0` because `$_` doesn't work when using `sh upgrade.sh`.
Replace `bash` by `sh` because we should not depend on bash.
And also remove unused `CUR_DIR` variable.


gwenn:
#1050 should fix clippy warning

thomcc (approved):

PR #1113, opened 2022-01-16
Uncomment config::DbConfig::SQLITE_DBCONFIG_RESET_DATABASE

link2xt:


link2xt:
I need it for https://github.com/deltachat/deltachat-core-rust/pull/2980

gwenn (approved):

thomcc:
I think we need to detect when this happens and throw cached statements away? I'm concerned this causes issues.

link2xt:
@thomcc In my case I don't have any cached statements at the point when I reset the database. This happens when I just created the tables but then decided I want to import a backup, so I reset the database and import.

I don't think some clever detection will help, because I have a connection pool and if other connections have cached statements they will not be cleared anyway.

There are probably other ways to break cached statements, like deleting and recreating the table etc., I would rather assume someone setting db_config flags knows what they are doing rather than trying to do some magic that will break in even more complex ways.

link2xt:
If I detect that all my cached statement start failing after some operation (resetting the database, migration etc.) I will insert `flush_prepared_statement_cache` explicitly or actually reopen connections.

thomcc:
My concern here is more if there's a way to cause memory unsafety through this.

I don't want to have to add an advisory for unsoundness caused by a niche API that could be replicated in a more obviously safe way (close all connections and delete the DB file or similar, which is how we always did this in firefox).

But I guess I'll try to look into it.

link2xt:
> My concern here is more if there's a way to cause memory unsafety through this.

Will prepared statements segfault if you try to execute them after resetting the database? If they just fail, that's expected. If they segfault, it's a bug in sqlite.

link2xt:
I can also reset the database from another process, even by running `sqlite3` binary and piping the following commands in:
```
.schema
.dbconfig reset_database on
vacuum
.dbconfig reset_database off
```
rusqlite cannot detect this and shouldn't try to.

thomcc (approved):
Fair enough

gwenn:
Should I merge ?

thomcc:
Go for it

Issue #827, opened 2020-10-27
Can't build for wasm target


i-schuetz:
```rust
cargo build --lib --target "wasm32-unknown-unknown"
```

With `rusqlite = {version = "0.24.1", features = ["bundled"]}`:

```
...
cargo:warning=sqlite3/sqlite3.c:14012:10: fatal error: 'stdio.h' file not found
```
```
with `rusqlite = {version = "0.24.1"}`:

```
rust-lld: error: unable to find library -lsqlite3
```

It also happens with 0.23.1.

**Comment by thomcc** [comment]
I think the only wasm target we currently support is wasm32-wasi, although I could be mistaken. @gwenn or @TjeuKayim (who added support for wasm32-wasi) might know.

**Comment by TjeuKayim** [comment]
Compiling Rust that depends on C stdlib to WebAssembly is not possible with `wasm32-unknown-unknown`. So you need to use WASI or Emscripten to compile SQLite and link it to rusqlite. https://github.com/rustwasm/team/issues/291#issuecomment-645482430

**Comment by JimiHend** [comment]
I have the same pb with `cargo build --lib --target "wasm32-wasi"`

**Comment by TjeuKayim** [comment]
Try to install wasi-sdk and set linker environment variables as explained at <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_target/spec/wasm32_wasi/index.html>.

Enable these feature in `Cargo.toml` to use a simple SQLite VFS that is compatible with WASI:

```toml
[dependencies.rusqlite]
features = ["bundled", "wasm32-wasi-vfs"]
```

This is an example for Ubuntu to compile a project depending on rusqlite to WebAssembly:

```sh
sudo apt install libtinfo5
wget https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-11/wasi-sdk_11.0_amd64_ubuntu20.04.deb
sudo dpkg -i wasi-sdk_11.0_amd64_ubuntu20.04.deb
rm wasi-sdk_11.0_amd64_ubuntu20.04.deb
export CC_wasm32_wasi=/opt/wasi-sdk/bin/clang
export CARGO_TARGET_WASM32_WASI_LINKER=/opt/wasi-sdk/bin/clang
RUSTFLAGS="-C target-feature=-crt-static" cargo build --lib --target "wasm32-wasi"
```

i-schuetz:
[...]

PR #1116, opened 2022-01-21
Fix compiler target check for wasm32-wasi

polyrand:
When building with:

```sh
cargo build --lib --target wasm32-wasi --features bundled,wasm32-wasi-vfs
```

The `CARGO_CFG_TARGET_ENV` is an empty string.

The change can be verified by running the command above with and without this change. The build is failing in both cases (#827), but without this change, the [`wasm32-wasi`-specific flags](https://github.com/rusqlite/rusqlite/blob/ddb7141c6dee4b8956af85b2e4a01a28e5fdbacc/libsqlite3-sys/build.rs#L246): `-DSQLITE_OS_OTHER` and `DLONGDOUBLE_TYPE` are not being used by `clang`

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -242,7 +242,7 @@ mod build_bundled {
             cfg.flag("-DHAVE_LOCALTIME_R");
         }
         // Target wasm32-wasi can't compile the default VFS
-        if is_compiler("wasm32-wasi") {
+        if env::var("TARGET") == Ok("wasm32-wasi".to_string()) {
             cfg.flag("-DSQLITE_OS_OTHER")
                 // https://github.com/rust-lang/rust/issues/74393
                 .flag("-DLONGDOUBLE_TYPE=double");


gwenn:
This PR partially rollbacks  #961 related to cross-compilation...

polyrand:
I see. According to [the `target_env` docs](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env):

> For historical reasons, this value is only defined as not the empty-string when actually needed for disambiguation. 

I can see why this is empty when using `wasm32-wasi` as the target. I can't check this PR makes `wasm32-wasi` compilation possible, since it's not working for me for other reasons. But I can verify that `clang` is not being called with the necessary flags and that [the relevant code block](https://github.com/rusqlite/rusqlite/blob/ddb7141c6dee4b8956af85b2e4a01a28e5fdbacc/libsqlite3-sys/build.rs#L245) is being skipped.

gwenn:
> I see. According to [the `target_env` docs](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env):
> 
> > For historical reasons, this value is only defined as not the empty-string when actually needed for disambiguation.
> 
> I can see why this is empty when using `wasm32-wasi` as the target. I can't check this PR makes `wasm32-wasi` compilation possible, since it's not working for me for other reasons. But I can verify that `clang` is not being called with the flags defined in `build.rs` and that [the relevant code block](https://github.com/rusqlite/rusqlite/blob/ddb7141c6dee4b8956af85b2e4a01a28e5fdbacc/libsqlite3-sys/build.rs#L245) is being skipped.

Maybe we should check both ?
```rust
        if env::var("TARGET") == Ok("wasm32-wasi".to_string()) || is_compiler("wasm32-wasi") {
```

libsqlite3-sys/build.rs
```diff
@@ -242,7 +242,7 @@ mod build_bundled {
             cfg.flag("-DHAVE_LOCALTIME_R");
         }
         // Target wasm32-wasi can't compile the default VFS
-        if is_compiler("wasm32-wasi") {
+        if env::var("TARGET") == Ok("wasm32-wasi".to_string()) || is_compiler("wasm32-wasi") {
```
rkusa, line 245: Suggestion: With the following change, `.to_string()` wouldn't be necessary:

```suggestion
        if env::var("TARGET").as_deref() == Ok("wasm32-wasi") || is_compiler("wasm32-wasi") {
```

trevyn:
@rkusa Have you enabled the `wasm32-wasi-vfs` feature? There is a VFS with `sqlite3_os_init` included for wasi, but it's gated behind that feature:

https://github.com/rusqlite/rusqlite/blob/ddb7141c6dee4b8956af85b2e4a01a28e5fdbacc/libsqlite3-sys/build.rs#L244-L252

rkusa:
@trevyn Good call, `wasm32-wasi-vfs` indeed fixes the error and does not require to provide a custom `sqlite3_os_init`. However, in my particular use-case, where I am going to use a custom VFS, and where I want to keep the amount of WASI imports low, I am better of keeping the feature disabled. Because the feature requires the following WASI imports, which are not necessary otherwise:

```
"wasi_snapshot_preview1"."fd_close": [I32] -> [I32]
"wasi_snapshot_preview1"."fd_fdstat_get": [I32, I32] -> [I32]
"wasi_snapshot_preview1"."fd_filestat_get": [I32, I32] -> [I32]
"wasi_snapshot_preview1"."fd_prestat_get": [I32, I32] -> [I32]
"wasi_snapshot_preview1"."fd_prestat_dir_name": [I32, I32, I32] -> [I32]
"wasi_snapshot_preview1"."fd_read": [I32, I32, I32, I32] -> [I32]
"wasi_snapshot_preview1"."fd_seek": [I32, I64, I32, I32] -> [I32]
"wasi_snapshot_preview1"."fd_sync": [I32] -> [I32]
"wasi_snapshot_preview1"."path_filestat_get": [I32, I32, I32, I32, I32] -> [I32]
"wasi_snapshot_preview1"."path_open": [I32, I32, I32, I32, I32, I64, I64, I32, I32] -> [I32]
"wasi_snapshot_preview1"."path_unlink_file": [I32, I32, I32] -> [I32]
```

Again, this is for my personal use-case.

xpepermint:
Status?

thomcc:
I'll have a fix for this and a bunch of other issues in the build script later this weekend (ended up rewriting the build.rs to fix a ton of bugs, took longer than expected but basically all thats left is a bit more testing).

Not sure why I didn't merge this earlier, but at this point it would just conflict.

xpepermint:
Ow, that sounds perfect!

dennismartensson:
@thomcc looks like this is still an issue in main are you still planing on doing the other build.rs fixes and this in one go?

PR #1129, opened 2022-02-27
Prepare release 0.27.0

gwenn:


thomcc (approved):

thomcc:
I was hoping to do some other stuff for the breaking libsqlite3-sys change, but this release has several SQLite features people want (and was requested) so we shouldn't block it, since I won't get to that stuff for at least another week.

PR #1133, opened 2022-03-07
Implement `Params` for tuples, improve documentation some.

thomcc:
Someone pointed out to me that rusqlite should support passing tuples in, which would make `params!` mostly redundant. So I did that.

This also improves the README:

1. Remove the comment about `rust-postgres`. We're not very close to their API at all anymore, and haven't been for a long time, so it's just misleading.

2. Point people towards bundled. The number 1 question I see about rusqlite is about build issues. In practice, most folks are better off using bundled unless they're building a generic library. I don't think it should be on by default though, since default features are very hard to turn off.

This probably should be a separate PR, but it came from the same feedback I got, and I was already there.

@gwenn do you agree?

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,45 +88,34 @@ newer SQLite version; see details below.
 Rusqlite provides several features that are behind [Cargo
 features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section). They are:
 
-* [`load_extension`](https://docs.rs/rusqlite/latest/rusqlite/struct.LoadExtensionGuard.html)
+* [`load_extension`](https://docs.rs/rusqlite/~0/rusqlite/struct.LoadExtensionGuard.html)
   allows loading dynamic library-based SQLite extensions.
-
-* [`backup`](https://docs.rs/rusqlite/latest/rusqlite/backup/index.html)
+* [`backup`](https://docs.rs/rusqlite/~0/rusqlite/backup/index.html)
   allows use of SQLite's online backup API. Note: This feature requires SQLite 3.6.11 or later.
-
-* [`functions`](https://docs.rs/rusqlite/latest/rusqlite/functions/index.html)
+* [`functions`](https://docs.rs/rusqlite/~0/rusqlite/functions/index.html)
   allows you to load Rust closures into SQLite connections for use in queries.
   Note: This feature requires SQLite 3.7.3 or later.
-
 * `window` for [window function](https://www.sqlite.org/windowfunctions.html) support (`fun(...) OVER ...`). (Implies `functions`.)
-
-* [`trace`](https://docs.rs/rusqlite/latest/rusqlite/trace/index.html)
+* [`trace`](https://docs.rs/rusqlite/~0/rusqlite/trace/index.html)
   allows hooks into SQLite's tracing and profiling APIs. Note: This feature
   requires SQLite 3.6.23 or later.
-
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -8,11 +8,30 @@
 [![Dependency Status](https://deps.rs/repo/github/rusqlite/rusqlite/status.svg)](https://deps.rs/repo/github/rusqlite/rusqlite)
 [![Discord Chat](https://img.shields.io/discord/927966344266256434.svg?logo=discord)](https://discord.gg/nFYfGPB8g4)
 
-Rusqlite is an ergonomic wrapper for using SQLite from Rust. It attempts to expose
-an interface similar to [rust-postgres](https://github.com/sfackler/rust-postgres).
```
gwenn, line 12: At least we can still stay that `rusqlite` API has been inspired from `rust-postgres` (like `ToSql` / `FromSql`). And, for me, this link is convenient when someone challenges `rusqlite` API (and when `rust-postgres` does the same).
Also, by analogy, https://www.pgcon.org/2014/schedule/events/736.en.html:
> SQLite can be thought of as a derivative of PostgreSQL. SQLite was originally written from PostgreSQL 6.5 documentation, and the SQLite developers still use PostgreSQL as a reference platform to verify that SQLite is working correctly.

README.md
```diff
@@ -8,11 +8,30 @@
...
+[dependencies]
+# `bundled` causes us to automatically compile and link in an up to date
+# version of SQLite for you. This avoids many common build issues, and
+# avoids depending on the version of SQLite on the users system (or your
+# system), which may be old or missing. It's the right choice for most
+# programs that control their own SQLite databases.
+#
+# That said, it's not ideal for all scenarios and in particular, generic
+# libraries built around `rusqlite` should probably not enable it, which
+# is why it is not a default feature -- it could become hard to disable.
```
gwenn, line 27: Perfect explanation !

README.md
thomcc: Hm, fair enough. In practice ISTM we're pretty far away already, and both of us are moving further apart...  I'll see if I can think of a better way to word it that reflects the history without implying anything more.

I want to avoid implying there's any compatibility or intent of compatibility or similarity between the two libraries. My hesitation is mostly that some users may see this and expect/hope that they will be able to use the similarities to enable use of SQLite for development and postgres for production, which will... cause them frustration if they try to do so with `rusqlite` and `rust-postgres`. IMO that situation is not good for anybody[^1], and really, I think it'd end up being very frustrating for the user, so I'd rather not encourage it at all.

Honestly users who want the compatibility will be way better served by using neither, and instead finding a database abstraction library that actually *intends* to provide compatibility across DBs (rather than hacking it together via clever `#[cfg()]`s or whatever), since there are several decent ones that do at this point.

[^1]: Well, for reasons I can't really place I feel like `rusqlite` fairs worse in that situation. Perhaps because we're not the one that is used in production, but maybe just because we're the one making the claim of similarity.

README.md
thomcc: I wish `cargo` had some notion for features that are really meant to be controlled by the application (or even the person performing the build, like a packager or user running `cargo install`). Really feels lacking for code that uses C libraries...

gwenn (approved):
Thanks

thomcc:
Wow, you're fast. I was about to say:

> I'm going to land this once tests pass since I want to change some other stuff in the repo, but LMK if you think working needs to be adjusted further, I'm happy to do so

But you beat me to it.

PR #1137, opened 2022-03-13
Introduce SqlInputError with offset

### LINKED ISSUE #1130: Error offsets
Are there any plans to include the results of the new `sqlite3_error_offset` into rusqlite's errors? I am working on an LSP server for sqlite and it would be useful for diagnostics.

gwenn:
Should fix #1130

- [ ] `sqlite3_declare_vtab` ?
- [ ] offset specified only when code is [SQLITE_ERROR](https://sqlite.org/rescode.html#error) ?
- [X] [Batch](https://github.com/rusqlite/rusqlite/blob/master/src/lib.rs#L958) => offset relative to `SqlInputError.sql` (not `Batch.sql`)
- [X] [execute_batch](https://github.com/rusqlite/rusqlite/blob/master/src/lib.rs#L480) => offset relative to `SqlInputError.sql` (not `execute_batch` sql parameter)
- [ ] `modern_sqlite` feature, I know...


thomcc (approved):
Looks fine to me. Anything else you want to do with it still?

PR #1138, opened 2022-03-15
README: update a sample code according to lib.rs

dondakeshimo:
Hello.

README is outdated about a sample code, and it doesn't work now.
I revised README according to [a documentation in `lib.rs`](https://github.com/dondakeshimo/rusqlite/blob/49d9556625edecee6cafc99054a53b891d99f951/src/lib.rs#L4-L49) .

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ rusqlite = { version = "0.27.0", features = ["bundled"] }
 Simple example usage:
 
 ```rust
-use rusqlite::{Connection, Result};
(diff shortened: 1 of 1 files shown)


dondakeshimo:
Thank you for you quick reply.
I understand that It is just a bad timing to try a sample code in github now, and the code is not outdated.

I'll close this PR.
Thank you.

thomcc:
Sorry about that!

Issue #900, opened 2021-02-14
Unable to create any sort of writeable Module


LikeLakers2:
Hi, I was looking at the virtual tables module of rusqlite. While the tools provided seem very useful, everything within the module seems to only serve the purposes of making a read-only module -- writeable modules are out of the question unless I directly use the FFI interface.

Are there any plans to allow read-and-write modules? Or is rusqlite staying with just read-only modules for the time being?

gwenn:
As far as I know, there is no plan to implement writable vtab.
And we still miss some features on read-only vtab (see #639).

gwenn:
Version 0.28.0 released.

PR #1141, opened 2022-03-17
Add support to updatable virtual tables

### LINKED ISSUE #730: Why is xUpdate None?
Is there a particular reason why vtab [`xUpdate` is `None`](https://github.com/rusqlite/rusqlite/blob/master/src/vtab/mod.rs#L109) other than there simply not being a trait for it?

gwenn:
Fix #730, #900

TODO
[X] simple implementation at least for test / example.
[X] eponymous updatable VTab

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -62,11 +62,15 @@ pub enum VTabKind {
     /// Non-eponymous
     Default,
     /// [`create`](CreateVTab::create) == [`connect`](VTab::connect)
+    ///
+    /// See [SQLite doc](https://sqlite.org/vtab.html#eponymous_virtual_tables)
     Eponymous,
     /// No [`create`](CreateVTab::create) / [`destroy`](CreateVTab::destroy) or
     /// not used
     ///
     /// SQLite >= 3.9.0
+    ///
+    /// See [SQLite doc](https://sqlite.org/vtab.html#eponymous_only_virtual_tables)
     EponymousOnly,
 }
 
@@ -835,6 +839,13 @@ impl InnerConnection {
         module: &'static Module<'vtab, T>,
         aux: Option<T::Aux>,
     ) -> Result<()> {
+        use crate::version;
+        if version::version_number() < 3_009_000 && module.base.xCreate.is_none() {
+            return Err(Error::ModuleError(format!(
+                "Eponymous-only virtual table not supported by SQLite version {}",
+                version::version()
+            )));
+        }
         let c_name = str_to_cstring(module_name)?;
         let r = match aux {
             Some(aux) => {
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -84,6 +84,43 @@ const ZERO_MODULE: ffi::sqlite3_module = unsafe {
     .module
 };
 
+/// Create an modifiable virtual table implementation.
+///
+/// Step 2 of [Creating New Virtual Table Implementations](https://sqlite.org/vtab.html#creating_new_virtual_table_implementations).
+#[must_use]
+pub fn update_module<'vtab, T: UpdateVTab<'vtab>>() -> &'static Module<'vtab, T> {
+    #[allow(clippy::needless_update)]
+    &Module {
+        base: ffi::sqlite3_module {
+            iVersion: 2,
+            xCreate: Some(rust_create::<T>),
+            xConnect: Some(rust_connect::<T>),
+            xBestIndex: Some(rust_best_index::<T>),
+            xDisconnect: Some(rust_disconnect::<T>),
+            xDestroy: Some(rust_destroy::<T>),
+            xOpen: Some(rust_open::<T>),
+            xClose: Some(rust_close::<T::Cursor>),
+            xFilter: Some(rust_filter::<T::Cursor>),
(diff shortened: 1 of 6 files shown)


gwenn:
I don't understand why 6d885c8 doesn't compile.
Maybe because of https://github.com/rust-lang/rust/issues/60502 ?

thomcc:
We might have to explicitly put it as a const or something, and return a reference to that. Const promotion doesn't work that reliably.

It's possible that keeping this const isn't viable also.

src/vtab/mod.rs
```diff
@@ -57,6 +57,19 @@ use crate::{str_to_cstring, Connection, Error, InnerConnection, Result};
...
+pub enum VTabKind {
+    /// Non-eponymous
+    Default,
+    /// [`create`](CreateVTab::create) == [`connect`](VTab::connect)
+    Eponymous,
+    /// No [`create`](CreateVTab::create) / [`destroy`](CreateVTab::destroy) or
+    /// not used
+    ///
+    /// SQLite >= 3.9.0
+    EponymousOnly,
```
thomcc, line 70: We need to add a check for `crate::version_number() >= 3009000` for attempts to register an `EponymousOnly` module, and return an `Err(...)`.

Prior to 3.9.0 SQLite will crash on eponymous-only modules, because it doesn't check for `xCreate` being NULL.

src/vtab/mod.rs
```diff
@@ -57,6 +57,19 @@ use crate::{str_to_cstring, Connection, Error, InnerConnection, Result};
 // ffi::sqlite3_vtab => VTab
 // ffi::sqlite3_vtab_cursor => VTabCursor
 
+/// Virtual table kind
```
thomcc, line 60: I think this needs better docs. Can you use this? Do a quick read-through of it to make sure I didn't make any obvious mistakes, also.

```rs
/// Virtual table kind, indicating whether or not the table is
/// [eponymous](https://sqlite.org/vtab.html#eponymous_virtual_tables).
#[non_exhaustive]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum VTabKind {
    /// Non-eponymous tables, which are the most flexible.
    Default,
    /// Eponymous tables exist implicitly in the main schema of the database,
    /// and do not require setup such as `CREATE VIRTUAL TABLE` to be used.
    ///
    /// These should implement [`CreateVTab::create`] and [`VTab::connect`]
    /// identically. (Which is the default behavior if [`CreateVTab::create`] is
    /// not explicitly overridden)
    Eponymous,
    /// Eponymous-only tables are tables which may only be used eponymously, and
    /// cannot be used with `CREATE VIRTUAL TABLE`.
    ///
    /// This is often useful for table valued functions, although note that this
    /// prevents use as a temporary table, in patterns like the following:
    ///
    /// ```sql
    /// CREATE VIRTUAL TABLE temp.something USING my_vtab(args, here);
    /// ```
    ///
    /// But whether or not this pattern makes sense to support depends on the
    /// implementation of your table.
    ///
    /// Eponymous-only tables will not have the [`CreateVTab::create`] and
    /// [`CreateVTab::destroy`] methods called. [`VTab::connect`] will be used
    /// instead (and cleanup will be performed by [`Drop::drop`]).
    ///
    /// If the SQLite version is older than 3.9.0, `rusqlite` will produce an
    /// error when you attempt to register a `EponymousOnly` virtual table
    EponymousOnly,
}
```

thomcc (commented):
This looks good but needs a couple changes.

src/vtab/vtablog.rs
```diff
@@ -0,0 +1,300 @@
...
+
+impl VTabLog {
+    fn connect_create(
+        _: &mut VTabConnection,
+        _: Option<&()>,
+        args: &[&[u8]],
+        is_create: bool,
+    ) -> Result<(String, VTabLog)> {
+        static N_INST: AtomicUsize = AtomicUsize::new(1);
+        let i_inst = N_INST.fetch_add(1, Ordering::SeqCst);
```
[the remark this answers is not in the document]
thomcc: Actually, this doesn't matter since it's test only.

src/vtab/mod.rs
gwenn: https://sqlite.org/vtab.html#eponymous_only_virtual_tables
> Note that prior to [version 3.9.0](https://sqlite.org/releaselog/3_9_0.html) (2015-10-14), SQLite did not check the xCreate method for NULL before invoking it. So if an eponymous-only virtual table is registered with SQLite [version 3.8.11.1](https://sqlite.org/releaselog/3_8_11_1.html) (2015-07-29) or earlier and a [CREATE VIRTUAL TABLE](https://sqlite.org/lang_createvtab.html) command is attempted against that virtual table module, a jump to a NULL pointer will occur, resulting in a crash.

A crash seems safe to me.

thomcc: Crash from null pointer dereference is undefined behavior. Absolutely not safe. Just because it is deterministic and is not exploitable does not mean it's memory safe.

gwenn: Fixed

thomcc (approved):
LGTM. i'll have a follow-up PR up for the unwinding shortly.

PR #1142, opened 2022-03-18
Updated simple example for current release

YesSeri:
The simple example in the readme file didn't work with the latest release. I don't know when it broke, but I copied the version from [here](https://docs.rs/rusqlite/latest/rusqlite/) to github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ rusqlite = { version = "0.27.0", features = ["bundled"] }
 Simple example usage:
 
 ```rust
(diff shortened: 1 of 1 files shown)


thomcc:
It will work in the next release, and works in the version currently on master.

thomcc:
Closing, there's no need to make this change. The version on https://crates.io/crates/rusqlite is correct for the latest release, the version on github is correct for not-yet-released code.

Perhaps it shouldn't have been changed in advance, but at this point there's no reason to undo the change.

PR #1149, opened 2022-03-30
Prep libsqlite3-sys release 0.24.2

thomcc:
Requested in https://github.com/mozilla/application-services/issues/4893.


thomcc:
We should actually cut a release of rusqlite too, since https://crates.io/crates/libsqlite3-sys shows the new (unreleased) tuple syntax. I probably won't have time to do this until tomorrow after work though.

PR #1151, opened 2022-04-02
Disable winsqlite3 on 32 bit targets

thomcc:
`winsqlite3.dll` actually has a totally custom ABI on 32-bit targets which is different than SQLite typically uses on those platforms (and on any other -- literally this is the only time that SQLite uses an ABI other than `extern "C"`, as far as I can tell -- it even would have required modification to SQLite itself (and its header) to do so).

This is what it the bindgen output looks for winsqlite3: https://gist.github.com/thomcc/4bbbecd0bd79e5ed181fcc9960c59e77. Note that this applies both to the exported functions, *and* to callbacks, which makes it very hard for us to support. (On 64 bit targets, the ABI is the same as everywhere else -- `extern "C"`).

In practice, the feature currently doesn't actually work at all on these targets, since we try to pass `extern "C" fn` as callbacks to functions that expect `extern "stdcall"`, which gives a strange compile error. This makes that explicit, by emitting a compile error on these targets, rather than a mysterious error.

Eventually it could be possible to support this, but I think it might require some new Rust features, such as allowing a macro after `extern` (as in `extern abi!() { ... }` and `extern abi!() fn`, so that we could swap `extern "stdcall"` and `extern "C"`)... but this would still be pretty unfortunate, as we'd have a feature that would break code when enabled, which is generally bad form. We could emit wrappers (for non-varargs funcs), but that's a big maintenance burden.

It's very tempting to me to say that we should remove this feature, as it has been the cause of several headaches now, and people should almost certainly just use bundled instead... But I guess it's not really that bad to keep just for 64 bit targets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,7 @@ bundled-windows = ["libsqlite3-sys/bundled-windows"]
 with-asan = ["libsqlite3-sys/with-asan"]
 column_decltype = []
 wasm32-wasi-vfs = ["libsqlite3-sys/wasm32-wasi-vfs"]
+# Note: doesn't support 32-bit.
 winsqlite3 = ["libsqlite3-sys/winsqlite3"]
 
 # Helper feature for enabling most non-build-related optional features
--- a/libsqlite3-sys/Cargo.toml
+++ b/libsqlite3-sys/Cargo.toml
@@ -35,8 +35,13 @@ session = ["preupdate_hook", "buildtime_bindgen"]
 in_gecko = []
 with-asan = []
 wasm32-wasi-vfs = []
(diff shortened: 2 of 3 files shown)


thomcc:
> In practice, the feature currently doesn't actually work at all on these targets, but this makes that explicit, by emitting a compile error for them.

Note that if it did work, it would probably crash on the first call into SQLite. (At least, without some mitigation to address ABI mismatch).

bdbai:
Hi @thomcc , just wondering if [`extern "system"`](https://doc.rust-lang.org/reference/items/external-blocks.html#abi) will work out of the box?

PR #1153, opened 2022-04-03
Fix non-bundled tests against macOS system SQLite

thomcc:
On macOS we fail some tests when they're run against system SQLite. These may be bugs in the macOS system SQLite (since `PRAGMA compile_options` says that `BUG_COMPATIBLE_20160819` is enabled, which seems a little ominous), but the fixes for these are easy enough.

I only bothered checking that the mode behaves as expected for bundled in `test_pragma_query_row` and not in `pragma_update_and_check`, since the latter seems to be mostly about testing that compilation doesn't fail.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1515,15 +1515,28 @@ mod test {
     #[test]
     fn test_pragma_query_row() -> Result<()> {
         let db = Connection::open_in_memory()?;
-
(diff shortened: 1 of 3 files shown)


thomcc:
I'm just going to land this after CI is green since it's trivial and is bothering me right now.

PR #1154, opened 2022-04-03
Rename enum variants to Rust-friendly names

### LINKED ISSUE #1095: Rename enum variants to Rust-friendly names
We have a lot of enumerations where the variants have names that are identical to the FFI names, which is bad style and looks weird.

However, the SQLite FFI names are still pretty important, and it would be a mistake for us not to clearly explain what each variant is. These are how the official docs will refer to these values, and it will help folks figure out what SQLite docs to read to understand the value, *and* what variant the official docs are referring to when they reference some variant.

That is, for a given enumeration, users should to easily be able to both: determine the Rust name for a given FFI name, and vice versa. (It also would be an enormous breaking change if we change all of this at once without some consideration for easy migration path)

To solve these issues, I propose that we do the following steps:

1. Rename the variants to use CamelCase names.
2. Reference the FFI names in the documentation of the variants.
3. Add inherent constants with the `SQLITE_` to both allow people to find them, and significantly reduce the breakage this change would cause.

I don't mind doing this, but @gwenn, do you have strong objections to this?

## Example

For example, on `Limit` steps 1 and 2 would look like:

```rs
// ... attributes/docs omitted ...
pub enum Limit {
    /// The maximum size of any string or BLOB or table row, in bytes.
    ///
    /// Equivalent to `SQLITE_LIMIT_LENGTH` from the C API.
    Length = ffi::SQLITE_LIMIT_LENGTH,
    /// The maximum length of an SQL statement, in bytes.
    ///
    /// Equivalent to `SQLITE_LIMIT_SQL_LENGTH` from the C API.
    SqlLength = ffi::SQLITE_LIMIT_SQL_LENGTH,
    /// The maximum number of columns in a table definition or in the result set
    /// of a SELECT or the maximum number of columns in an index or in an
    /// ORDER BY or GROUP BY clause.
    ///
    /// Equivalent to `SQLITE_LIMIT_COLUMN` from the C API.
    Columns = ffi::SQLITE_LIMIT_COLUMN,
	// ... rest of the limits ...
}
```

And then step 3 would be providing the following:

```rs
impl Limit {
    /// An alias for [`Limit::Length`], provided for compatibility with the C API.
    pub const SQLITE_LIMIT_LENGTH: Self = Self::Length;
    /// An alias for [`Limit::SqlLength`], provided for compatibility with the C API.
    pub const SQLITE_LIMIT_SQL_LENGTH: Self = Self::SqlLength;
    /// An alias for [`Limit::Columns`], provided for compatibility with the C API.
    pub const SQLITE_LIMIT_COLUMN: Self = Self::Columns;
	// ... rest of the limits ...
}
```

thomcc:
Fixes #1095.

I think this is technically breaking, because you can `use` an enum's variant to bring it into scope, but this is not allowed for constants. So code that does this would break:

```rs
use rusqlite::DbConfig::SQLITE_DBCONFIG_ENABLE_FKEY;
conn.db_config(SQLITE_DBCONFIG_ENABLE_FKEY)
```

~~Short of that, other cases should not be broken.~~ EDIT: This was not possible for `IndexConstraintOp`, since it had the `SQLITE_INDEX_CONSTRAINT_FUNCTION` variant. I doubt many people use this, because we don't expose `xFindFunction`, so I just renamed it.

Other notes:

1. Cases which don't correspond to SQLite's constants (for example, `UNKNOWN` variants) are present as associated constants, but have had the capitalized versions deprecated, since the "compatibility with SQLite C API" argument doesn't hold. Some of these probably also should be `TryFrom` impls rather than `From`.

2. Some DbConfig values were gated on `feature = "modern_sqlite"`. This has been removed, in favor of noting that they will fail if you try to use it in an unsupported version. This is useful if you want to toggle the feature if it's supported, but do nothing if not. For example, it seems reasonable to allow code like the following:

    ```rs
    if let Err(e) = conn.set_db_config(rusqlite::DbConfig::Defensive, true) {
        // Might be in an old version of SQLite, continue anyway...
        log::warn!("Failed to enable defensive mode: {:?}", e);
    }
    ```

3. I only bothered with the value tests for cases where there were a large number of variants.

4. In several cases, I also tried to improve the docs.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,65 +6,203 @@ use crate::error::check;
 use crate::ffi;
 use crate::{Connection, Result};
 
-/// Database Connection Configuration Options
-/// See [Database Connection Configuration Options](https://sqlite.org/c3ref/c_dbconfig_enable_fkey.html) for details.
+/// Per-Connection databse configuration options.
+///
+/// See [Database Connection Configuration Options][dbconfig] for details.
+///
+/// [dbconfig]: https://sqlite.org/c3ref/c_dbconfig_enable_fkey.html
 #[repr(i32)]
-#[allow(non_snake_case, non_camel_case_types)]
 #[non_exhaustive]
-#[allow(clippy::upper_case_acronyms)]
 pub enum DbConfig {
-    //SQLITE_DBCONFIG_MAINDBNAME = 1000, /* const char* */
-    //SQLITE_DBCONFIG_LOOKASIDE = 1001,  /* void* int int */
+    //SQLITE_DBCONFIG_MAINDBNAME = 1000,
+    //SQLITE_DBCONFIG_LOOKASIDE = 1001,
     /// Enable or disable the enforcement of foreign key constraints.
-    SQLITE_DBCONFIG_ENABLE_FKEY = 1002,
+    ///
+    /// Note that by default these are off in most versions of SQLite, although
+    /// if you are using `features = "bundled"`, they will be on by default
+    /// (because the bundled build is performed with the compiler flag
+    /// `-DSQLITE_DEFAULT_FOREIGN_KEYS=1`).
+    ///
+    /// It is frequently useful to temporarially disable this during migrations.
+    ///
+    /// Equivalent to `SQLITE_DBCONFIG_ENABLE_FKEY` from the C API.
+    ForeignKeysEnabled = 1002,
+
     /// Enable or disable triggers.
-    SQLITE_DBCONFIG_ENABLE_TRIGGER = 1003,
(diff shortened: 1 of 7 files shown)


gwenn (approved):
You should have drop old entries.

thomcc:
> You should have drop old entries.

Hm, you mean I should delete the old variants?

If you feel that way, I'll deprecate and #[doc(hidden)] them. After a couple versions we can delete them. I don't want people to dread updating `rusqlite`, and deleting them all in one go would break very many users.

PR #1155, opened 2022-04-04
Improve the docs for opening the connection a bit

thomcc:
I got some questions about use of `NO_MUTEX` by default. I've expanded the docs and fixed some issues with them.

I also updated the note about rust-postgres so that it matches the README.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,9 @@
-//! Rusqlite is an ergonomic wrapper for using SQLite from Rust. It attempts to
-//! expose an interface similar to [rust-postgres](https://github.com/sfackler/rust-postgres).
+//! Rusqlite is an ergonomic wrapper for using SQLite from Rust.
+//!
+//! Historically, the API was based on the one from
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -346,27 +350,59 @@ impl Drop for Connection {
...
+    ///
+    ///     Rusqlite enforces thread-safety at compile time, so additional
+    ///     locking is not needed and provides no benefit. (See the
+    ///     documentation on [`OpenFlags::SQLITE_OPEN_FULL_MUTEX`] for some
+    ///     additional discussion about this).
+    ///
+    /// Most of these are also the default settings for the C API, although
+    /// technically the default locking behavior is controlled by the flags used
+    /// when compiling SQLite -- rather than let it vary, we choose `NO_MUTEX`
+    /// because it's a fairly clear best choice for users this library.
```
gwenn, line 400: _of_ this library

gwenn (approved):

src/lib.rs
thomcc: Crap, missed this comment before clicking land. Will fix.

PR #1160, opened 2022-04-07
Make possible to specify subtype of SQL function

### LINKED ISSUE #1157: Unable to call set_result_subtype at the appropriate time
I see that you have added a binding for `sqlite3_result_subtype`, which is great, that's exactly the thing I need! However, this needs be called /after/ `sqlite3_result_*` in order to take effect, which isn't currently possible. 

Do you have something in mind for how to do it? I'd be happy to implement it if it's reasonably straightforward.

gwenn:
Fix #1157 
Breaking changes.


gwenn:
I would like to do the same C impl:
https://github.com/sqlite/sqlite/blob/2e12890732f8c23719a46e56175323cf7d1daccb/src/test_func.c#L662
in Rust:
https://github.com/rusqlite/rusqlite/pull/1160/files#diff-676f1190ad1b28e57e9eac75de61cccc59420615e3da10f91a493fb48d0711ffR1098
but there is a lifetime issue.
Could you please help ?
Thanks.

PR #1164, opened 2022-04-17
Add `Connection::release_memory` method

nvzqz:
This is behind a new `release_memory` feature flag. It simply calls [`sqlite3_db_release_memory`](https://www.sqlite.org/c3ref/db_release_memory.html) on the raw handle.

Open questions:
- How should we expose [`sqlite3_release_memory`](https://www.sqlite.org/c3ref/release_memory.html)?
- Can we add `#[doc(alias = "sqlite3_db_release_memory")]`? It would require at least Rust 1.48. This can perhaps be its own issue/PR.
- Should types with exclusive access to `Connection` expose this operation? Or expose the underlying connection reference?


thomcc (approved):
LGTM

PR #1165, opened 2022-04-17
derive Eq for Type

maxwell8888:


thomcc (approved):
There are probably a few more we could derive here, hm.

PR #1215, opened 2022-08-17
Revert part of "Drop bindgen layout tests"

dkg:
This reverts parts of commit 53b1b598cb5ca63befa043a05905b681e8890dc7.

@thomcc observed that more was cut out from the shipped
bindgen_bundled_version.rs than should have been.

I've reverted the change to re-generated code (which it looks like i
did not regenerate as expected), and instead manually trimmed out all
the *_test_layout_* functions marked as #![test].

Note that {sqlite3,sqlcipher}/bindgen_bundled_version.rs both have
comments claiming automated generation, but now they are in fact a
mixture of automatic generation and manual tuning.


thomcc:
Thanks. I'll give this a more in-depth look for why this is needed later.

gwenn:
See https://github.com/rusqlite/rusqlite/blob/4d8b8e71cd42fb58c2845e2d09a7075972e561e9/libsqlite3-sys/build.rs#L618

PR #1221, opened 2022-08-21
Change default minimal SQLite API version

gwenn:
From 3.6.8 to 3.14.0.
Use `old_sqlite` feature to keep 3.6.8 (or 3.7.16) as the minimal version.
Use `modern_sqlite` for SQLite API > 3.14.0.
Also remove old 3.6.23 and 3.7.7 bindings.


thomcc:
I think there's an argument we shouldn't support these at all, as they have tons of security vulnerabilities. There's even an argument that under these our APIs aren't safe.

gwenn:
> I think there's an argument we shouldn't support these at all, as they have tons of security vulnerabilities. There's even an argument that under these our APIs aren't safe.

But in this case what should be the minimal version supported by rusqlite ?
The last available version ?

thomcc:
> But in this case what should be the minimal version supported by rusqlite ? The last available version ?

Yeah I'm not sure. I think that's probably too aggressive. It's a tricky problem... We probably should not have an `old_sqlite` feature though, would be my hunch.

gwenn:
> > But in this case what should be the minimal version supported by rusqlite ? The last available version ?
> 
> Yeah I'm not sure. I think that's probably too aggressive. It's a tricky problem... We probably should not have an `old_sqlite` feature though, would be my hunch.

Ok, il will try to remove `old_sqlite` feature but I guess we will have the same issue with `winsqlite3` (< 3.14)...

thomcc:
`winsqlite3` is a mess. It only works now on targets where `extern "C"` and `extern "system"` are the same. In retrospect, we probably shouldn't have added it since statically linking SQLite only adds ~1.0-1.5MB (for comparison, the Rust stdlib is ~7MB).

But yeah, idk. I think it's reasonable for us to cut off old versions at some point. Otherwise we have to support this stuff forever, no matter how broken and dangerous to use it is.

gwenn:
I was wrong: `winsqlite3` seems not impacted which means >= 3.14.0 on the VM or I missed something.

gwenn:
I chose version 3.14.0 but you may have a better choice ?

thomcc:
Seems fine for now, eventually libsqlite3-sys should be mostly agnostic to the version (and then i'd feel better about being more aggressive about this in rusqlite).

PR #1224, opened 2022-09-12
Upgrade SQLite bundled version to 3.39.3

gwenn:


gwenn:
@dkg, @thomcc if you look at these [differences](https://github.com/rusqlite/rusqlite/pull/1224/files#diff-2ce4120470728e5c0a4cae79b24a0ba49e06eb4c5520d63ec1c8f0fe83df03b3), you can see that something wrong happened with PR #1213. Because there should be almost no difference for bindings between 3.39.2 and 3.39.3...

PR #1230, opened 2022-09-30
Respect missing values in session's conflict resolver

xpepermint:
When applying a `changeset` (e.g. for `UPDATE`), the conflict resolver’s `item` may include "missing" values since only changed columns are provided. Trying to read unexisting columns will cause the conflict resolver to panic.

The current code properly handles `SqliteFailure` related errors (e.g. `ParameterOutOfRange`) but will panic for operations performed on valid but at the same time missing values.

This small PR adds checks in `ChangesetItem` object that eliminate these issues and thus properly handle the `Result<ValueRef>` by throwing `InvalidColumnIndex` when a column is missing.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -11,7 +11,7 @@ use std::slice::{from_raw_parts, from_raw_parts_mut};
 
 use fallible_streaming_iterator::FallibleStreamingIterator;
 
-use crate::error::{check, error_from_sqlite_code};
+use crate::error::{check, error_from_sqlite_code, Error};
 use crate::ffi;
 use crate::hooks::Action;
 use crate::types::ValueRef;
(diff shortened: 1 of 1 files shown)


gwenn:
https://sqlite.org/session/sqlite3changeset_new.html
> If the change is an UPDATE and does not include a new value for the requested column, *ppValue is set to NULL and SQLITE_OK returned.

So only `sqlite3changeset_new` should be affected, no ?

xpepermint:
I believe so ...

gwenn:
> I believe so ...

I mean `sqlite3changeset_old` and `sqlite3changeset_conflict` should **not** be affected. But in your PR they are !

xpepermint:
A precaution `.is_null` check should be there anyway.

gwenn:
I checked again SQLite source code.
And `sqlite3changeset_conflict` cannot return a null value.
But `sqlite3changeset_old` can:
https://github.com/sqlite/sqlite/blob/99ace8263426ee9001d3994e6a8275d4fd8b8fc3/ext/session/sqlite3session.c#L3513-L3516
> Or, if that particular value is not included in the record (because the change is an UPDATE and the field was not modified and is not a PK column), set *ppValue to NULL.

gwenn:
Sorry for the delay...

PR #1252, opened 2022-11-11
Fix FromSql/ToSql for serde_json::Value

### LINKED ISSUE #882: Support for Number (Integer/Float) and Null in serde_json
I was trying to serialize query result into a serde_json Value inside `query_map` but I got an InvalidColumn error, which was a bit confusing because README.md claims that serde_json is supported (via cargo optional feature). 

After Looking around I found that the [serde_json.rs](https://github.com/rusqlite/rusqlite/blob/master/src/types/serde_json.rs) has FromSQL implementation only for `Text` and `Blob` types, and thus the error.

Is this by design or a mistake?

if a bug then I'd love to try and extend support for other serde_json types.

I am trying to implement a function in my application that would execute arbitrary SQLs and return a `json_serve::Map` vector using the column name as the key and the `json_serve::Value` as the Value.

gwenn:
Fix #882 for valid JSON values

### DIFF
--- a/src/types/serde_json.rs
+++ b/src/types/serde_json.rs
@@ -1,24 +1,62 @@
 //! [`ToSql`] and [`FromSql`] implementation for JSON `Value`.
 
-use serde_json::Value;
+use serde_json::{Number, Value};
 
 use crate::types::{FromSql, FromSqlError, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
-use crate::Result;
+use crate::{Error, Result};
 
-/// Serialize JSON `Value` to text.
+/// Serialize JSON `Value` to text:
+///
+///
+/// | JSON   | SQLite    |
+/// |----------|---------|
+/// | Null     | NULL    |
+/// | Bool     | 'true' / 'false' |
+/// | Number   | INT or REAL except u64 |
+/// | _ | TEXT |
 impl ToSql for Value {
     #[inline]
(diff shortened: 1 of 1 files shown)


brandonros:
```rust
let connection = rusqlite::Connection::open("./foo.sqlite").unwrap();
let query = "SELECT * FROM foo";
let mut statement = connection.prepare(&query).unwrap();
let column_count = statement.column_count();
let mut rows = statement.query(()).unwrap();
let mut results = Vec::new();
while let Some(row) = rows.next().unwrap() {
    let mut row_map = serde_json::map::Map::new();
    for column_index in 0..column_count {
        let name = row.as_ref().column_name(column_index).unwrap().to_string();
        let sqlite_value_ref = row.get_ref_unwrap(column_index);
        let json_value: serde_json::Value = FromSql::column_result(sqlite_value_ref).unwrap();
        row_map.insert(name, json_value);
    }
    results.push(row_map);
}
```

```
Integer(1673966400)
Integer(1673967600)
Text([76, 111, 110, 103])
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Other(Error("expected value", line: 1, column: 1))', src/main.rs:515:94
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Is there a better way to pull all rows from SQLite database file into Rust as serde_json maps?

PR #1259, opened 2022-11-20
Activate SQLITE_OMIT_DEPRECATED

gwenn:
See [Recommended compile-time options](https://sqlite.org/compile.html#recommended_compile_time_options)
And https://sqlite.org/compile.html#omit_deprecated


thomcc:
TBH, I don't know if we should do this -- If functions are deprecated in the future, it would make it easy for us to need to make a breaking change when updating bundled SQLite.

PR #1277, opened 2023-01-09
Try probing OpenSSL using pkg-config

ldm0:
Developer might install OpenSSL via package manager, probing it with pkg-config will make life easier.

`rust-openssl` does this too https://github.com/sfackler/rust-openssl/blob/2c0c391f8abea99abdf016175d5d7c257891541d/openssl-sys/build/find_normal.rs#L194


gwenn:
I guess you can ignore the code coverage error (which may be related to #1274).
But you should fix the Clippy warnings.
Thanks.

thomcc:
I'd like to take a look at this before it lands (but give me a couple days).

PR #1289, opened 2023-02-13
Make it build for wasm32-wasi

Maaarcocr:
`CARGO_CFG_TARGET_ENV` is not being set when compiling to `wasm32-wasi` and if it isn't then the flags used for compiling sqlite correctly for it do not get applied. We can use `TARGET` instead, which seems to be always set. 

Also, `-DSQLITE_THREADSAFE=0` is needed as a flag, as without it, sqlite will use `pthread.h` which is not defined for wasi targets.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -19,11 +19,11 @@ fn android_target() -> bool {
 }
 
 /// Tells whether a given compiler will be used `compiler_name` is compared to
-/// the content of `CARGO_CFG_TARGET_ENV` (and is always lowercase)
+/// the content of `TARGET` (and is always lowercase)
 ///
 /// See [`win_target`]
 fn is_compiler(compiler_name: &str) -> bool {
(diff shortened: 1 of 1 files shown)


libsqlite3-sys/build.rs
```diff
@@ -19,11 +19,11 @@ fn android_target() -> bool {
 }
 
 /// Tells whether a given compiler will be used `compiler_name` is compared to
-/// the content of `CARGO_CFG_TARGET_ENV` (and is always lowercase)
+/// the content of `TARGET` (and is always lowercase)
 ///
 /// See [`win_target`]
 fn is_compiler(compiler_name: &str) -> bool {
-    env::var("CARGO_CFG_TARGET_ENV").map_or(false, |v| v == compiler_name)
+    env::var("TARGET").map_or(false, |v| v == compiler_name)
```
gwenn, line 26: I am not sure but for me this changes breaks cross compilation.

PR #1304, opened 2023-03-22
Add Statement.readonly() accessor

jgraettinger:
Condition on the `extra_check` feature, which matches the implementation within RawStatement.

Thanks

### DIFF
--- a/src/raw_statement.rs
+++ b/src/raw_statement.rs
@@ -197,7 +197,6 @@ impl RawStatement {
     }
 
     // does not work for PRAGMA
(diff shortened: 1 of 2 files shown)


jgraettinger:
To clarify, you mean remove cfg(extra_check) from RawStatement so that the `Statement::readonly` may be used without that feature?

gwenn:
> To clarify, you mean remove cfg(extra_check) from RawStatement so that the `Statement::readonly` may be used without that feature?

Yes.

jgraettinger:
Done

PR #1313, opened 2023-04-05
Implement FromSql & ToSql for std::num::NonZero types

itsxaos:
Pros:
- pretty straightforward implementation
- support for more standard library types
- won't (shouldn't) break anything because of orphan rules

Cons:
- clutters up docs (and code) a little
- probably not very extensively used

I also added a little note about STRICT tables to the docs.

### DIFF
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -298,4 +298,48 @@ mod test {
         check_ranges::<u32>(&db, &[-2, -1, 4_294_967_296], &[0, 1, 4_294_967_295]);
         Ok(())
     }
+
+    #[test]
+    fn test_nonzero_ranges() -> Result<()> {
+        let db = Connection::open_in_memory()?;
+
+        macro_rules! check_ranges {
+            ($nz:ty, $out_of_range:expr, $in_range:expr) => {
+                for &n in $out_of_range {
+                    assert_eq!(
+                        db.query_row("SELECT ?1", [n], |r| r.get::<_, $nz>(0)),
+                        Err(Error::IntegralValueOutOfRange(0, n))
+                    );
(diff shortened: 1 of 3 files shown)


src/types/to_sql.rs
```diff
@@ -186,12 +218,27 @@ macro_rules! to_sql_self_fallible(
                 )))
             }
         }
+    );
+    (non_zero $t:ty) => (
+        impl ToSql for $t {
```
itsxaos, line 223: I wanted to implement this by delegating to the zero-able type, but it doesn't seem possible because the `NonZero::get` returns by-value and `ToSql::to_sql` is allowed to borrow from the temporary.

src/types/from_sql.rs
```diff
@@ -96,6 +96,15 @@ macro_rules! from_sql_integral(
                 i.try_into().map_err(|_| FromSqlError::OutOfRange(i))
             }
         }
+    );
+    (non_zero $nz:ty, $z:ty) => (
+        impl FromSql for $nz {
+            #[inline]
+            fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+                let i = <$z>::column_result(value)?;
```
gwenn, line 104: This seems wrong to me because SQLite supports only two kinds of integer (http://sqlite.org/c3ref/value_blob.html):
> sqlite3_value_int	→	32-bit INTEGER value
> sqlite3_value_int64	→	64-bit INTEGER value

And `rusqlite` actually uses only one: `sqlite3_value_int64`
So `<$z>` should always be `i64`.

gwenn (approved):

gwenn:
Thanks

PR #1314, opened 2023-04-13
Use track_caller for panicking methods

kornelski:
When `get_unwrap()` panics, the panic message points to `.cargo/registry/src/index.crates.io-6f17d22bba15001f/rusqlite-0.29.0/src/row.rs`, which is unhelpful for tracking down the cause of the panic.  With `#[track_caller]`, Rust uses the method caller's location, so the user will clearly see which of their queries and column getters are failing.

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -257,6 +257,7 @@ impl<'stmt> Row<'stmt> {
     /// * If the underlying SQLite integral value is outside the range
     ///   representable by `T`
     /// * If `idx` is outside the range of columns in the returned query
+    #[track_caller]
     pub fn get_unwrap<I: RowIndex, T: FromSql>(&self, idx: I) -> T {
         self.get(idx).unwrap()
     }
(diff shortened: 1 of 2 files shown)


gwenn:
For `column_name_unwrap` and `Value::from`, there should be no panic by contract.
So I am not sure that `track_caller` is needed.

kornelski:
`Value::from` has `expect` about UTF-8 validity. Is `ValueRef::Text` always guaranteed to hold UTF-8? If so, it's surprising that it's `ValueRef::Text(&'a [u8])` and not ``ValueRef::Text(&'a str)`.

gwenn:
> Is `ValueRef::Text` always guaranteed to hold UTF-8?

https://github.com/rusqlite/rusqlite/blob/ef4ae94799f70e59f71d900a170bad0dc4d1a10f/src/types/value_ref.rs#L226-L233
And http://sqlite.org/c3ref/value_blob.html:
> sqlite3_value_text	→	UTF-8 TEXT value

gwenn (approved):

gwenn:
Thanks.

PR #1317, opened 2023-04-16
Force linking against system sqlite libs

icp1994:
This commit introduces new env var, LIBSQLITE3_SYS_USE_PKG_CONFIG, which can be set to non-zero values to force building against sqlite libraries from the system overriding bundled features.

Prior art in other -sys crates:

- [ssh2-rs](https://github.com/alexcrichton/ssh2-rs/commit/a35a60aafedf101dc67ff24bab2cc5817f292fb7)
- [zstd-rs](https://github.com/gyscos/zstd-rs/commit/22e359a53f4b577a45ae28bed6a1717c1b0fb6a9)
- [rust-openssl](https://github.com/sfackler/rust-openssl/commit/830658ec0bdae8ba1f6310f2e8116750d01c1b82)

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -36,7 +36,11 @@ fn main() {
             .expect("Could not copy bindings to output directory");
         return;
     }
-    if cfg!(all(
+
+    println!("cargo:rerun-if-env-changed=LIBSQLITE3_SYS_USE_PKG_CONFIG");
+    if env::var_os("LIBSQLITE3_SYS_USE_PKG_CONFIG").map_or(false, |s| s != "0") {
+        build_linked::main(&out_dir, &out_path);
+    } else if cfg!(all(
         feature = "sqlcipher",
         not(feature = "bundled-sqlcipher")
     )) {


gwenn:
Currently `rusqlite` assumes that the [modern_sqlite](https://github.com/rusqlite/rusqlite/blob/ef4ae94799f70e59f71d900a170bad0dc4d1a10f/Cargo.toml#L48) feature can be used when the `bundled` feature is activated.
But this is not necessary case with your change (`LIBSQLITE3_SYS_USE_PKG_CONFIG=1`)...

icp1994:
Hmm, can you explain a bit more? As per my understanding, the modern_sqlite feature will in turn activate the libsqlite3-sys/bundled_bindings feature. That feature is handled inside [build_linked](https://github.com/rusqlite/rusqlite/blob/v0.29.0/libsqlite3-sys/build.rs#L357).

gwenn:
> As per my understanding, the modern_sqlite feature will in turn activate the libsqlite3-sys/bundled_bindings feature. That feature is handled inside [build_linked](https://github.com/rusqlite/rusqlite/blob/v0.29.0/libsqlite3-sys/build.rs#L357).

Sorry, you are right.

gwenn (approved):

gwenn:
I guess we can ignore the codecov error.

PR #1321, opened 2023-05-02
fix compilation for target wasm32-wasi

wasm-forge:
This PR partially reverts https://github.com/rusqlite/rusqlite/pull/961 that has introduced a bug in detection of `wasm32-wasi` target. That PR replaced the check of `TARGET` with a check of `CARGO_CFG_TARGET_ENV`. That however is incorrect because `wasm32-wasi` is a CPU architecture target (similar to `x86` and `armv8`) and not an operating system.

This PR only affects `wasm32-wasi` compilation and is a no-op for other targets.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -240,7 +240,7 @@ mod build_bundled {
             cfg.flag("-DHAVE_LOCALTIME_R");
         }
         // Target wasm32-wasi can't compile the default VFS
-        if is_compiler("wasm32-wasi") {
+        if env::var("TARGET").map_or(false, |v| v == "wasm32-wasi") {
             cfg.flag("-DSQLITE_OS_OTHER")
                 // https://github.com/rust-lang/rust/issues/74393
                 .flag("-DLONGDOUBLE_TYPE=double");


gwenn:
See #1116 and #1289

wasm-forge:
@gwenn, thanks for the reply and linking the existing attempts to fix the bug. I apologize for sending a PR with an empty description. I fixed it now.

The bug was introduced in https://github.com/rusqlite/rusqlite/commit/5511a12e57af7bead3d32bdc0618cf754e56a97e?diff=split#diff-c1c013e096409a5aea1f123c04254379057791785963eba5ef253fdd686570b8L120

My PR reverts the change that introduced the bug. I verified locally that wasm32-wasi works locally after this partial revert.

My fix looks very similar to PR [#1116](https://github.com/rusqlite/rusqlite/pull/1116), but it doesn't have the `|| is_compiler(..)` part which is superfluous. If you would like to proceed with PR [#1116](https://github.com/rusqlite/rusqlite/pull/1116), I am happy to close my PR and continue the work in [#1116](https://github.com/rusqlite/rusqlite/pull/1116).

The PR [#1289](https://github.com/rusqlite/rusqlite/pull/1289) is incorrect because it is changing the definition of `is_compiler()`.

gwenn:
Thanks for the explanation.

PR #1322, opened 2023-05-08
Add data argument to csvtab

marshauf:
Adds missing [csv virtual table](https://www.sqlite.org/csv.html) `data` argument.

### DIFF
--- a/src/vtab/csvtab.rs
+++ b/src/vtab/csvtab.rs
@@ -22,6 +22,7 @@
 //! }
 //! ```
 use std::fs::File;
+use std::io::Cursor;
 use std::marker::PhantomData;
 use std::os::raw::c_int;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


src/vtab/csvtab.rs
```diff
@@ -58,6 +59,7 @@ struct CsvTab {
     base: ffi::sqlite3_vtab,
     /// Name of the CSV file
     filename: String,
+    data: String,
```
gwenn, line 62: I am not sure but `filename` and `data` are exclusive.
So maybe we should introduce something like:
```rust
enum Source {
  File(String),
  Data(String),
}

impl Read for Source {
...
}
```

edmellum:
This would be useful to me too. I'd be happy to bring the PR up to the current version of Rusqlite if that would help get this merged? 😄

Issue #1323, opened 2023-05-08
Add ToSql/FromSql optional impls for smallstr crate


ultrabear:
This is a feature request for ToSql/FromSql implementations for the `SmallString` type from smallstr.

If this change is welcome, but its too much effort to implement I may be able to PR it sometime, please let me know if that's the case!

gwenn:
And you cannot use https://docs.rs/smallstr/0.3.0/smallstr/struct.SmallString.html#method.as_str ?

ultrabear:
That doesn't cover FromSql, which is the main thing I want to support (not having to allocate)

gwenn:
> That doesn't cover FromSql.
Indeed.
So just make sure to do your changes behind a feature.

ultrabear:
closing as the sentiment is to not implement this

PR #1325, opened 2023-05-15
impl ToSql and FromSql for SmallString from smallstr crate

ultrabear:
This implements the feature proposed in github issue #1323. 

This code passes clippy pedantic, fmt, and a new test written for it. 

One consideration is that `smallstr 0.3` relies on `smallvec 1.8`, wheras this crate holds a dependency on `smallvec 1.6`. I'm not too deep in dependency resolution knowledge but as I understand this wont pose a real problem

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,6 +122,7 @@ fallible-iterator = "0.2"
 fallible-streaming-iterator = "0.1"
 uuid = { version = "1.0", optional = true }
 smallvec = "1.6.1"
+smallstr = { version = "0.3", optional = true }
 
 [dev-dependencies]
 doc-comment = "0.3"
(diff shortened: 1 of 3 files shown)


thomcc:
I don't think this is worth doing? You can already read into whatever string type you want in a 0-copy way if you want.

This adds some ergonomics in exchange for requiring we do a semver-major bump if smallstr 0.4 comes out (which would cause issues for users still on 0.3).

ultrabear:
The main reason I had wanted it added was implementing it outside of rusqlite requires a newtype, and most people (including me) would not put that much effort into saving such a small amount of allocations, but by adding it here it lowers the cost of using it for the average rusqlite user

As for version bumps, smallstr has been 0.3 for over a year now, might be worth asking the maintainer directly but it seems to be stable where it is

ultrabear:
closing as the sentiment is to not implement this

PR #1327, opened 2023-05-18
remove depth from Savepoint

### LINKED ISSUE #1326: Savepoint depth is unnecessary
Savepoints nest with the same name naturally.  RELEASE and ROLLBACK with a savepoint name will always match the inner-most savepoint with a matching name.  [Keeping track of the depth is redundant](https://sqlite.org/lang_savepoint.html#transaction_nesting_rules):

> The RELEASE command starts with the most recent addition to the transaction stack and releases savepoints backwards in time until it releases a savepoint with a matching savepoint-name. Prior savepoints, even savepoints with matching savepoint-names, are unchanged. If the RELEASE command causes the transaction stack to become empty (if the RELEASE command releases the outermost transaction from the stack) then the transaction commits.
> 
> The ROLLBACK command with a TO clause rolls back transactions going backwards in time back to the most recent SAVEPOINT with a matching name. The SAVEPOINT with the matching name remains on the transaction stack, but all database changes that occurred after that SAVEPOINT was created are rolled back. If the savepoint-name in a ROLLBACK TO command does not match any SAVEPOINT on the stack, then the ROLLBACK command fails with an error and leaves the state of the database unchanged.

Taywee:
Savepoints with the same name nest properly without issue.  This doesn't work with rusqlite unless savepoints are always released on drop.

closes #1326

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -91,7 +91,6 @@ pub struct Transaction<'conn> {
 pub struct Savepoint<'conn> {
     conn: &'conn Connection,
     name: String,
-    depth: u32,
     drop_behavior: DropBehavior,
     committed: bool,
 }
@@ -158,13 +157,13 @@ impl Transaction<'_> {
     /// ```
     #[inline]
     pub fn savepoint(&mut self) -> Result<Savepoint<'_>> {
-        Savepoint::with_depth(self.conn, 1)
+        Savepoint::new_(self.conn)
     }
 
     /// Create a new savepoint with a custom savepoint name. See `savepoint()`.
     #[inline]
     pub fn savepoint_with_name<T: Into<String>>(&mut self, name: T) -> Result<Savepoint<'_>> {
-        Savepoint::with_depth_and_name(self.conn, 1, name)
+        Savepoint::with_name_(self.conn, name)
     }
 
     /// Get the current setting for what happens to the transaction when it is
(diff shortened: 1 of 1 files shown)


src/transaction.rs
```diff
@@ -352,7 +345,7 @@ impl Savepoint<'_> {
         }
         match self.drop_behavior() {
             DropBehavior::Commit => self.commit_().or_else(|_| self.rollback()),
-            DropBehavior::Rollback => self.rollback(),
+            DropBehavior::Rollback => self.rollback().and_then(|_| self.commit_()),
```
gwenn, line 348: To be checked.
Related to https://github.com/rusqlite/rusqlite/issues/1326#issuecomment-1553157537 ?

Taywee: Yes, related to the comment.  Without a commit call, the savepoint is left on the stack, so dropping a savepoint in rollback mode without this change will leave it on the savepoint stack.  It's different from non-savepoint transactions, where the ROLLBACK also releases it.

In practice, this hasn't caused much of an issue, because the savepoint is removed when rolling back or committing a savepoint (or transaction) higher in the stack, so creating a savepoint and then not interacting with it at all in the future is usually functionally the same as not doing anything at all.  A spot where this could really make a difference is if you create a savepoint that's not in a transaction, then the drop of the savepoint will leave you in that transaction and not in autocommit mode like you might expect:

```rust
let mut conn = Connection::open(db_path)?;
assert!(conn.is_autocommit());
{
    let savepoint = conn.savepoint()?;
}
assert!(conn.is_autocommit());
```

That last assert fails.

As a result, this fails with QueryReturnedNoRows:

```rust
    {
        let mut conn = Connection::open(&db_path)?;
        conn.execute("CREATE TABLE sample (id INTEGER PRIMARY KEY)", ())?;
        {
            let savepoint = conn.savepoint()?;
        }
        conn.execute("INSERT INTO sample (id) VALUES (1)", ())?;
    }
    {
        let mut conn = Connection::open(&db_path)?;
        let id: i64 = conn.query_row("SELECT id FROM sample", (), |row| row.get(0))?;
    }
```

Because the savepoint wasn't released.

I assume this issue is the reason the savepoint depth was thought to be necessary, because with the depth, you can at least always commit or roll back the right savepoint, while ignoring the dangling ones higher up.  This works fine until you are not interacting with a higher transaction.

Taywee: 1. No, with the depth, you still need to release the savepoint.  Notice https://sqlite.org/lang_savepoint.html

> The [ROLLBACK](https://sqlite.org/lang_transaction.html) command without a TO clause rolls backs all transactions and leaves the transaction stack empty.
> 
> The ROLLBACK command with a TO clause rolls back transactions going backwards in time back to the most recent SAVEPOINT with a matching name. The SAVEPOINT with the matching name remains on the transaction stack, but all database changes that occurred after that SAVEPOINT was created are rolled back. If the savepoint-name in a ROLLBACK TO command does not match any SAVEPOINT on the stack, then the ROLLBACK command fails with an error and leaves the state of the database unchanged.

To note: "The SAVEPOINT with the matching name remains on the transaction stack".  With the depth or without, there is a bug here.  My example code has the wrong behavior on the latest release version.

> The RELEASE command starts with the most recent addition to the transaction stack and releases savepoints backwards in time until it releases a savepoint with a matching savepoint-name. Prior savepoints, even savepoints with matching savepoint-names, are unchanged. If the RELEASE command causes the transaction stack to become empty (if the RELEASE command releases the outermost transaction from the stack) then the transaction commits.

2.  You don't actually need to remove the depth, but it's not buying any extra functionality or convenience.  It's extra stuff done in the code which partially masks the "release" bug.

gwenn: > 1. No, with the depth, you still need to release the savepoint.  Notice https://sqlite.org/lang_savepoint.html

So could you please provide a minimal unit test (that fails without `self.rollback().and_then(|_| self.commit_())` in current master or release branch) ?

Taywee: Yes, absolutely.  I should have done that in the first place.  It's added now.

gwenn: This is unexpected to me:
```sql
sqlite> CREATE TABLE t AS SELECT 1 as n;
sqlite> SAVEPOINT s; -- first savepoint
sqlite> UPDATE t SET n=2;
sqlite> SAVEPOINT s; -- second savepoint
sqlite> UPDATE t SET n=3;
sqlite> ROLLBACK TO s; -- second savepoint rollbacked
sqlite> RELEASE s; -- expected first savepoint to be released but seems to be the second one !!!
sqlite> RELEASE s; -- expected failure but first savepoint released
```

Taywee: Yes, I agree. It was also totally unexpected to me, and caused a lot of headaches for me when using savepoints heavily (not with rusqlite; it's a C++ codebase) until I discovered what was actually happening.  It's a strange and specific difference in semantics between `SAVEPOINT` and `TRANSACTION`.

I think the idea is that you could roll back the savepoint and then start it again from the beginning without having to re-create it.  I would have preferred consistency with other transaction operations, personally.

gwenn:
I guess that:
https://github.com/rusqlite/rusqlite/blob/841014bfd11f0077ff2f4c8a0f11c919d0e9cbb7/src/transaction.rs#L347
is buggy.
In case of error on RELEASE, we ROLLBACK the savepoint.
But we should ROLLBACK AND RELEASE, no?
```sql
sqlite> PRAGMA foreign_keys = true;
sqlite> CREATE TABLE r(n INTEGER PRIMARY KEY NOT NULL);
sqlite> CREATE TABLE f(n REFERENCES r(n) DEFERRABLE INITIALLY DEFERRED);
sqlite> SAVEPOINT s;
sqlite> INSERT INTO f VALUES (0);
sqlite> RELEASE s;
Runtime error: FOREIGN KEY constraint failed (19)
sqlite> ROLLBACK TO s;
sqlite> COMMIT; -- does not failed s is still on the stack...
```

Taywee:
Ah, yes, you're right.  I had overlooked that.

PR #1341, opened 2023-06-05
Serialize and deserialize database

### LINKED ISSUE #880: Is there any way to open a database from an &[u8]?
I'm working on a project where I get the contents of an sqlite3 database in the form of a pointer to a sequence of bytes. Is there any way I can open this as a database in rusqlite? I see `open_in_memory` but that seems to only create a new in-memory database, I don't see a way to specify opening from some already existing memory.

gwenn:
Fix #880
Port of [Tcl impl](https://github.com/sqlite/sqlite/blob/25e87ed1baa41de59ca37a5cda65a2104916632a/src/tclsqlite.c#L2622-L2696)
Partial duplicate of #786

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,6 @@ modern-full = [
 ]
 
 bundled-full = ["modern-full", "bundled"]
-default = ["serialize"]
 
 [dependencies]
 time = { version = "0.3.0", features = ["formatting", "macros", "parsing"], optional = true }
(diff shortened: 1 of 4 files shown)


gwenn:
@thomcc would you mind reviewing this minimum viable API ?
You should be able to serialize the data anywhere (disk / network) (with the `Deref<[u8]>` impl).
But to deserialize, users will need `sqlite3_malloc` / `sqlite3_realloc`...
(maybe we will be able to provide a nicer API when `std::alloc:: Allocator` is stable)

thomcc (approved):
This looks basically reasonable to me, sorry for the delay, I missed this in my inbox until now.

dimo414:
Any plans for a 0.30 to pick up this PR? Would love to use it!

PR #1346, opened 2023-06-10
Captured identifiers in SQL strings

### LINKED ISSUE #1253: Captured identifiers in SQL strings
I am not sure it is feasible but:
With the [Tcl interface to SQLite](https://sqlite.org/tclsqlite.html), you can do:
```tcl
db1 eval {INSERT INTO t1 VALUES(5,$bigstring)}
```
where `$bigstring` is a tcl variable.

And Rust [1.58.0](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings) supports this syntax:
```rust
println!("Hello, {person}!");
```
Is this feature only available to the standard library ?
If not, we should be able to improve `rusqlite` API !

gwenn:
Fix #1253


gwenn:
@thomcc would you mind reviewing this PR ?
Maybe current implementation is too restrictive: https://github.com/rusqlite/rusqlite/pull/1346/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R218-R223 ?

thomcc:
I'll try to review this over the weekend.

gwenn:
Official SQLite parser can be configured to support Tcl expression:
https://github.com/sqlite/sqlite/blob/78ad745ab9f36fa394c4701933d462b8453fe4a5/src/tokenize.c#L486-L499
I will try to make the parser allow some rust expression.

gwenn:
https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html#alternative-solution---interpolation
> If any expressions beyond identifiers become accepted in format strings, then the RFC author expects that users will inevitably ask "why is my particular expression not accepted?". This could lead to feature creep, ...

So maybe we should not ?

gwenn:
Sorry, last commit b86d932 is wrong: without an SQLite patch or a rewrite of original query (that we don't want), we cannot support neither Rust syntax `{....}` nor expression.
I will rollback

gwenn:
@thomcc would you mind reviewing this PR ?

PR #1354, opened 2023-06-25
Statement::column_table_name

baummarten:
Wrapper around `sqlite3_column_table_name` from https://www.sqlite.org/capi3ref.html#sqlite3_column_database_name

### DIFF
--- a/src/raw_statement.rs
+++ b/src/raw_statement.rs
@@ -70,6 +70,24 @@ impl RawStatement {
     }
 
     #[inline]
+    pub fn column_table_name(&self, idx: usize) -> Option<&CStr> {
+        let idx = idx as c_int;
+        if idx < 0 || idx >= self.column_count() as c_int {
+            return None;
+        }
+
+        unsafe {
+            let ptr = ffi::sqlite3_column_table_name(self.ptr, idx);
+
+            assert!(
+                !ptr.is_null(),
+                "Null pointer from sqlite3_column_table_name: Out of memory?"
+            );
+            Some(CStr::from_ptr(ptr))
+        }
+    }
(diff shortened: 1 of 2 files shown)


gwenn:
See https://sqlite.org/c3ref/column_database_name.html
> These APIs are only available if the library was compiled with the [SQLITE_ENABLE_COLUMN_METADATA](https://sqlite.org/compile.html#enable_column_metadata) C-preprocessor symbol.

Which means that you must introduce a feature...

src/raw_statement.rs
```diff
@@ -69,6 +69,24 @@ impl RawStatement {
...
 
+    #[inline]
+    pub fn column_table_name(&self, idx: usize) -> Option<&CStr> {
+        let idx = idx as c_int;
+        if idx < 0 || idx >= self.column_count() as c_int {
+            return None;
+        }
+
+        unsafe {
+            let ptr = ffi::sqlite3_column_table_name(self.ptr, idx);
```
gwenn, line 80: `sqlite3_column_table_name` depends on `SQLITE_ENABLE_COLUMN_METADATA` compile option.
So you must introduce a feature.

gwenn (changes requested):

baummarten:
Right, sorry, I didn't notice that. So it would have to depend on `bundled`, or else it would risk breaking someone's build if this feature gets enabled and the system sqlite wasn't compiled with this option. Seems more trouble than it's worth then; since writing the PR I found out I was using it wrongly to disambiguate columns after a join, where I should have been usin `as` to rename them instead. Nevermind then, sorry for the trouble.

PR #1357, opened 2023-07-07
Fixes generate_series to handle NULL arguments

davidselassie:
SQLite's `generate_series` behavior has it return an empty result set whenever any of the arguments are `NULL`.

```
SQLite version 3.37.2 2022-01-06 13:25:41
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> SELECT value FROM generate_series(0, 3);
┌───────┐
│ value │
├───────┤
│ 0     │
│ 1     │
│ 2     │
│ 3     │
└───────┘
sqlite> SELECT value FROM generate_series(0, NULL);
sqlite> SELECT value FROM generate_series(NULL);
sqlite> SELECT value FROM generate_series(0, 3, NULL);
sqlite> SELECT value FROM generate_series(0, NULL, 2);
sqlite> SELECT value FROM generate_series(NULL, 3, 2);
sqlite> 
```

rusqlite's re-implementation of `generate_series` that's behind the `series` Cargo feature has code suggesting it should have the same behavior of returning no rows, but instead errors out if any of the arguments are `NULL`.

```
SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("Invalid filter parameter type Null at index 0"))
```

This error was due to the `args.get(i)?` originally being inferred to be `args.get::<i64>(i)?` since they're being assigned to the parameters directly and thus `NULL` causes an error.

This PR fixes the behavior of rusqlite's `generate_series` to match that of SQLite's extension. It does this by interpreting those arguments as `Option<i64> and if they are `None`, then marks that the range should be set to result in no rows.

Adds some tests to the existing series module tests that show that `NULL` arguments result in no rows.

### DIFF
--- a/src/vtab/series.rs
+++ b/src/vtab/series.rs
@@ -200,19 +200,19 @@ unsafe impl VTabCursor for SeriesTabCursor<'_> {
         let mut idx_num = QueryPlanFlags::from_bits_truncate(idx_num);
         let mut i = 0;
         if idx_num.contains(QueryPlanFlags::START) {
-            self.min_value = args.get(i)?;
+            self.min_value = args.get::<Option<_>>(i)?.unwrap_or_default();
             i += 1;
         } else {
             self.min_value = 0;
         }
         if idx_num.contains(QueryPlanFlags::STOP) {
-            self.max_value = args.get(i)?;
+            self.max_value = args.get::<Option<_>>(i)?.unwrap_or_default();
             i += 1;
         } else {
             self.max_value = 0xffff_ffff;
         }
         if idx_num.contains(QueryPlanFlags::STEP) {
-            self.step = args.get(i)?;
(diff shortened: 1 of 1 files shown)


gwenn:
See:
https://www.sqlite.org/cgi/src/info/3328e828e28ca719
And:
https://github.com/rusqlite/rusqlite/commit/b60ffe1940deb3a41d5a97969682f181be0e40a5#diff-1592000f020c13e61a6b81bdd6bec4f8c2a54bcadb4949cb660bdb81c5b1c346R215-R222

gwenn:
> self.min_value = args.get::<Option<i64>>(i)?.unwrap_or_default();

Yes, could you please use this ?
Thanks.

gwenn:
Thanks.

PR #1367, opened 2023-07-14
Add FromSql/ToSql support for PathBuf and Utf8PathBuf types.

tarka:


gwenn:
I am afraid that comment https://github.com/rusqlite/rusqlite/pull/1325#issuecomment-1548592607 also applies here.
(`&[u8] or &str <-> X`)

tarka:
Fair enough. Would just the stdlib `PathBuf` be useful? If so I can update the PR.

gwenn:
Neither https://github.com/sfackler/rust-postgres nor https://github.com/launchbadge/sqlx seem to support `Path` / `PathBuf` so ...

thomcc:
Also it requires adding a new error type... I would prefer users handle it.

PR #1375, opened 2023-07-29
Fix ergonomics of Statements

### LINKED ISSUE #1337: Lifetime requirement on `Statement` is very unergonomic
The examples of this repository do something like `connection.prepare("SELECT ...").query(...)`, in other words they prepare a statement and then immediately execute it.

Correct me if I'm wrong, but an optimized application should at initialization open the database and prepare all the statements that it will need, and then later execute these statements. No statement is prepared after initialization.

However, the fact that `Statement` has a lifetime parameter makes this very difficult to do. It is for example not possible to store a `Connection` and a `Statement` in the same struct.

This is a well-known problem in the Rust ecosystem, and my opinion is that using lifetimes on long-lived objects is almost always "wrong" (in the sense that it leads to unergonomic designs).

I would suggest three possible ways to fix this:

- Have `Statement` always hold an `Arc<Connection>`.
- Make `Statement` generic over a pointer to a connection. For example the user could use `Statement<&'a Connection>` or `Statement<Arc<Connection>>` are their leisure. Note that this might lead to unsafety because the user could provide a custom malicious implementation of the `Deref` trait that doesn't always deref to the same connection.
- Have the `Connection` some kind of `Vec<*mut ffi::Statement>` (most likely a `Slab` so that you can insert and remove elements in `O(1)` time), and turn `Statement` into some kind of `StatementIndex` that represents an index in this container.

bbqsrc:
Fixes #1337.

This patch does the following:

- All querying functions now consume the `Statement`
- Merges `CachedStatement` into `Statement`
- Makes `RawStatement` responsible for knowing whether the statement is cached or not
- Updates the `Statement` drop implementation to handle the caching
- Updates the lifetime name on `Rows` from `'stmt` to `'conn` to make it clear the lifetime is for the connection

The majority of the lines changed are changes to the tests to make the queries no longer need to be `mut`. All tests passed with `--features bundled-full,modern-full`.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -4,7 +4,7 @@ use crate::raw_statement::RawStatement;
 use crate::{Connection, PrepFlags, Result, Statement};
 use hashlink::LruCache;
 use std::cell::RefCell;
-use std::ops::{Deref, DerefMut};
+
 use std::sync::Arc;
 
 impl Connection {
@@ -35,7 +35,7 @@ impl Connection {
     /// Will return `Err` if `sql` cannot be converted to a C-compatible string
     /// or if the underlying SQLite call fails.
     #[inline]
-    pub fn prepare_cached(&self, sql: &str) -> Result<CachedStatement<'_>> {
+    pub fn prepare_cached(&self, sql: &str) -> Result<Statement<'_>> {
         self.cache.get(self, sql)
     }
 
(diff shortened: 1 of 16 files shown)


gwenn:
Are you sure you actually fix the original #1337 issue.
Because the original issue is about the lifetime on `Statement` which is not touched by your PR.

bbqsrc:
As explicitly stated in the description, the lifetime has changed even if the number of lifetimes did not. Previously the lifetime was dependent on the statement. Instead, it is now dependent on the connection.

It solves the ergonomics problem caused by the previous lifetime without needing to introduce the overhead of Arc or similar smart pointers and still guarantees a statement cannot outlive a connection and cause UB.

bbqsrc:
That said, I suppose it’s up to interpretation. The lifetime can be removed entirely if Arcs are preferred.

My interpretation of the follow up comments was that using prepare_cached was the preferred approach to handling the basic ergonomics issue, and inferred that the lifetime issues were relating to the actual usage of the statement type, hence my PR.

gwenn:
The original issue is on https://github.com/rusqlite/rusqlite/blob/e093d7db7bc4ee33d24bddd381c288d702e3de3d/src/statement.rs#L18-L19.

bbqsrc:
Very constructive. Good job.

gwenn:
it's reciprocal.

gwenn:
@bbqsrc I guess that you want something like [CursorWithOwnership](https://docs.rs/sqlite/latest/sqlite/struct.CursorWithOwnership.html).
In contrast to the [Cursor](https://docs.rs/sqlite/latest/sqlite/struct.Cursor.html) which is like rusqlite [Rows](https://docs.rs/rusqlite/latest/rusqlite/struct.Rows.html)...

Issue #1358, opened 2023-07-07
[clippy] arc_with_non_send_sync - interrupt_lock


gwenn:
```
error: usage of `Arc<T>` where `T` is not `Send` or `Sync`
  --> src/inner_connection.rs:48:29
   |
48 |             interrupt_lock: Arc::new(Mutex::new(db)),
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: consider using `Rc<T>` instead or wrapping `T` in a std::sync type like `Mutex<T>`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
   = note: `#[deny(clippy::arc_with_non_send_sync)]` on by default
```
Clippy seems to suggest that the `Arc` / `Mutex` is useless because the connection cannot be shared by multiple threads, no ?

gwenn:
https://github.com/rust-lang/rust-clippy/issues/11076 => false positive

nathaniel-daniel:
I think this is actually https://github.com/rust-lang/rust-clippy/issues/11382, as https://github.com/rust-lang/rust-clippy/issues/11076 has been closed already. There is no generic code here. Clippy seems confused as to why a pointer, which should make the `Arc` and the type that contains it not `Send`, is being used over an `Rc`, missing the [unsafe `Send` impl for `InnerConnection`](https://github.com/rusqlite/rusqlite/blob/36264dc3cc4bce89b9f20dc030d49d4c6a0045b8/src/inner_connection.rs#L40).

gwenn:
https://github.com/rusqlite/rusqlite/pull/408/files
`*mut ffi::sqlite3` / `InterruptHandle` are `Send`able
So for me, Clippy is wrong (and PR #1397 is invalid).
But I am not completely sure.

nathaniel-daniel:
> *mut ffi::sqlite3 / InterruptHandle are Sendable

Well technically, `mut ffi::sqlite3` is not (since its a pointer), while `InterruptHandle` is (since it is done with `unsafe impl Send`), which is tripping up Clippy. Clippy doesn't seem to consider `unsafe impl Send` stuff in this lint.

> So for me, Clippy is wrong (and PR https://github.com/rusqlite/rusqlite/pull/1397 is invalid).

[...]

PR #1381, opened 2023-08-26
Upgrade SQLite bundled version to 3.43.0

gwenn:


gwenn:
For the Clippy error, see #1358

PR #1386, opened 2023-09-05
expose begin concurrent

Monadic-Cat:
Hello! This PR exposes `BEGIN CONCURRENT` in `TransactionBehavior`, so we can take advantage of writer concurrency as described here: https://www.sqlite.org/cgi/src/doc/begin-concurrent/doc/begin_concurrent.md

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -15,6 +15,13 @@ pub enum TransactionBehavior {
     /// EXCLUSIVE prevents other database connections from reading the database
     /// while the transaction is underway.
     Exclusive,
+    /// CONCURRENT is only available in "wal" or "wal2" mode, and allows multiple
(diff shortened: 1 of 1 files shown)


Monadic-Cat:
> Note that this is not present in any released version of SQLite, so I'm not sure we should expose it by default.

That's fair. I somehow hadn't considered that. Would it make sense to add this behind a feature flag, or maybe just wait until it gets released to add it here?

thomcc:
I think it's been several years, so I don't have high confidence it will ever be released. I think it was an experiment that didn't pan out. So I would prefer waiting for it to be released.

Monadic-Cat:
I'll be back again with this PR if/when it does pan out. Thank you for your time.

PR #1387, opened 2023-09-07
Change FnMut to  Fn in create_scalar_function

branchseer:
The `x_func` parameter of `create_scalar_function` can be re-entered, so it must be an `Fn`.

Code Example:

```rust
let mut conn = rusqlite::Connection::open_in_memory().unwrap();
conn.create_scalar_function("recursive_fn", 1, FunctionFlags::default(), |ctx| {
    let param = ctx.get::<i64>(0)?;
    println!("enter recursive_fn {param}");
    let conn_ref = unsafe { ctx.get_connection()? };
    if param >= 1 {
        conn_ref.query_row("SELECT recursive_fn(?)", params![ param - 1 ], |_| Ok(()))?;
    }
    println!("exit recursive_fn {param}");
    Ok(0)
}).unwrap();
conn.execute_batch("SELECT recursive_fn(3)").unwrap();
```

Output:
```
enter recursive_fn 3
enter recursive_fn 2
enter recursive_fn 1
enter recursive_fn 0
exit recursive_fn 0
exit recursive_fn 1
exit recursive_fn 2
exit recursive_fn 3
```

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -446,7 +446,7 @@ impl Connection {
         x_func: F,
     ) -> Result<()>
     where
-        F: FnMut(&Context<'_>) -> Result<T> + Send + 'static,
+        F: Fn(&Context<'_>) -> Result<T> + Send + 'static,
         T: SqlFnOutput,
     {
         self.db
@@ -549,7 +549,7 @@ impl InnerConnection {
         x_func: F,
     ) -> Result<()>
     where
-        F: FnMut(&Context<'_>) -> Result<T> + Send + 'static,
+        F: Fn(&Context<'_>) -> Result<T> + Send + 'static,
         T: SqlFnOutput,
     {
         unsafe extern "C" fn call_boxed_closure<F, T>(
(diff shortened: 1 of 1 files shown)


gwenn:
Just for reference:
https://github.com/rusqlite/rusqlite/commit/29494f46f6cf7528caed429048f1a60afe8e7134

gwenn:
Is there any safety issue ?
In general, `FnMut` can be recursive, no ? (https://users.rust-lang.org/t/mutability-and-recursion/35080)

branchseer:
@mrothNET 

> Currently, there is no safe way to obtain a connection reference within the callback function.

Well, we can get the connection in the callback using thread-local storage without any `unsafe`:

```rust
thread_local! {
    static CONN: RefCell<Option<Connection>> = RefCell::new(None);
}

CONN.with_borrow_mut(|conn_mut| *conn_mut = Some(Connection::open_in_memory().unwrap()));
CONN.with_borrow(|conn| {
    let conn = conn.as_ref().unwrap();
    let mut num = 0;

    conn.create_scalar_function("recursive_fn", 1, FunctionFlags::default(), move |ctx| {
        let num = &mut num;
        println!("num is mutably borrowed");
        let param = ctx.get::<i64>(0)?;
        CONN.with_borrow(|conn| -> rusqlite::Result<()> {
            let conn = conn.as_ref().unwrap();
            if param >= 1 {
                conn.query_row("SELECT recursive_fn(?)", params![ param - 1 ], |_| Ok(()))?;
            }
            Ok(())
        })?;
        drop(num);
        println!("num's mutable reference is dropped");
        Ok(0)
    }).unwrap();

    conn.execute_batch("SELECT recursive_fn(3)").unwrap();
})
```

My point is, custom functions are by design reentrant. I believe the developers of SQLite have never felt the need to add any check or lock to prevent them from being re-entered, and that makes them `Fn`.

PR #1390, opened 2023-09-16
Change deserialize to support embedded database buffer

retro-git:
I'm working on a project where I use SQLite as a storage medium for a large amount of static, readonly data. I wanted to embed the database file directly into the executable using the include_bytes! macro so the program can be distributed as a single executable. To make this approach work, I need to be able to call deserialize() on the byte buffer.

However, the current implementation expects an OwnedData argument to deserialize(), assuming the input data is a result of the serialize() function. To cater to the embedded use-case, I have made the following changes:

1. Modified the deserialize function to accept a byte slice.
2. Added a to_slice method for OwnedData.

Due to these changes, the into_raw function isn't invoked anymore. This means the SQLITE_DESERIALIZE_FREEONCLOSE flag can be removed (otherwise causes a double free). The Drop implementation of OwnedData will handle memory cleanup.

I'm open to feedback and suggestions if there's an alternate or more optimal approach.

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -34,6 +34,10 @@ impl OwnedData {
         std::mem::forget(self);
         raw
     }
+
+    pub fn as_slice(&self) -> &[u8] {
+        unsafe { std::slice::from_raw_parts(self.ptr.as_ptr(), self.sz) }
+    }
 }
 
 impl Drop for OwnedData {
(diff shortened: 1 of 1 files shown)


gwenn:
> However, the current implementation expects an OwnedData argument to deserialize(), assuming the input data is a result of the serialize() function.

The only constraint is:
https://github.com/rusqlite/rusqlite/blob/6176b11a3008d749cf501ff5d3782935e5348af3/src/serialize.rs#L25-L27

If you can't then please add a _new_ method.
Which (I guess) should be unsafe because the reference must be kept valid until the database is closed.
Or to say otherwise, your current implementation is wrong...

gwenn:
https://github.com/launchbadge/sqlx/pull/2759

PR #1395, opened 2023-10-06
Make WindowAggregate::value pass mutable value ref

### LINKED ISSUE #1394: Custom window functions and mutability
I am working on [sqlite-hashes](https://github.com/nyurik/sqlite-hashes) -- adding hash functions like `md5` and `sha256` to SQLite.

In order to create `sha256_concat(val)` - a hash over a sequence, each value must be consistently ordered.  SQLite does NOT guarantee the order of execution for the regular aggregate functions even if they have an `ORDER BY` in the query, so the result in BOTH of these examples is NOT defined! 🤦 (see [group_concat ordering question](https://stackoverflow.com/questions/1897352/sqlite-group-concat-ordering)):

```sql
SELECT sha256_concat(val) FROM tbl ORDER BY val;
SELECT sha256_concat(val) FROM (SELECT val FROM tbl ORDER BY val));
```

I think I can get around it with a custom windowing function, but it has some limitations too:  the aggregation state cannot be "inversed" (once a value is added to the hash, it cannot be removed). It is also harder to "partially-compute" a value (compute an interim hash, and continuing to add values to it).

Should the `xValue` receive a mutable state, just like `xInverse`?  This way partial computations could be achieved somehow? Or do you have any other suggestions on implementing something like this?

Thanks!  See also my related question on [stackoverflow](https://stackoverflow.com/questions/77246426/ensuring-ordering-for-a-custom-aggregate-windowing-function-in-sqlite) (I focus on the actual SQL usage there).

nyurik:
Fixes #1394

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -301,7 +301,7 @@ where
 {
     /// Returns the current value of the aggregate. Unlike xFinal, the
     /// implementation should not delete any context.
-    fn value(&self, acc: Option<&A>) -> Result<T>;
+    fn value(&self, acc: Option<&mut A>) -> Result<T>;
 
     /// Removes a row from the current window.
     fn inverse(&self, ctx: &mut Context<'_>, acc: &mut A) -> Result<()>;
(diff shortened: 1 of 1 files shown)


nyurik:
Seems like there is an unrelated Clippy issue.  I can submit a PR for auto-checking that, similar to what I have in [my crate](https://github.com/nyurik/sqlite-hashes):
* automatically adds a `git-push` hook to run `cargo clippy && cargo test` before pushing (and explains how to skip it)
* use `justfile` (not required, but convenient) for all sorts of dev life automation
*

nyurik:
@gwenn hi, a friendly ping - let me know if anything else is needed here and in other PRs. Thanks!!

PR #1396, opened 2023-10-06
Add `#[derive(Clone, Copy...` on all bitflags

nyurik:
This will make it easier to use them, e.g. I won't need to re-create `FunctionFlags` for registering multiple similar functions.


gwenn:
`Eq` and `PartialEq` seem useless

nyurik:
They are clearly not as useful as `Copy`, agree, but I am sure there could be usecases where users will want to pass them around and compare them, wouldn't they?  Also, there could be some implementers that wrap your library, and they might need to take extra actions before passing bitfield into the inner lib.   But this is up to you - my main req is to copy bitfields to pass them around more easily (e.g. have a global const/static, etc).

Speaking of derive - there are many enums that also do not implement `Copy`, e.g. `types::Type` should be both `Copy` and `Hash` (someone may want to have a lookup for type -> value conversion of sorts?).  Do you have any overall rule of thumb of what gets which derives? Thx!

gwenn:
>  I am sure there could be usecases where ...

YAGNI: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

nyurik:
@gwenn YAGNII is my fav principle too, but the cost of submitting a PR and discussing it, and waiting for release is also pretty high, so what's the middle ground?  I will add the clone/copy where makes sense - that's what I do need. Is that ok?

gwenn:
> I will add the clone/copy where makes sense - that's what I do need. Is that ok?

Yes, please.

nyurik:
i tried to clean them up a bit, adding just the copy/clone where it seem to make sense.  Overall, looking at many `pub enum` cases, it seems the derives are very unorganized/ad-hoc, without much consistency.  I will try to tame my OCD though :)

nyurik:
@gwenn should i make any changes to this PR?

src/vtab/mod.rs
```diff
@@ -322,7 +323,7 @@ pub trait UpdateVTab<'vtab>: CreateVTab<'vtab> {
 
 /// Index constraint operator.
 /// See [Virtual Table Constraint Operator Codes](https://sqlite.org/c3ref/c_index_constraint_eq.html) for details.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
gwenn, line 326: Created only by SQLite itself (readonly on rusqlite side)

src/vtab/mod.rs
```diff
@@ -59,6 +59,7 @@ use crate::{str_to_cstring, Connection, Error, InnerConnection, Result};
 // ffi::sqlite3_vtab_cursor => VTabCursor
 
 /// Virtual table kind
+#[derive(Copy, Clone, Debug)]
```
gwenn, line 62: Used only as a constant at compile time.

src/vtab/series.rs
```diff
@@ -28,7 +28,7 @@ const SERIES_COLUMN_STOP: c_int = 2;
 const SERIES_COLUMN_STEP: c_int = 3;
 
 bitflags::bitflags! {
-    #[derive(Clone, Copy)]
+    #[derive(Clone, Copy, Debug)]
```
gwenn, line 31: Use only to pass information between `best_index` and `filter` methods.

src/session.rs
```diff
@@ -655,7 +655,7 @@ impl Connection {
 /// See [here](https://sqlite.org/session.html#SQLITE_CHANGESET_CONFLICT) for details.
 #[allow(missing_docs)]
 #[repr(i32)]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
gwenn, line 658: Created only by SQLite (readonly on rusqlite side)

src/vtab/mod.rs
nyurik: reverted

src/vtab/mod.rs
nyurik: reverted

src/session.rs
nyurik: reverted

src/vtab/series.rs
nyurik: reverted

nyurik (commented):
thx, reverted the ones you suggested, and also merged main for ease of review.  You may want to squash when merging though

PR #1398, opened 2023-10-07
Use proper var names in trait definition

nyurik:
The underscores are too confusing, plus IDE automatically copies them into the implementation, and the `_` is not what most people would expect.

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -272,11 +272,11 @@ where
     /// call to [`step()`](Aggregate::step) to set up the context for an
     /// invocation of the function. (Note: `init()` will not be called if
     /// there are no rows.)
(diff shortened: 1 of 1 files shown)


gwenn:
> The underscores are too confusing
But then you have Clippy warnings...

nyurik:
> > The underscores are too confusing
> 
> But then you have Clippy warnings...

Why?  The `rusqlite` will compile just fine (see the build log - there are no related issues). On the other hand, the implementer of the trait will be responsible for either using them or preceding them with an underscore.  This is a common practice.  But i do expect all parameter names to be properly named so I know what they are without looking them up.

gwenn:
@nyurik 
Sorry, you are right !

PR #1411, opened 2023-11-07
Make fallible iterator dependencies optional

### LINKED ISSUE #833: Consider making `fallible_iterator` an optional or dev dependency?
I'm pretty okay with our use of `fallible_streaming_iterator`, but I don't really see the value of `fallible_iterator` here. It's mostly redundant with `Iterator` now, and most of its ergonomics can be achieved with `transpose()`.

We only implement it for `Map` which, aside from `FallibleIterator` is redundant with the `MappedRows`.

Because we don't reexport `fallible_iterator`, consumers have to depend on the `fallible_iterator` crate directly in order to take advantage of that — However, looking through https://crates.io/crates/rusqlite/reverse_dependencies it looks like nobody does (for the first couple pages anyway).

We use it in some tests and I'm okay with it as a dev dep (or even an optional dep), but yeah, as it is I think we should remove it and just use `Iterator`s.

tnull:
Fixes #833.

For various reasons it's nice to reduce the number of mandatory downstream dependencies. In this PR, we make `fallible_iterator` and `fallible_streaming_iterator` optional dependencies so that they can be disabled by users that don't require them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ name = "rusqlite"
 members = ["libsqlite3-sys"]
 
 [features]
+default = ["fallible-streaming-iterator"]
 load_extension = []
 # hot-backup interface: 3.6.11 (2009-02-18)
 backup = []
@@ -108,7 +109,8 @@ modern-full = [
     "uuid",
     "vtab",
     "window",
-    "fallible-iterator"
+    "fallible-iterator",
+    "fallible-streaming-iterator",
 ]
 
 bundled-full = ["modern-full", "bundled"]
@@ -122,7 +124,7 @@ serde_json = { version = "1.0", optional = true }
 csv = { version = "1.1", optional = true }
 url = { version = "2.1", optional = true }
 fallible-iterator = { version = "0.3", default-features = false, optional = true }
-fallible-streaming-iterator = "0.1"
+fallible-streaming-iterator = { version = "0.1", default-features = false, optional = true }
 uuid = { version = "1.0", optional = true }
 smallvec = "1.6.1"
 rusqlite-macros = { path = "rusqlite-macros", version = "0.1.0", optional = true }
--- a/src/collation.rs
+++ b/src/collation.rs
@@ -167,6 +167,7 @@ impl InnerConnection {
 }
 
 #[cfg(test)]
+#[cfg(feature = "fallible-streaming-iterator")]
 mod test {
     use crate::{Connection, Result};
     use fallible_streaming_iterator::FallibleStreamingIterator;
--- a/src/row.rs
+++ b/src/row.rs
@@ -1,5 +1,6 @@
 #[cfg(feature = "fallible-iterator")]
 use fallible_iterator::FallibleIterator;
+#[cfg(feature = "fallible-streaming-iterator")]
 use fallible_streaming_iterator::FallibleStreamingIterator;
 use std::convert;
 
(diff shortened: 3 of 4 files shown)


gwenn:
I see no change in [Github Actions script](https://github.com/rusqlite/rusqlite/blob/master/.github/workflows/main.yml).
So I assume that the associated code is not tested anymore !

src/vtab/series.rs
```diff
@@ -283,8 +283,10 @@ mod test {
     use crate::ffi;
     use crate::vtab::series;
     use crate::{Connection, Result};
+    #[cfg(feature = "fallible-iters")]
     use fallible_iterator::FallibleIterator;
```
gwenn, line 287: Please move this statement in the associated `test_series_module`

gwenn:
It seems that there is no [negative features](https://slint.dev/blog/rust-adding-default-cargo-feature).
And as said in #833, only `fallible_iterator` should be disabled (by default).

gwenn:
> > It seems that there is no [negative features](https://slint.dev/blog/rust-adding-default-cargo-feature).
> 
> I'm confused: I didn't attempt to introduce a negative feature? All features are additive and can still be used together, i.e., there is no "mutual exclusion" and no conflicts?

Do you agree that we are in the same situation ?
That is: no "default" and we would like to introduce negative features / optionally removing (not adding) dependencies.

And sorry to bother you, but for people using the "default-features = false" with `rusqlite`, it seems wise to have a PR that remove **only** `fallible-iterator`.

gwenn:
> I'm not sure how many users would depend on FallibleStreamingIterator in the first place.

https://github.com/search?q=lang%3Atoml+rusqlite+fallible-streaming-iterator+path%3ACargo.toml&type=code
https://github.com/search?q=lang%3Atoml+rusqlite+fallible-iterator+path%3ACargo.toml&type=code

Even the rust-playground !
https://github.com/rust-lang/rust-playground/blob/103d56130a265977854df8850238c2b719ad12f5/compiler/base/Cargo.toml#L343

> and depending on an ~unmaintained single-person project is unfortunately no real option for us.

https://crates.io/crates/fallible-streaming-iterator/reverse_dependencies
`arrow2` !

gwenn:
If you don't mind I will do a PR that removes **only** `fallible_iterator` as described in #833.

gwenn:
Current PR doesn't **remove** `fallible_iterator` !
Then, in a dedicated PR, we can make `fallible_streaming_iterator` optional without omitting `#[cfg_attr(docsrs, doc(cfg(feature = "fallible_streaming_iterator")))]`...

gwenn:
Minimal change #1416

gwenn:
Sorry I changed my mind !
https://docs.rs/fallible-iterator/latest/fallible_iterator/
> This will return the proper value when the reader operates successfully, but if it encounters an IO error, the result will either be slightly higher than expected if the error is transient, or it may run forever if the error is returned repeatedly!

So all `rusqlite` implementations of `Iterator` on `Rows` are wrong as described above.

gwenn:
You don't iterate on `Rows` (or only directly with a `while` loop) ?

gwenn:
> We have a single instance of `query_map`

So you are (potentially) impacted by the issue described by @sfackler:
> if it encounters an IO error, the result will either be slightly higher than expected if the error is transient, or it may run forever if the error is returned repeatedly!

gwenn:
https://blog.yoshuawuyts.com/fallible-iterator-adapters/
> Out of the 63 current stream adapters available, 30 take closures. Of those 30 adapters only 10 can bubble up a Result or have a counterpart that can do so. That means that 20 methods, or almost a third of the functionality in Iterator is currently incompatible with fallible operations.

gwenn:
So, evidences suggest that it is wise to keep using `Fallible[Streaming]Iterator` instead of standard `Iterator` for `Rows`.
[postgres](https://docs.rs/postgres/latest/postgres/struct.RowIter.html) does the same.

PR #1422, opened 2023-12-17
Improve VersionMismatch error message

nyurik:
The current error is a bit unclear, so added some clarification of which version is which.

### DIFF
--- a/libsqlite3-sys/src/error.rs
+++ b/libsqlite3-sys/src/error.rs
@@ -292,7 +292,10 @@ impl ::std::fmt::Display for InitError {
                 compile_time,
                 runtime,
             } => {
-                write!(f, "SQLite version mismatch: {runtime} < {compile_time}")
+                write!(
+                    f,
+                    "SQLite version mismatch:  runtime v{runtime}  <  compiled v{compile_time}"
+                )
             }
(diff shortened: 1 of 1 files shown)


gwenn:
You should not try to guess [sqlite3_version](http://sqlite.org/c3ref/libversion.html).
Which is not (easily) comparable.
Hence the use of `sqlite3_libversion_number`.

nyurik:
@gwenn i'm not "guessing" it for comparison reason, just formatting it for ease-of reading.  Much better than simply print raw number i think

gwenn:
> @gwenn i'm not "guessing" it for comparison reason, just formatting it for ease-of reading. Much better than simply print raw number i think

For me,
https://github.com/rusqlite/rusqlite/pull/1422/files#diff-068438172230084847676a638395a6ee9914094c2cb2d76f2c5de07f0d136ad4R285 try to guess the `sqlite3_version`

nyurik:
@gwenn would you be ok with the first revision of this PR -- it simply prints the number with some context: https://github.com/rusqlite/rusqlite/pull/1422/commits/017d91d4e45268082c0da6dcbb31394dbc372e2a

nyurik:
@gwenn friendly ping with above question. Thx!

PR #1425, opened 2023-12-18
Implement ergonomic loadable extension macro

### LINKED ISSUE #1424: Simplify loadable extension implementation by providing boilerplate code
When implementing loadable extension, it seems every/most implementations will have this code (copied mostly from the example). In the spirit of "Rusqlite is an ergonomic wrapper" concept, I think it would be good to provide a macro or some other means to avoid code duplication and complexity in the user's code.

```rust
use std::os::raw::{c_char, c_int};
use rusqlite::ffi;
use rusqlite::ffi::SQLITE_NOTICE;
use rusqlite::{to_sqlite_error, Connection, Result};

#[allow(clippy::not_unsafe_ptr_arg_deref)]
#[no_mangle]
pub extern "C" fn sqlite3_extension_init(
    db: *mut ffi::sqlite3,
    pz_err_msg: *mut *mut c_char,
    p_api: *mut ffi::sqlite3_api_routines,
) -> c_int {
    if p_api.is_null() {
        return ffi::SQLITE_ERROR;
    } else if let Err(err) = extension_init(db, p_api) {
        return unsafe { to_sqlite_error(&err, pz_err_msg) };
    }
    ffi::SQLITE_OK
}

fn extension_init(db: *mut ffi::sqlite3, p_api: *mut ffi::sqlite3_api_routines) -> Result<()> {
    let db = unsafe { Connection::extension_init2(db, p_api)? };
    
    // Function registration code goes here

    rusqlite::trace::log(SQLITE_NOTICE, "Loaded sqlite_hashes extension");
    Ok(())
}

### PR DESCRIPTION
**Pull request by nyurik**
This simplifies loadable extension code and removes the need for unsafe code.

Add a new (feature gated) macro called `sqlite3_extension_init!(my_init_fn)` that creates the needed unsafe SQLite3 extension entry point. The macro calls user's function, passing ownership of a safe database `Connection` struct.

Some of the macro content could, in theory, be moved to some function, or could be left as is.

Closes #1424

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1238,6 +1238,60 @@ impl InterruptHandle {
     }
 }
 
+/// Adds an entrypoint implementation for a SQLite extension.
+/// This macro must be used once in the root of a crate that implements a SQLite extension.
+///
+/// # Example
+/// ```
+/// use rusqlite::Connection;
+/// use rusqlite::ffi::SQLITE_NOTICE;
+/// use rusqlite::sqlite3_extension_init;
+/// use rusqlite::trace::log;
+///
+/// sqlite3_extension_init!(init_loadable_extension);
+///
+/// fn init_loadable_extension(db: Connection) -> ::rusqlite::Result<()> {
+///    db.create_scalar_function(
+///         "rusqlite_test_function",
+///         0,
+///         FunctionFlags::SQLITE_DETERMINISTIC,
+///         |_ctx| {
+///             Ok(ToSqlOutput::Owned(Value::Text(
+///                 "Rusqlite extension loaded correctly!".to_string(),
+///             )))
+///         },
+///     )?;
+///     log(SQLITE_NOTICE, "Rusqlite extension initialized");
+///     Ok(())
+/// }
+///```
+#[cfg(feature = "loadable_extension")]
+#[cfg_attr(docsrs, doc(cfg(feature = "loadable_extension")))]
+#[macro_export]
+macro_rules! sqlite3_extension_init {
+    ($func: ident) => {
+        #[allow(clippy::not_unsafe_ptr_arg_deref)]
+        #[no_mangle]
+        pub extern "C" fn sqlite3_extension_init(
+            db: *mut ::rusqlite::ffi::sqlite3,
+            pz_err_msg: *mut *mut std::os::raw::c_char,
+            p_api: *mut ::rusqlite::ffi::sqlite3_api_routines,
+        ) -> std::os::raw::c_int {
+            if p_api.is_null() {
+                return ::rusqlite::ffi::SQLITE_ERROR;
+            }
+
+            let res = unsafe { ::rusqlite::Connection::extension_init2(db, p_api) }.and_then($func);
+
+            if let Err(err) = res {
+                return unsafe { ::rusqlite::to_sqlite_error(&err, pz_err_msg) };
+            }
+
+            ::rusqlite::ffi::SQLITE_OK
+        }
+    };
+}
+
 #[cfg(doctest)]
 doc_comment::doctest!("../README.md");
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/lib.rs [hunk 1]
```diff
@@ -1238,6 +1238,60 @@ impl InterruptHandle {
...
+/// }
+///```
+#[cfg(feature = "loadable_extension")]
+#[cfg_attr(docsrs, doc(cfg(feature = "loadable_extension")))]
+#[macro_export]
+macro_rules! sqlite3_extension_init {
+    ($func: ident) => {
+        #[allow(clippy::not_unsafe_ptr_arg_deref)]
+        #[no_mangle]
+        pub extern "C" fn sqlite3_extension_init(
```
gwenn, line 1275: https://sqlite.org/loadext.html#programming_loadable_extensions
> You will do well to customize the name of your entry point to correspond to the name of the shared library you will be generating, rather than using the generic "sqlite3_extension_init" name. Giving your extension a custom entry point name will enable you to statically link two or more extensions into the same program without a linker conflict, if you later decide to use static linking rather than run-time linking.

nyurik: @gwenn thx, good point, thx. Looking further at [sqlite3_load_extension](https://sqlite.org/c3ref/load_extension.html), unless the user actually calls an API or specifies entry point by hand, sqlite uses heuristic to find this entry point automatically:

> t will first try the generic extension name "sqlite3_extension_init". If that does not work, it constructs a entry point using the template "sqlite3_X_init" where the X is replaced by the lowercase equivalent of every ASCII character in the filename after the last "/" and before the first following "." omitting the first three characters if they happen to be "lib". So, for example, if the filename is "/usr/lib/libmathfunc-4.8.so" the entry point name would be "sqlite3_mathfunc_init". Or if the filename is "./SpellFixExt.dll" then the entry point would be called "sqlite3_spellfixext_init".

Should the macro take optional parameter to override the extension name (in the spirit of lower entry barrier?).  Otherwise each user would have to read this doc and decide the name of their extension.

asg017:
One ergonomic option, in [`sqlite-loadable-rs`](https://github.com/asg017/sqlite-loadable-rs) we have a wrapper around sqlite extension entrypoints like so:

```rust
#[sqlite_entrypoint]
pub fn sqlite3_xyz_init(db: *mut sqlite3) -> Result<()> {
    define_scalar_function(db, "xyz", 2, xyz, FunctionFlags::DETERMINISTIC)?;
    Ok(())
}
```

The APIs are quite different, but using a `sqlite_extension` proc macro is been quite nice. It infers the entrypoint name based on the function (in this case `sqlite3_xyz_init`), which users can just use `sqlite3_extension_init` to match SQLite's default. 

`sqlite-loadable-rs` is a bit messy and it might be hard to port the code over directly, but feel free to steal the API design if you like it! Under the hood it just generates the correct `pub unsafe extern "C" fn` entrypoint code.

nyurik:
@gwenn are there any changes you want me to do on this one?

gwenn:
See #1494

t-kalinowski:
Just passing by; I'm starting to look into writing a SQLite extension and came across this PR.  Seems like it's a nice improvement and ready to merge. Did this just stall out, or was there another reason it wasn't merged?

gwenn:
@t-kalinowski Could you please explain why
https://github.com/rusqlite/rusqlite/blob/master/examples/loadable_extension.rs#L21-L29
is too much ?
Considering that we cannot enforce the `sqlite3_extension_init` name.

t-kalinowski:
Not too much, I'm just starting to look into it. The example seems fine at first impression, was just curious about this PR status. Should this PR be closed?

nyurik:
I will merge in the main branch in a sec to make sure it is up to date

t-kalinowski:
I'm really not trying to rush anything; just as I'm looking through the docs I wanted to check if the example extension will soon be stale because of this PR. I don't have enough context to be able to say anything about whether this PR should be merged or not.

nyurik:
I made a few suggestions in the #1529 -- I think that should close this issue.  Thx for all the feedback!

PR #1429, opened 2023-12-24
Remove modern_sqlite and vtab from CI in loadable ext

nyurik:
This matches the recent changes in the example docs https://github.com/rusqlite/rusqlite/pull/1426


gwenn:
Thanks.

Issue #1211, opened 2022-08-09
Feature request: libsqlite-sys built-in sqlite math functions


evbo:
The latest versions of SQLite come with basic math functions included by default. For instance, if you compile a sqlite3 binary manually, on Linux you have access to functions like `exp` and `ln`, which provide the only built-in mechanism for calculating a cumulative product.

Here are steps for how to currently include math functions when building a rusqlite project:

```bash
but no matter what it resolves `3.31` with `bundle` mode off, which is lower than even my system version.

@zuurr#9735 I think there might be a bug in `SQLITE3_LIB_DIR` env var. In order to use the sqlite i build it's also necessary to prepend its path to `LD_LIBRARY_PATH`. 

Here's how I *successfully* get the math functions, instead of `bundled` I do:

```bash
wget https://www.sqlite.org/2022/sqlite-autoconf-3390200.tar.gz
tar -xvf sqlite-autoconf-3390200.tar.gz
cd sqlite-autoconf-3390200
mkdir build && cd build
../configure # If you want static SQLite only, add --enable-shared=no
make -j6 && su -c "make install"

# necessary for rusqlite to detect your compiled libraries
export LD_LIBRARY_PATH="/path/to/sqlite/shared/libs:$LD_LIBRARY_PATH"
```

Then set the documented env var in cargo.toml:
> [env]
> SQLITE3_LIB_DIR = "/path/to/sqlite/shared/libs"

It would be great if the above steps were automatically performed with a `bundled` build of rusqlite, since by default these math functions should be included in `3.39.2` version of SQLite. Unless they are intentionally removed (for smaller disk footprint?)?

Thanks for considering this enhancement!

gwenn:
https://sqlite.org/lang_mathfunc.html
> The math functions shown below are part of the [SQLite amalgamation source file](https://sqlite.org/amalgamation.html) but are only active if the amalgamation is compiled using the [-DSQLITE_ENABLE_MATH_FUNCTIONS](https://sqlite.org/compile.html#enable_math_functions) compile-time option.

So you should try [LIBSQLITE3_FLAGS]https://github.com/rusqlite/rusqlite/blob/d6780ece27fdd127394aaff17743054984f731ad/libsqlite3-sys/build.rs#L273).

Something like
```sh
LIBSQLITE3_FLAGS="-DSQLITE_ENABLE_MATH_FUNCTIONS" cargo build --feature bundled
```

evbo:
[...]

PR #1437, opened 2024-01-16
Enable built-in SQL math functions when building bundled SQLite

itsxaos:
I noticed that I lost the built-in modulo function as I switched to the bundled SQLite and I tracked it down to `SQLITE_ENABLE_MATH_FUNCTIONS` not being set when building it.

The built-in math functions (i.e. being able to do `SELECT mod(5, 3)`) are described here: https://www.sqlite.org/lang_mathfunc.html, and the compile-time option here: https://www.sqlite.org/compile.html#enable_math_functions.

I don't think there are any compability concerns with enabling these because:
a) per https://www.sqlite.org/compile.html#enable_math_functions it's enabled automatically by the configure script for Unix and windows (so it's probably more common for it to be enabled than disabled)
b) it's not an error to use `Connection::create_scalar_function` to overwrite these built-in functions (I tested this, but not very extensively).

As far as I can tell this would only be an issue for someone if they somehow relied on those functions *not* being defined, which seems unlikely.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -128,6 +128,7 @@ mod build_bundled {
             .flag("-DSQLITE_ENABLE_FTS5")
             .flag("-DSQLITE_ENABLE_JSON1")
             .flag("-DSQLITE_ENABLE_LOAD_EXTENSION=1")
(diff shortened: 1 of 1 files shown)


itsxaos:
Whoops, seems I overlooked #1211...
Feel free to close if your opinion hasn't changed!

PR #1441, opened 2024-01-25
Expose the include directory of the bundled sqlite version

weiznich:
This commit exposes the include directory of the bundled sqlite version via by instructing cargo to set the `DEP_SQLITE3_INCLUDE` enviroment variable as documented in the cargo book:
https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate

This enables other sys crates to rely on the bundled sqlite version and use it as base for their compilation if they have a (sub-)dependency on libsqlite3.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -114,6 +114,7 @@ mod build_bundled {
         {
             super::copy_bindings(lib_name, "bindgen_bundled_version", out_path);
         }
+        println!("cargo:include={}/{lib_name}", env!("CARGO_MANIFEST_DIR"));
         println!("cargo:rerun-if-changed={lib_name}/sqlite3.c");
         println!("cargo:rerun-if-changed=sqlite3/wasm32-wasi-vfs.c");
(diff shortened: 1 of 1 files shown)


gwenn:
https://github.com/search?q=lang%3Atoml+%22cargo%3Ainclude%3D%22&type=code
No one is using this (yet) ?

weiznich:
As that's a feature of `build.rs` files you would need to search for `lang:rust` and not for `lang:toml`, right? 

https://github.com/search?q=lang%3ARust++%22cargo%3Ainclude%3D%22&type=code

Or is that trying to say: Noone is using libsqlite3-sys in that way yet?

gwenn:
> As that's a feature of `build.rs` files you would need to search for `lang:rust` and not for `lang:toml`, right?

Sorry, you are right.

weiznich:
Thanks for the fast response :heart:

gwenn:
libsqlite3-sys v0.28.0 released

weiznich:
Thanks ❤️

PR #1462, opened 2024-03-03
OwningRows / OwningStatement examples

gwenn:
From https://github.com/rusqlite/rusqlite/issues/1337#issuecomment-1858900665
And https://github.com/rusqlite/rusqlite/discussions/1198#discussioncomment-2989869


gwenn:
Maybe for Statement we can let user have their own implementation of the cache.
(with a nice side effect: `hashlink` dependency becomes optional)
And if we let user retrieve the cache key (possibly a simple index), a cached statement can be retrieved with O(1) versus the current O(nlog(n)).

Cargo.toml
```diff
@@ -145,6 +145,8 @@ tempfile = "3.1.0"
 regex = "1.5.5"
 uuid = { version = "1.0", features = ["v4"] }
 unicase = "2.6.0"
+ouroboros = "0.18"
+self_cell = { git = "https://github.com/Voultapher/self_cell.git" }
```
gwenn, line 149: FIXME

PR #1486, opened 2024-03-30
Add preupdate hook

aschey:
This is a continuation of the [previous PR](https://github.com/rusqlite/rusqlite/pull/897) by @MidasLamb (thank you!)

Per feedback on the original proposal, this adds a new feature flag (`preupdate_hook`) which will enable the preupdate API. I moved the hooks implementation into its own folder so the preupdate hook can be easily feature-flagged separately. Let me know if you'd like this to be structured differently and I'd be happy to make any adjustments.


Cargo.toml
```diff
@@ -95,6 +96,7 @@ modern-full = [
     "extra_check",
     "functions",
     "hooks",
+    "preupdate_hook",
```
gwenn, line 99: Please don't touch `modern-full` because most of the time SQLITE_ENABLE_PREUPDATE_HOOK is not activated by default. So if you use `modern-full` and link against the SQLite3 library installed on your OS, the build will probably fail.

src/hooks/preupdate_hook.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// The callback parameters are:
+    ///
+    /// - the name of the database ("main", "temp", ...),
+    /// - the name of the table that is updated,
+    /// - a variant of the PreUpdateCase enum which allows access to extra functions depending
+    /// on whether it's an update, delete or insert.
+    #[inline]
+    pub fn preupdate_hook<'c, F>(&'c self, hook: Option<F>)
+    where
+        F: FnMut(Action, &str, &str, &PreUpdateCase) + Send + 'c,
```
gwenn, line 134: The `'c` (instead of `'static`) seems wrong to me but I am not sure.

src/hooks/preupdate_hook.rs
```diff
@@ -0,0 +1,333 @@
...
+    {
+        self.db.borrow_mut().preupdate_hook(hook);
+    }
+}
+
+impl InnerConnection {
+    #[inline]
+    pub fn remove_preupdate_hook(&mut self) {
```
gwenn, line 142: `remove_preupdate_hook` seems useless / redundant to me.

gwenn:
I am afraid that I should reject this PR because you haven't fork the previous PR #897.
We should see @MidasLamb contributions in git commits.

aschey:
> I am afraid that I should reject this PR because you haven't fork the previous PR #897. We should see @MidasLamb contributions in git commits.

Apologies, thanks for pointing this out. I had merged in the original PR, but I think I accidentally lost the original commit history during a rebase. I wasn't paying attention and missed that. I reapplied the changes on top of the original PR now.

Cargo.toml
aschey: Good to know, thanks! Fixed.

src/hooks/preupdate_hook.rs
aschey: Yes, you're correct. This was a remnant from the original PR where all the other hooks had lifetimes attached. Missed that when I was updating this, thanks!

gwenn:
Must read / check: https://github.com/rusqlite/rusqlite/issues/1048
Nice to have a compilation failure test like:  https://github.com/rusqlite/rusqlite/pull/1052 (without UnwindSafe / RefUnwindSafe)

aschey:
> Must read / check: #1048 Nice to have a compilation failure test like: #1052 (without UnwindSafe / RefUnwindSafe)

Got it, thanks. It needs an explicit `'static` lifetime on the callback. Fixed that and added a complication failure test as mentioned.

aschey:
@gwenn I just realized that the `preupdate_hook` features would never get tested in CI since it's not automatically enabled by `bundled-full`. I updated the github actions config to also enable it wherever it currently enables `bundled-full`, `session`, and `buildtime_bindgen`. Let me know if you'd like to handle this differently.

gwenn:
Thanks

PR #1502, opened 2024-05-13
fix: build should support lib64 dir for OpenSSL3.x

southorange0929:


libsqlite3-sys/build.rs
```diff
@@ -167,7 +167,14 @@ mod build_bundled {
                         }
                     }
                     Some(openssl_dir) => {
-                        let lib_dir = lib_dir.unwrap_or_else(|| openssl_dir.join("lib"));
+                        let lib_dir = lib_dir.unwrap_or_else(|| {
+                            // OpenSSL 3.0 now puts it's libraries in lib64/ by default,
+                            // check for both it and lib/.
+                            if openssl_dir.join("lib64").exists() {
+                                return openssl_dir.join("lib64");
```
gwenn, line 174: Should depend on `target`, no ?

southorange0929: I think not, for some architectures, OpenSSL version 3.x is installed under the lib64 directory.

This approach is also in sync with [rust-openssl](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/find_normal.rs#L20).

gwenn: But `rust-openssl` uses both `lib64` **and** `lib`.

southorange0929: Yes, I've now fully aligned its logic with that of openssl.

gwenn: Thanks.

southorange0929:
Hi how about to merge this pr? Thanks

gwenn:
Thanks.

gwenn:
libsqlite3-sys v0.29.0 released

PR #1505, opened 2024-05-17
Fix csvtab doc example

llamda:
Fixes ```SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("file ' 'my_csv.csv'' does not exist"))``` when running the example code. 
(It looks for a file name with a leading space currently)


gwenn:
Thanks for the issue report.
But we should fix the module instead.
See original / official impl: https://github.com/sqlite/sqlite/blob/master/ext/misc/csv.c#L413

PR #1518, opened 2024-06-28
 Demonstrates retrieving function list

pchemguy:


gwenn:
[Table-valued functions](https://sqlite.org/vtab.html#tabfunc2) are documented by SQLite and can be used like SQL table and listed [here](https://sqlite.org/vtablist.html).
So there is nothing specific to `rusqlite` here...
https://github.com/rusqlite/rusqlite/blob/eebbbb008633b36f2580b29501800fb8bc37e36a/src/pragma.rs#L388-L396

PR #1522, opened 2024-06-28
An example retrieving SQLite version via the dedicated API

pchemguy:


gwenn:
As it is just a function call, there is no example needed.

PR #1523, opened 2024-06-28
Demonstrates reading and updating SQLite ids

pchemguy:


gwenn:
https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.pragma_update should be used when new value is dynamic.
Like https://github.com/rusqlite/rusqlite/blob/eebbbb008633b36f2580b29501800fb8bc37e36a/src/pragma.rs#L404
And https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.pragma_query_value is also convenient.
Like https://github.com/rusqlite/rusqlite/blob/eebbbb008633b36f2580b29501800fb8bc37e36a/src/statement.rs#L1322 and  https://github.com/rusqlite/rusqlite/blob/eebbbb008633b36f2580b29501800fb8bc37e36a/src/pragma.rs#L334

pchemguy:
And I would learn about it if there was an example in the examples directory, which was not.

gwenn:
https://docs.rs/rusqlite/latest/rusqlite/?search=pragma
+
https://docs.rs/rusqlite/latest/src/rusqlite/pragma.rs.html#175-187
+
Ctrl-F `pragma_query_value`

pchemguy:
And I am supposed to know that I need to search for "pragma_query_value" how exactly!?

gwenn:
Nope, only `pragma` which  gives you only 5 methods in the result...
Among these 5 methods, there are `pragma_query_value` and `pragma_update`.
Which means 2 / 5 are relevant.

pchemguy:
I am sorry, but I do not understand you. What I meant is that to search for the "pragma_query_value" string, which has nothing to do with SQLite, I needed to know about it in the first place, which does not quite makes sense. But I have no clue what you tried to say in your last comment.

gwenn:
> But I have no clue what you tried to say in your last comment.

https://docs.rs/rusqlite/latest/rusqlite/?search=pragma

PR #1524, opened 2024-06-28
Minimalistic SQLite Version statement example

pchemguy:
Retrieves SQLite version number via a SELECT statement


gwenn:
https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.query_row should be used in such as case.
Which already has an example...

PR #1525, opened 2024-06-28
An example retrieving SQLite version number via the dedicated API

pchemguy:


gwenn:
https://docs.rs/rusqlite/latest/rusqlite/fn.version_number.html
As it is just a function call, there is no example needed.

PR #1529, opened 2024-07-16
Improve loadable extension docs

nyurik:
A minor documentation improvement on a loadable extension - moving loadable instructions to the whole example rather than the init function


nyurik:
The CI is failing here until #1530

PR #1530, opened 2024-07-16
Minor clippy fix

nyurik:


libsqlite3-sys/build.rs
```diff
@@ -506,6 +507,7 @@ mod bindings {
 
     static PREBUILT_BINDGENS: &[&str] = &["bindgen_3.14.0"];
 
+    #[allow(clippy::needless_pass_by_value)]
```
nyurik, line 510: note that this is actually a bug in Clippy - I just fixed it there, hopefully will be merged and become available in a few months:  https://github.com/rust-lang/rust-clippy/pull/13113

thomcc:
@gwenn what's the issue with this?

gwenn:
There is no such clippy warnings (at least neither with stable (see https://github.com/rusqlite/rusqlite/actions/runs/10122368512/job/27994429531) nor with nightly).
And there are unrelated changes like [main.yml](https://github.com/rusqlite/rusqlite/pull/1530/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3).

nyurik:
@gwenn I used pedantic to find these. Some are probably less important than others. Not cloning token stream is probably a small perf optimization that may speed up some builds. I could split it into multiple PRs if you want

nyurik:
I removed CI changes, now this PR only has a few simply clippy lints

gwenn:
This warnings don't seem to be raised by Github action workflow or locally.

nyurik:
should they be added to the `[lints.clippy]` in Cargo.toml?  Without the minimum rust version, it is hard to say which features are ok to use

gwenn:
The default Clippy lints seem ok, no ?

nyurik:
the defaults pass, yes.  I am actually not certain which lint suggested this one, but clearly a good change. Also, one missing semicolon.

```diff
-Duration::try_milliseconds(789).unwrap()
+Duration::milliseconds(789)
```

PR #1531, opened 2024-07-16
Applied some spellchecker suggestions

nyurik:


nyurik:
The CI is failing here until #1530

src/column.rs
```diff
@@ -60,7 +60,7 @@ impl Statement<'_> {
     /// https://www.sqlite.org/c3ref/column_name.html
     /// > The returned string pointer is valid...
     ///
-    /// `column_name` reference can become invalid if `stmt` is reprepared
+    /// `column_name` reference can become invalid if `stmt` is re-prepared
```
gwenn, line 63: https://sqlite.org/search?s=d&q=reprepared

src/hooks/mod.rs
```diff
@@ -54,7 +54,7 @@ pub struct AuthContext<'c> {
     /// The database name, if applicable.
     pub database_name: Option<&'c str>,
 
-    /// The inner-most trigger or view responsible for the access attempt.
+    /// The innermost trigger or view responsible for the access attempt.
```
gwenn, line 57: https://sqlite.org/search?s=d&q=inner-most

nyurik: Are you implying that SQLite docs should be updated as well, or that they introduced a new keyword with a new meaning?  https://www.merriam-webster.com/dictionary/innermost -- is this the meaning?

nyurik:
@gwenn I am happy to adjust my PR - do you want to just keep these as is despite being somewhat non-standard?  Need some feedback :)

gwenn:
@nyurik Why not just use the same terms as the ones used by SQLite documentation ?
https://en.wiktionary.org/wiki/innermost
> equivalent to inner +‎ -most.

nyurik:
ok, will revert

gwenn:
Thanks.

PR #1532, opened 2024-07-17
Allow setting default connection transaction behavior

seddonm1:
Hi,
It would be nice if a per connection default `transaction_behavior` could be set instead of having to use `transaction_with_behavior` everywhere. 

Then, for example:
- a connection of writing could be set `conn.set_transaction_behavior(TransactionBehavior::Immediate)`
- read-connections would be `conn.set_transaction_behavior(TransactionBehavior::Deferred)` 

This PR adds this behavior with the default value remaining unchanged as `TransactionBehavior::Deferred`.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -1,18 +1,10 @@
 use crate::{Connection, Result};
 use std::ops::Deref;
 
-impl Connection {
-    /// Set the default transaction behavior for the connection.
-    pub fn set_transaction_behavior(&mut self, behavior: TransactionBehavior) {
-        self.transaction_behavior = behavior;
-    }
-}
-
 /// Options for transaction behavior. See [BEGIN
 /// TRANSACTION](http://www.sqlite.org/lang_transaction.html) for details.
 #[derive(Copy, Clone)]
 #[non_exhaustive]
-
 pub enum TransactionBehavior {
     /// DEFERRED means that the transaction does not actually start until the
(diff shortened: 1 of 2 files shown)


src/transaction.rs
```diff
@@ -1,10 +1,18 @@
...
+    pub fn set_transaction_behavior(&mut self, behavior: TransactionBehavior) {
+        self.transaction_behavior = behavior;
+    }
+}
+
 /// Options for transaction behavior. See [BEGIN
 /// TRANSACTION](http://www.sqlite.org/lang_transaction.html) for details.
 #[derive(Copy, Clone)]
 #[non_exhaustive]
+
```
gwenn, line 15: Oops !

gwenn:
At least, we should document that this flag / field is ignored by manual / direct transactions:
```rust
let mut conn = Connection::open_in_memory()?;
conn.set_transaction_behavior(TransactionBehavior::Immediate);
conn.execute_batch("BEGIN")?; // deferred
...
```

src/transaction.rs
seddonm1: Wow. Sorry.

seddonm1:
Thanks for the very quick review @gwenn and thanks for being open to this new feature.

I have refactored the code based on your comments and add an example so it is consistent with the rest of the code.

gwenn:
Thanks.

PR #1558, opened 2024-08-26
`impl FromSql` for various heap-allocated string and blob slices

H2CO3:
Just what it says on the tin. This adds `impl FromSql for`:

* `Cow<'_, T> where T: ToOwned + FromSql`
* `Rc<str>`
* `Arc<str>`
* `Box<[u8]>`
* `Rc<[u8]>`
* `Arc<[u8]>`

Unit tests are included and passing. This doesn't seem to break anything used by the current test suite, and adding concrete `impl`s shouldn't usually be a breaking change, so it seems we're good to go.

### DIFF
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -1,5 +1,6 @@
 use super::{Value, ValueRef};
 use std::error::Error;
+use std::borrow::Cow;
 use std::fmt;
 
 /// Enum listing possible errors from [`FromSql`] trait.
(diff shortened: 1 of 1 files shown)


gwenn:
I don't konw why postgres doesn't support these types:
https://docs.rs/postgres-types/latest/postgres_types/trait.FromSql.html
?

H2CO3:
> I don't konw why postgres doesn't support these types: https://docs.rs/postgres-types/latest/postgres_types/trait.FromSql.html ?

No idea, to be honest. Seems like an omission. They don't seem to break anything, and I can't think of anything obvious why they shouldn't be supported, either. (And I can give good arguments as to why they should.)

gwenn:
@thomcc Would you mind giving your review ?

thomcc (approved):
Looks like a good change to me.

H2CO3:
@thomcc @gwenn Sorry, I have zero idea how to please `rustfmt` (I don't use it). Last time it failed because lines were too long, now it fails because they are too short. 😭 Can you help me format the code correctly? I don't think I'll be able to figure it out in a reasonable time, given that every change apparently requires a manual re-run on your part.

thomcc:
You should just run `cargo fmt` in the repository, and it will format it correctly for you.

H2CO3:
@thomcc Done, albeit the formatting of the getter closures became very… peculiar, to put it mildly.

gwenn:
Thanks!

PR #1569, opened 2024-09-24
Fixed pre-release `wasm32-wasip(\d)` targets not enabling wasi compile flags in `bundled` mode.

Earthmark:
The `wasm32-wasi` target is being deprecated in favor of pre-release targets like `wasm32-wasip(\d)`.

This instead matches the prefix wasi target, so any future pre-releases will probably, maybe, work out of the box.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -239,7 +239,7 @@ mod build_bundled {
         if !win_target() {
             cfg.flag("-DHAVE_LOCALTIME_R");
         }
-        if env::var("TARGET").map_or(false, |v| v == "wasm32-wasi") {
+        if env::var("TARGET").map_or(false, |v| v.starts_with("wasm32-wasi")) {
             cfg.flag("-USQLITE_THREADSAFE")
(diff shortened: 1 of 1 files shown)


Earthmark:
`wasm32-wasip1-threads` may need a more complicated solution, but this at least works for `wasm32-wasip1` and `wasm32-wasip2`.

gwenn:
@thomcc If possible, could you please help me validate this PR (because I know nothing about wasm) ?

Earthmark:
As a note on this, wasm32-wasi was supported previously, that target was renamed to wasm32-wasip1, which is what caused the build issue. The target itself is the same to my knowledge.

For that build target at least, this would be identical. There is p2 on nightly, but my goal here was getting the renamed version of p1 building.

gwenn:
Sorry for the delay.
Looks like https://github.com/gyscos/zstd-rs/commit/fc3a73100b6d3f2ff6dd3da0ff14dd9185ba02f0#diff-d10afeb74f1c422688106d5e75cd08ac2ac0328ef71bb829a5f127f1a105e1f6R138

PR #1583, opened 2024-10-18
Fix Batch impl

### LINKED ISSUE #1581: `Batch` iterator gets stuck on syntax error
While building [Hiqlite](https://github.com/sebadob/hiqlite) I stumbled about an issue with the `Batch` iterator / stream.

When I consume some sql `String` using `Batch::new()` and then iterate over the data, I end up in an endless loop if one of the statements has a syntax error. The iterator kind of "gets stuck" and with each follwing `.next()`, it will return the same syntax error over and over again until end of time, or actually until the process is OOM killed.

My solution for now is to simply `panic!` on any syntax errors, which is bad but at least way better than consuming all memory and getting killed later. The `panic` is on purpose to spot errors early.

A short example:

```rust
let mut batch = Batch::new(&conn, req.sql.as_ref());
// we can at least assume 2 statements in a batch execute
let mut res = Vec::with_capacity(2);

loop {
    match batch.next() {
        Ok(Some(mut stmt)) => {
            res.push(stmt.execute([]).map_err(Error::from));
        }
        Ok(None) => break,
        Err(err) => {
            // TODO if there is a syntax error in the batch stmt, this will loop forever
            // -> Err(_) will NOT remove the bad element from the Batch
            panic!("{}", err);
            // res.push(Err(Error::from(err)));
        }
    }
}
```

I would expect the stream / iterator to remove an item on error, or is this simply not possible because of the way it's built internally, and I would need to exit, return an error early and drop all following statements? If so, an update to the docs would be very helpful mentioning, that the iterator can never finish, if there is any error.

gwenn:
- document that there is no parsing error recovery
- and remove `Iterator` impl to prevent users from looping indefinitely on Some(Err(..))

Fix #1581

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1141,7 +1141,6 @@ impl<'conn> fallible_iterator::FallibleIterator for Batch<'conn, '_> {
     /// Iterates on each batch statements.
     ///
     /// Returns `Ok(None)` when batch is completed.
-    #[expect(clippy::should_implement_trait)] // fallible iterator
     fn next(&mut self) -> Result<Option<Statement<'conn>>> {
         while self.tail < self.sql.len() {
             let sql = &self.sql[self.tail..];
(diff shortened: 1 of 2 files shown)


gwenn:
Version 0.33.0 released

Issue #1570, opened 2024-09-24
Add safe exposure of `sqlite3_wal_hook` 


vincent-herlemont:
I'd like to propose adding support for the [`sqlite3_wal_hook`](https://www.sqlite.org/c3ref/wal_hook.html) in `rusqlite`, inspired by the existing [`commit_hook`](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.commit_hook) and [`rollback_hook`](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.rollback_hook) etc. This would allow users to monitor Write-Ahead Logging (WAL) events via a safe Rust interface.

**Proposed Changes:**
1. **Feature Flag:**
   - Introduce a feature flag (e.g., `"wal_hook"` or added with the `"hooks"` feature ) to keep this functionality optional.
2. **Safe Callback Interface:**
   - Similar to `commit_hook`, provide a safe Rust closure interface for the WAL hook, ensuring it adheres to Rust's borrowing and ownership rules.
   - Avoid direct exposure of raw pointers, offering a safe abstraction instead.

gwenn:
You should be able to reuse the existing `hooks` feature because sqlite3_wal_hook was introduced before 3.14 (https://github.com/rusqlite/rusqlite/blob/95680270eca6f405fb51f5fbe6a214aac5fdce58/libsqlite3-sys/sqlite3/sqlite3ext.h#L233).

gwenn:
http://sqlite.org/c3ref/wal_hook.html
> Note that the sqlite3_wal_autocheckpoint() interface and the wal_autocheckpoint pragma both invoke sqlite3_wal_hook() and will overwrite any prior sqlite3_wal_hook() settings.

So it seems not possible to use a closure / `Fn` but only a `fn` (like [busy_handler](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.busy_handler)).

gwenn:
Version 0.33.0 released

PR #1594, opened 2024-11-10
Add safe binding to sqlite3_wal_hook

### LINKED ISSUE #1187: Support WAL hook
[`sqlite3_wal_hook`](https://www.sqlite.org/c3ref/wal_hook.html) is a hook which rusqlite currently doesn't have bindings for; it's useful if you want fine-grained control over the WAL mechanism and want to know when to trigger it. It'd be nice if rusqlite had a convenient binding for it like it does for commit hooks etc.

gwenn:
Fix #1187 and #1570

### DIFF
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -7,7 +7,7 @@ use std::ptr;
 
 use crate::ffi;
 
-use crate::{Connection, InnerConnection};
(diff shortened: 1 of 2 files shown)


gwenn:
TODO: give access to the `sqlite3` handle / `&Connection`.
But creating a `Connection` / `StatementCache` might be overkill for just calling `sqlite3_wal_checkpoint`

PR #1599, opened 2024-11-16
Switch deprecated target wasm32-wasi to wasm32-wasip1

wasm-forge:
The rust target `wasm-wasi` is being replaced with more specific `wasm32-wasip1` target and `wams32-wasi` will be deprecated starting 2025 ([https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html)).

This change adds `wasm32-wasip1` to the build rules.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -239,7 +239,9 @@ mod build_bundled {
         if !win_target() {
             cfg.flag("-DHAVE_LOCALTIME_R");
         }
-        if env::var("TARGET").map_or(false, |v| v.starts_with("wasm32-wasi")) {
+        if env::var("TARGET").map_or(false, |v| {
+            v.starts_with("wasm32-wasi") || v.starts_with("wasm32-wasip1")
+        }) {
             cfg.flag("-USQLITE_THREADSAFE")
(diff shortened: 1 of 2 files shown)


libsqlite3-sys/build.rs
```diff
@@ -239,7 +239,9 @@ mod build_bundled {
         if !win_target() {
             cfg.flag("-DHAVE_LOCALTIME_R");
         }
-        if env::var("TARGET").map_or(false, |v| v.starts_with("wasm32-wasi")) {
+        if env::var("TARGET").map_or(false, |v| {
+            v.starts_with("wasm32-wasi") || v.starts_with("wasm32-wasip1")
```
gwenn, line 243: `wasm32-wasip1` starts with `wasm32-wasi` !
See #1569

gwenn (changes requested):

libsqlite3-sys/build.rs
wasm-forge: Oops! Sorry, I got it wrong, and "fixed" the non-existing bug! 
Looking forward to seeing the new release.

PR #1622, opened 2025-01-15
Upgrade to sqlite 3.48

dvic:


gwenn:
Large diffs are suspicious.
And  `SQLITE_PREPARE_DONT_LOG` should be added in lib.rs...

dvic:
> Large diffs are suspicious.

I understand, I've reverted the sqlite changes and just updated `upgrade.sh`

> And `SQLITE_PREPARE_DONT_LOG` should be added in lib.rs...

Done!

gwenn:
See #1623 (no large diffs except for sqlite3.c which is expected)

dvic:
> See #1623 (no large diffs except for sqlite3.c which is expected)

Awesome, thanks 👍!

PR #1629, opened 2025-01-27
Update error type for `ValueRef` methods

stdonnelly:
This corrects the doc comments that explain the error variant returned by `ValueRef` methods.

The current doc comments say that the methods will return `Err(Error::InvalidColumnType)` if the case of `self` is not correct. The `Err` case of the results returned by these methods is actually `FromSqlError`, and specifically `FromSqlError::InvalidType` when the case of `self` is not correct.

I had a quick glance at every line matching `//.*Error::` and `//.*Err\(` and could not find any other comments that seemed incorrect.


gwenn:
You are right.
Thanks.

Issue #1342, opened 2023-06-06
How to pass dynamic named parameters to rusqlite query/execute through `params_from_iter` function?


ethever:
How to pass dynamic named parameters to rusqlite query/execute through `params_from_iter` function?
I saw the `Sealed` trait has been implemented to `Params` trait, but there is not such `bind_parameters_named` method for `ParamsFromIter` structure.

ethever:
How to make the `bind_parameters_named` function to able to accept a `Iterator`?
pub(crate) fn bind_parameters_named<'a, T: 'a + ?Sized, I>(&mut self, params: I) -> Result<()>
    where
        T: ToSql,
        I: IntoIterator<Item = (&'a str, &'a T)>,

gwenn:
There is no `NamedParamsFromIter` for [named parameters](https://docs.rs/rusqlite/latest/rusqlite/trait.Params.html#named-parameters).
As a workaround, you can use [raw_bind_parameter](https://docs.rs/rusqlite/latest/rusqlite/struct.Statement.html#method.raw_bind_parameter)

bwo:
are there any plans to introduce such functionality? It would be very useful.

bwo:
that doesn't mean that a more convenient API wouldn't be appreciated, and in fact I was stymied trying to do just that (I kept getting errors that the size of my slice wasn't known at compile time—obviously, if I knew that, I wouldn't need anything dynamic in the first place).

thomcc:
If you get errors that the size of your slice isn't known at compile time, it usually means you just need to pass it behind a reference.

thomcc:
Or perhaps collect into a vec and use `vec.as_slice()`.

bwo:
Regardless—granted that there are workarounds, the question whether a simple API will be offered is reasonable. Indeed, the docs say:

> As a slice of `&[(&str, &dyn ToSql)]`. This is what essentially all of these boil down to in the end, conceptually at least. In theory you can pass this as `stmt`.

But while "all of these" boil down to that in the end, that isn't the sole API provided. Other, more convenient ones are, too. I take it from your replies that the answer to my question (will a more convenient way to do this be introduced) is "no".

thomcc:
Yeah, I don't really have strong feelings.

PR #1638, opened 2025-02-03
add iterator wrapper for named query parameters

e00E:
This is like ParamsFromIter but with named parameters.

This is useful for the same reasons as ParamsFromIter. Sometimes you cannot use the macro and the Params implementations for arrays and tuples don't work either.

### DIFF
--- a/src/params.rs
+++ b/src/params.rs
@@ -221,7 +221,7 @@ impl Sealed for &[(&str, &dyn ToSql)] {}
 impl Params for &[(&str, &dyn ToSql)] {
     #[inline]
     fn __bind_in(self, stmt: &mut Statement<'_>) -> Result<()> {
-        stmt.bind_parameters_named(self)
+        stmt.bind_parameters_named(self.iter().copied())
     }
 }
 
@@ -301,7 +301,7 @@ macro_rules! impl_for_array_ref {
         impl<T: ToSql + ?Sized> Sealed for &[(&str, &T); $N] {}
         impl<T: ToSql + ?Sized> Params for &[(&str, &T); $N] {
             fn __bind_in(self, stmt: &mut Statement<'_>) -> Result<()> {
-                stmt.bind_parameters_named(self)
+                stmt.bind_parameters_named(self.iter().copied())
             }
         }
         impl<T: ToSql> Sealed for [T; $N] {}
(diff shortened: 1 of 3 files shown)


src/params.rs
```diff
@@ -221,7 +221,7 @@ impl Sealed for &[(&str, &dyn ToSql)] {}
 impl Params for &[(&str, &dyn ToSql)] {
     #[inline]
     fn __bind_in(self, stmt: &mut Statement<'_>) -> Result<()> {
-        stmt.bind_parameters_named(self)
+        stmt.bind_parameters_named(self.iter().copied())
```
gwenn, line 224: Seems bad  to me

gwenn:
https://github.com/rusqlite/rusqlite/pull/908#issuecomment-790797515

gwenn:
https://github.com/rusqlite/rusqlite/issues/1342

src/params.rs
e00E: Why?

e00E:
> And you don't want to use [raw_bind_parameter](https://docs.rs/rusqlite/latest/rusqlite/struct.Statement.html#method.raw_bind_parameter) ?

No. I don't like the caveats that this method comes with.

e00E:
> [#908 (comment)](https://github.com/rusqlite/rusqlite/pull/908#issuecomment-790797515)

That comment is basically this PR. I wasn't aware of it. I added a change to make this PR allow string values too.

src/params.rs
```diff
@@ -451,3 +451,45 @@ where
...
+
+/// Adapter type which allows any iterator over `(&str, ToSql)` values to
+/// implement [`Params`].
+///
+/// This struct is created by the [`params_named_from_iter`] function.
+///
+/// This struct is similar to [`ParamsFromIter`]. See its documentation for more
+/// details. The difference is that this struct works on named parameters. The
+/// iterator yields the `ToSql` parameter and its name.
+pub struct ParamsNamedFromIter<I>(I);
```
e00E, line 463: Do you prefer the name ParamsNamedFromIter or NamedParamsFromIter?

gwenn:
> > And you don't want to use [raw_bind_parameter](https://docs.rs/rusqlite/latest/rusqlite/struct.Statement.html#method.raw_bind_parameter) ?
> 
> No. I don't like the caveats that this method comes with.

Which caveats ?

e00E:
With raw_bind_parameters you cannot pass the parameters as their name. And you have to use raw_query and raw_execute. Isn't it the same reasons as ParamsFromIter? If it makes sense to provide params as an iterator for index based params then it also makes sense for named params with the same reasons.

PR #1641, opened 2025-02-05
Add `pub use` to macros to make them usable to other crates

SciMind2460:
I intend to make the macros `to_sql_self`, `to_sql_self_fallible`, `from_value` and `from_sql_output` usable for other crates who intend their code to be interoperable with `rusqlite`. This means they can use the crate to make new implementations to be able to execute their values.
For example:
```rust
struct SQLType {
   value: i32,
   strings: String
}
to_sql_self!(SQLType)
```
This is needed due to the <a href="https://github.com/rust-lang/rust/issues/30191">Rust issue linked here.</a>
This will also help crate maintainers more easily implement the `ToSql` and `FromSql` traits.

### DIFF
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -107,6 +107,8 @@ macro_rules! from_sql_integral(
     )
 );
 
+pub use from_sql_integral;
+
 from_sql_integral!(i8);
 from_sql_integral!(i16);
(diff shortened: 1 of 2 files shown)


SciMind2460:
I realized the issue with my PR so I closed it, regarding publics

PR #1646, opened 2025-02-11
Deserialize impls

gwenn:
- from static bytes
- from `Read`ables with known size

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -1,6 +1,6 @@
 //! Serialize a database.
 use std::marker::PhantomData;
-use std::ops::{Deref, DerefMut};
+use std::ops::Deref;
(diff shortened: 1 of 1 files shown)


src/serialize.rs
```diff
@@ -95,28 +95,60 @@ impl Connection {
...
 
+    /// Deserialize from stream
+    pub fn deserialize_read(
+        &mut self,
+        schema: DatabaseName<'_>,
+        read: &mut dyn std::io::Read,
+        sz: usize,
+        read_only: bool,
+    ) -> Result<()> {
+        let ptr = unsafe { ffi::sqlite3_malloc(sz.try_into().unwrap()) }.cast::<u8>();
```
gwenn, line 106: check ptr is not null

PR #1664, opened 2025-03-18
added sqlite3_vtab_nochange() and sqlite3_value_nochange(x)

abhay-agarwal:
I noticed that these methods were stubbed out, so I added them at very low cost. Perhaps I misunderstood why they were stubbed but if not, here they are.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -81,3 +81,9 @@ pub(super) unsafe fn set_result(
         }
     }
 }
+
+#[inline]
+#[must_use]
+pub(super) unsafe fn no_change(ctx: *mut sqlite3_context) -> bool {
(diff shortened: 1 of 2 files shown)


gwenn:
(1) You cannot assume that `sqlite3_vtab_nochange` is available because rusqlite use 3.14 API by default (see CI build error)
(2) Ideally, we should make sure that `sqlite3_vtab_nochange` is used only from `xColumn`
(3) `src/context.rs` is used only to share common code between `functions` and `vtab` which is clearly not the case of `sqlite3_vtab_nochange`
(4) Ideally, we should make sure that `sqlite3_value_nochange` is used only from `xUpdate`.

PR #1672, opened 2025-04-15
columns_with_metadata: access column metadata from prepared statement

jgbyrne:
A simple binding for the column metadata functions:

sqlite3_column_database_name
sqlite3_column_table_name
sqlite3_column_origin_name

No docs or tests but I am happy to add them if you are interested in the PR.

Cheers,
Jack

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -3,7 +3,7 @@ use std::str;
 use crate::{Error, Result, Statement};
 
 /// Information about a column of a SQLite query.
-#[cfg(feature = "column_decltype")]
+#[cfg(any(feature = "column_decltype", feature = "column_metadata"))]
 #[cfg_attr(docsrs, doc(cfg(feature = "column_decltype")))]
 #[derive(Debug)]
 pub struct Column<'stmt> {
@@ -11,7 +11,7 @@ pub struct Column<'stmt> {
     decl_type: Option<&'stmt str>,
 }
 
-#[cfg(feature = "column_decltype")]
+#[cfg(any(feature = "column_decltype", feature = "column_metadata"))]
 #[cfg_attr(docsrs, doc(cfg(feature = "column_decltype")))]
 impl Column<'_> {
     /// Returns the name of the column.
@@ -26,6 +26,42 @@ impl Column<'_> {
     #[must_use]
     pub fn decl_type(&self) -> Option<&str> {
         self.decl_type
+    }
+}
+
+#[cfg(feature = "column_metadata")]
+#[derive(Debug)]
+pub struct ColumnMetadata<'stmt> {
+    name: &'stmt str,
+    database_name: Option<&'stmt str>,
+    table_name: Option<&'stmt str>,
+    origin_name: Option<&'stmt str>,
+}
+
+#[cfg(feature = "column_metadata")]
+impl ColumnMetadata<'_> {
+    #[inline]
+    #[must_use]
+    pub fn name(&self) -> &str {
+        self.name
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn database_name(&self) -> Option<&str> {
+        self.database_name
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn table_name(&self) -> Option<&str> {
+        self.table_name
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn origin_name(&self) -> Option<&str> {
+        self.origin_name
     }
 }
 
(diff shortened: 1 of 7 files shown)


gwenn:
http://sqlite.org/c3ref/column_database_name.html
> These APIs are only available if the library was compiled with the [SQLITE_ENABLE_COLUMN_METADATA](http://sqlite.org/compile.html#enable_column_metadata) C-preprocessor symbol.

jgbyrne:
Yeah this was the reason I gated the impl behind `column_metadata` just like is done for `column_decltype`

gwenn:
> Yeah this was the reason I gated the impl behind `column_metadata` just like is done for `column_decltype`

http://sqlite.org/c3ref/column_decltype.html
https://sqlite.org/compile.html#_options_to_omit_features
> SQLITE_OMIT_DECLTYPE
> 
> This option causes SQLite to omit support for the [sqlite3_column_decltype()](https://sqlite.org/c3ref/column_decltype.html) and [sqlite3_column_decltype16()](https://sqlite.org/c3ref/column_decltype.html) interfaces.

=> `sqlite3_column_decltype` is visible by default...

jgbyrne:
I'm pretty sure most (all?) major distros build with this flag on even if it's technically an optional extra.

jgbyrne:
If you don't want to merge it that's fine :p

gwenn:
> If you don't want to merge it that's fine :p

if you don't want to make it mergeable that's funny :p

jgbyrne:
Sorry I'm not trying to be glib, I just don't see what you want me to do here. Dynamically check whether the methods are present in the sqlite binary?

gwenn:
You need to activate the `SQLITE_ENABLE_COLUMN_METADATA` flag:
Like here:
https://github.com/rusqlite/rusqlite/blob/755bfa384a7a35f5aa817cc7fd5b76f4858ae25a/Cargo.toml#L59
https://github.com/rusqlite/rusqlite/blob/755bfa384a7a35f5aa817cc7fd5b76f4858ae25a/libsqlite3-sys/Cargo.toml#L28
https://github.com/rusqlite/rusqlite/blob/755bfa384a7a35f5aa817cc7fd5b76f4858ae25a/libsqlite3-sys/build.rs#L256-L258

src/column.rs
```diff
@@ -3,15 +3,15 @@ use std::str;
...
+#[cfg(any(feature = "column_decltype", feature = "column_metadata"))]
 #[cfg_attr(docsrs, doc(cfg(feature = "column_decltype")))]
 #[derive(Debug)]
 pub struct Column<'stmt> {
     name: &'stmt str,
     decl_type: Option<&'stmt str>,
 }
 
-#[cfg(feature = "column_decltype")]
+#[cfg(any(feature = "column_decltype", feature = "column_metadata"))]
```
gwenn, line 14: Are you sure ? You don't seem to use `Column`.

src/column.rs
```diff
@@ -3,15 +3,15 @@ use std::str;
 use crate::{Error, Result, Statement};
 
 /// Information about a column of a SQLite query.
-#[cfg(feature = "column_decltype")]
+#[cfg(any(feature = "column_decltype", feature = "column_metadata"))]
```
gwenn, line 6: Are you sure ? You don't seem to use `Column`.

src/column.rs
```diff
@@ -192,6 +256,39 @@ mod test {
...
+    #[cfg(feature = "column_metadata")]
+    fn test_columns_with_metadata() -> Result<()> {
+        use super::Column;
+
+        let db = Connection::open_in_memory()?;
+        let query = db.prepare("SELECT * FROM sqlite_master")?;
+
+        let col_mets = query.columns_with_metadata();
+        for col in &col_mets {
+            assert_eq!(&col.database_name.map(str::to_lowercase), &Some("main".to_owned()));
```
gwenn, line 269: Why `to_lowercase` ?

src/column.rs
```diff
@@ -192,6 +256,39 @@ mod test {
...
+            assert_eq!(&col.table_name.map(str::to_lowercase), &Some("sqlite_master".to_owned()));
```
gwenn, line 270: Why `to_lowercase` ?

src/column.rs
```diff
@@ -192,6 +256,39 @@ mod test {
...
+        }
+
+        let col_origins: Vec<Option<String>> = col_mets
+            .iter()
+            .map(|col| col.origin_name.map(str::to_lowercase))
```
gwenn, line 275: Why `to_lowercase` ?

src/column.rs
jgbyrne: The reason for `to_lowercase` here and in the other places is because I believe that when working with sqlite db / table / column names you should be case insensitive, right?

gwenn:
We might be able to have better coverage by using accessors instead of field.
And maybe by adding a dummy column like:
```sql
SELECT *, 1 FROM sqlite_schema;
```
(any expression should work)

gwenn:
Thanks.

gwenn:
Oops: `ColumnMetadata` is not public (outside the crate) !
(Sorry, I forget to check that)

jgbyrne:
#1675 1675

Issue #921, opened 2021-03-20
ToSql not implemented for rusqlite::types::ValueRef


bamilab:
Well, the title says it.
This makes is impossible to use `ValueRef` as parameters for `Statement::execute` and related functions.

bamilab:
I have fixed this for myself locally, will create a pull request soon.

chbug:
Hey, was this every pulled? I don't see it in the current rusqlite version. Thanks!

Issue #295, opened 2017-09-15
Implement `ToSql` for `ValueRef`?


jD91mZM2:
Doing `&[&false, "test"]` results in a mismatched type, since arrays can only be of one type. You can, however, get away with `&[Value::Integer(if false { 1 } else { 0 }), Value::Text("text".to_string())]`.
But as you can see,

> to_string

Cloning alert!

Any chance to implement ValueRef too? So `&[ValueRef::Integer(if false { 1 } else { 0 }), ValueRef::Text("string")]`?

jD91mZM2:
I as mistaken, apparently you can in fact use multiple variable types without a mismatched type problem. I thought I tried that, but apparently not.

PR #1674, opened 2025-04-16
Impl ToSql for ValueRef

GuillerLT:
Looks like it has been requested in the past in #295 and #921.

### DIFF
--- a/src/types/to_sql.rs
+++ b/src/types/to_sql.rs
@@ -292,6 +292,13 @@ impl ToSql for Value {
     }
 }
 
+impl ToSql for ValueRef<'_> {
+    #[inline]
+    fn to_sql(&self) -> Result<ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::Borrowed(*self))
+    }
+}
+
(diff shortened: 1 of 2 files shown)


gwenn:
Impl block should be moved in `to_sql.rs`.
And a test, showing why / when we need this, would be nice.

GuillerLT:
`Value` already implemented ToSql prior to this pull request but I don't think there were any tests so I included them in the new tests, let me know if that is not ok.
https://github.com/rusqlite/rusqlite/blob/755bfa384a7a35f5aa817cc7fd5b76f4858ae25a/src/types/to_sql.rs#L288

gwenn:
> `Value` already implemented ToSql prior to this pull request but I don't think there were any tests:

Codecov seems to suggest that you are wrong:
https://app.codecov.io/github/rusqlite/rusqlite/commit/755bfa384a7a35f5aa817cc7fd5b76f4858ae25a/blob/src/types/to_sql.rs?dropdown=coverage#L290

gwenn:
And your test doesn't comply to:
> And a test, showing **why / when we need this**, would be nice.

PR #1676, opened 2025-04-17
types: add support for Jiff's `Timestamp`

BurntSushi:
This is probably the most important type to have rusqlite traits
implemented for from Jiff.

Jiff's default parsing/formatting is already generally compatible with
what SQLite itself supports, so we start there.


gwenn:
Thanks !
Would you mind fixing the Clippy warning ?

Issue #1173, opened 2022-04-29
get rid of hashlink


Dushistov:
I am trying to optimize my dependency tree,
and see that rusqlite depend on hashlink,
which depend on hashbrown, ahash, getrandom.
Is it slightly overkill for "lru cache" with size of 16?

What do you think about PR that remove "hashlink" and implement 
"lru cache" on top of std library with usage of knowledge that "lru cache"
is small enough (16)?

thomcc:
I actually have a WIP PR for this that I never managed to get up. That said, it's more complicated than what you describe, because while your LRU cache usage may only be 16, it's completely reasonable to increase it by a lot. In Firefox we use it with 256 IIRC, which is big enough to worry about.

I'll try to finish this PR soon though, it's still possible to be much simpler.

8573:
Another option might be [`cart-cache`](https://github.com/jedisct1/rust-cart-cache), a small library that depends only on `slab` and implements a more sophisticated algorithm.

w-flo:
Yet another option might be hiding the statement cache behind a feature flag, so hashlink would be optional.

PR #1682, opened 2025-04-20
Make statement cache optional

gwenn:
Related to #1173


GuillerLT:
Am I doing something wrong or are the functions behind this feature not showing in the documentation?
https://docs.rs/rusqlite/0.38.0/rusqlite/?search=cache
Is this the desired behavior?

PR #1684, opened 2025-04-22
Fix `loadable_extension` build by generating all functions

SabrinaJewson:
Currently, building the library with `--all-features` fails because because bindings for functions like `sqlite3_preupdate_hook` are not generated when the `loadable_extension` feature is active. This commit makes sure these bindings are generated always, and simply filters out those functions which shouldn't be generated with this feature.

There are many minor formatting changes in
`bindgen_bundled_version_ext.rs`, because now the code is formatted by prettyplease and not syn.


SabrinaJewson:
I see. Do you know which ones? I would be happy to replace this PR with appropriate invocations of `#[cfg(not(feature = "loadable_extension"))]` – I just want the build to succeed.

gwenn:
> I see – I had assumed it would work because calling `sqlite3_initialize` compiles, but it fails with a link-time error if you do `./load`.

But we want compile-time error (like in current version):
```diff
diff --git a/examples/loadable_extension.rs b/examples/loadable_extension.rs
index 7d62ecf..b71ea17 100644
--- a/examples/loadable_extension.rs
+++ b/examples/loadable_extension.rs
@@ -35,6 +35,7 @@ pub unsafe extern "C" fn sqlite3_extension_init(
 }

 fn extension_init(db: Connection) -> Result<bool> {
+    unsafe { ffi::sqlite3_initialize() };
     db.create_scalar_function(
         c"rusqlite_test_function",
         0,
```

```sh
% cargo build --example loadable_extension --features "loadable_extension functions trace"
   Compiling rusqlite v0.35.0 (/Users/gwen/Rust/rusqlite-clean)
error[E0425]: cannot find function `sqlite3_initialize` in crate `ffi`
    --> examples/loadable_extension.rs:38:19
     |
38   |     unsafe { ffi::sqlite3_initialize() };
     |                   ^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `sqlite3_finalize`
...
error: could not compile `rusqlite` (example "loadable_extension") due to 1 previous error
```

gwenn:
For me `loadable_extension` is incompatible with `--all-features` (at least some of them will be ignored: `session`, `sqlcipher*`, `preupdate_hook` and maybe others).
I don't know if it is possible to detect the `--all-features` flag in `build.rs`.
Maybe we should ignore `loadable_extension` when `--all-features` is used.
This would be the least surprise:
https://en.wikipedia.org/wiki/Principle_of_least_astonishment

SabrinaJewson:
> Maybe we should ignore `loadable_extension` when `--all-features` is used.

I would be happy to implement that as well – though I will point out there is precedent for disabling items if `loadable_extension` is enabled, for example `config_log` does this.

(currently I just added a bunch of `#[cfg(not(feature = "loadable_extension"))]` attributes and it works with `--all-features`)

gwenn:
We may keep your changes but still fix the behaviour of `-all-features`.

gwenn:
See #1686 which makes most of your changes useless...

PR #1685, opened 2025-04-22
Replace use of function pointers with `Fn*` traits

### LINKED ISSUE #977: Allow Rust closures as the callback in Connection::trace?
Currently Connection::trace accepts `fn(&str)` as the callback, which precludes use of Rust closures without using globals to hackily simulate captured variables.  Would it be possible to "upgrade" this to accept a callback of type `std::ops::Fn(&str)`?  Because `fn(&str)` satisfies `Fn(&str)` this theoretically wouldn't break backward compatibility.

I've made a few minor tries at implementing this feature myself, but my Rust-fu is not quite there yet, as I'm not sure how to get a C function pointer to a closure (seems impossible without some hackery using some state that lives longer than the closure).

SabrinaJewson:
This commit replaces all instances of methods accepting `fn(_) -> _` with `F: Fn(_) -> _`, for greater flexibility.

A new internal abstraction is introduced to facilitate this change: `ThinBoxAny`, a type similar to `Option<Box<dyn Any>>` but stored in a single pointer in order to avoid inflating the size of the `Connection` type too much.

There are two exceptions, the deprecated `trace` and `profile` functions. Users who need this flexibility should migrate to `trace_v2` instead.

This change is, unfortunately, breaking. There are two ways in which it breaks code:
+ Code that uses something like `.wal_hook(None)` will now get errors because Rust cannot infer the type parameter of `F`. Users must now write `.wal_hook(None::<fn(&Wal, _) -> _>)`.
+ Code that uses something like `.wal_hook(|hook, _| {})` will now fail, because Rust’s type inference of closures in the presence of HRTBs is limited. Instead, users should write `.wal_hook(|hook: &Wal, _| {})`.

The change is conservative for now, generally requiring `Fn` and `Sync` even when this may not be strictly necessary. It is likely many of these bounds can be relaxed in future.

Fixes: #977

### DIFF
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -61,11 +61,11 @@ impl Connection {
     ///
     /// Will return `Err` if the destination path cannot be opened
     /// or if the backup fails.
-    pub fn backup<N: Name, P: AsRef<Path>>(
+    pub fn backup<N: Name, P: AsRef<Path>, F: FnMut(Progress)>(
         &self,
         name: N,
         dst_path: P,
-        progress: Option<fn(Progress)>,
+        mut progress: Option<F>,
     ) -> Result<()> {
         use self::StepResult::{Busy, Done, Locked, More};
         let mut dst = Self::open(dst_path)?;
@@ -74,7 +74,7 @@ impl Connection {
         let mut r = More;
         while r == More {
             r = backup.step(100)?;
-            if let Some(f) = progress {
+            if let Some(f) = progress.as_mut() {
                 f(backup.progress());
             }
         }
@@ -288,11 +288,11 @@ impl Backup<'_, '_> {
     ///
     /// Will return `Err` if any of the calls to [`step`](Backup::step) return
     /// `Err`.
-    pub fn run_to_completion(
+    pub fn run_to_completion<F: FnMut(Progress)>(
         &self,
         pages_per_step: c_int,
         pause_between_pages: Duration,
-        progress: Option<fn(Progress)>,
+        mut progress: Option<F>,
     ) -> Result<()> {
         use self::StepResult::{Busy, Done, Locked, More};
 
@@ -300,7 +300,7 @@ impl Backup<'_, '_> {
 
         loop {
             let r = self.step(pages_per_step)?;
-            if let Some(progress) = progress {
+            if let Some(progress) = progress.as_mut() {
                 progress(self.progress());
             }
             match r {
@@ -365,7 +365,7 @@ mod test {
 
         {
             let backup = Backup::new(&src, &mut dst)?;
-            backup.run_to_completion(5, Duration::from_millis(250), None)?;
+            backup.run_to_completion(5, Duration::from_millis(250), None::<fn(_)>)?;
         }
 
         let the_answer: i64 = dst.one_column("SELECT SUM(x) FROM foo")?;
@@ -396,7 +396,7 @@ mod test {
 
         {
             let backup = Backup::new_with_names(&src, TEMP_DB, &mut dst, MAIN_DB)?;
-            backup.run_to_completion(5, Duration::from_millis(250), None)?;
+            backup.run_to_completion(5, Duration::from_millis(250), None::<fn(_)>)?;
         }
 
         let the_answer: i64 = dst.one_column("SELECT SUM(x) FROM foo")?;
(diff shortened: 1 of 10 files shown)


gwenn:
http://sqlite.org/c3ref/busy_timeout.html
> There can only be a single busy handler for a particular [database connection](http://sqlite.org/c3ref/sqlite3.html) at any given moment. If another busy handler was defined (using [sqlite3_busy_handler()](http://sqlite.org/c3ref/busy_handler.html)) prior to calling this routine, that other busy handler is cleared.

And also:
https://sqlite.org/pragma.html#pragma_busy_timeout
> Each database connection can only have a single [busy handler](https://sqlite.org/c3ref/busy_handler.html). This PRAGMA sets the busy handler for the process, possibly overwriting any previously set busy handler.

SabrinaJewson:
I have added a commit to eagerly free old busy handlers if `busy_timeout` is called. I have not addressed the `PRAGMA` case, but I assume that it is pretty rare for a user to both use PRAGMA and `busy_timeout`, and even in such cases, it seems unlikely that simply freeing a function later than strictly necessary would be a problem. I believe this would be hard to detect anyway since rusqlite does not parse its SQL…?

gwenn:
Idem:
http://sqlite.org/c3ref/wal_hook.html
> A single database handle may have at most a single write-ahead log callback registered at one time. ... . Note that the [sqlite3_wal_autocheckpoint()](http://sqlite.org/c3ref/wal_autocheckpoint.html) interface and the [wal_autocheckpoint pragma](http://sqlite.org/pragma.html#pragma_wal_autocheckpoint) both invoke [sqlite3_wal_hook()](http://sqlite.org/c3ref/wal_hook.html) and will overwrite any prior [sqlite3_wal_hook()](http://sqlite.org/c3ref/wal_hook.html) settings.

Maybe we should not use `Fn*` when we cannot know when to free them...

gwenn:
Idem:
http://sqlite.org/c3ref/trace_v2.html
> Each call to either sqlite3_trace(D,X,P) or sqlite3_trace_v2(D,M,X,P) overrides (cancels) all prior calls to sqlite3_trace(D,X,P) or sqlite3_trace_v2(D,M,X,P) for the [database connection](http://sqlite.org/c3ref/sqlite3.html) D. Each database connection may have at most one trace callback.

https://sqlite.org/c3ref/profile.html
> Invoking either [sqlite3_trace()](https://sqlite.org/c3ref/profile.html) or [sqlite3_trace_v2()](https://sqlite.org/c3ref/trace_v2.html) will cancel the profile callback.

gwenn:
> 2. All the functions will be freed when the database conection is closed, so there is no global memory leak.

Except if a connection pool is used which is more likely.
> 3. All existing usages of these functions are with function pointers, which generally come from non-capturing closures,

I don't understand: if there is non-capturing closures, current`fn` should be fine, no ?

oriongonza:
As a user I'd much rather take the risk of having the allocated Fn for longer if it means not having to do the ffi myself in order to have data in the callbacks

Darckfast:
from what i understood from this RFC https://rust-lang.github.io/rfcs/2113-dyn-trait-syntax.html `dyn Fn(_)` would be the recommended syntax for trait objects, instead of `Fn(_)`

gwenn:
> from what i understood from this RFC https://rust-lang.github.io/rfcs/2113-dyn-trait-syntax.html `dyn Fn(_)` would be the recommended syntax for trait objects, instead of `Fn(_)`

I am pretty sure that cannot free `dyn Fn` like in https://github.com/rusqlite/rusqlite/blob/bc267609a4547e0c7bc156f424662ebaf05a06a7/src/functions.rs#L635

gwenn:
Partially fixed by #1791:
- [X] https://github.com/rusqlite/rusqlite/pull/1685#issuecomment-3734274318 => [sqlite3_set_clientdata](https://sqlite.org/c3ref/get_clientdata.html)
- [X] `sqlite3_busy_handler`
- [X] `sqlite3_trace_v2` (deprecated bindings to `sqlite3_trace`  / `sqlite3_profile ` removed)
- [X] `sqlite3_wal_hook`
- [ ] `SQLITE_CONFIG_LOG` (static)
- [ ] `sqlite3_collation_needed`
- [ ] `sqlite3_auto_extension`

PR #1686, opened 2025-04-24
Feature `loadable_extension` is incompatible with some other features

gwenn:
At least `preupdate_hook` and `session`.


SabrinaJewson:
This makes it worse, doesn’t it? Now buildling with `--all-features` doesn’t work at all, which I think would very be annoying for new contributors.

PR #1696, opened 2025-05-20
Add support for transaction to the vtab module

jwhear:
The virtual table interface allows xBegin, xSync, xCommit, and xRollback to be defined. This commit provides a trait, `TransactionVTab` and rustified versions of those methods.

Note that the `module` macro is getting a bit cumbersome and should possibly be redesigned. Additionally the `readonly_module`, `update_module`, and `update_module_with_tx` simply aren't going to scale if other orthogonal virtual table features are added. It's not obvious to me what a better design would be but I'm happy to take suggestions.

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -103,16 +103,17 @@ const ZERO_MODULE: ffi::sqlite3_module = unsafe {
 };
 
 macro_rules! module {
-    ($lt:lifetime, $vt:ty, $ct:ty, $xc:expr, $xd:expr, $xu:expr) => {
+    ($lt:lifetime, $vt:ty, $ct:ty, $xcreate:expr, $xdestroy:expr, $xupdate:expr,
+         $xbegin:expr, $xsync:expr, $xcommit:expr, $xrollback:expr) => {
     &Module {
         base: ffi::sqlite3_module {
             // We don't use V3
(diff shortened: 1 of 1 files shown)


gwenn:
To be compared with:
- https://docs.rs/sqlite3_ext/latest/sqlite3_ext/vtab/trait.TransactionVTab.html
- https://docs.rs/sqlite3_ext/latest/sqlite3_ext/vtab/trait.VTabTransaction.html
- https://github.com/CGamesPlay/sqlite3_ext/blob/7e5b44dbe959fe05e5dd11f5d0901bb68bfe3b74/src/vtab/stubs.rs#L288-L328
- https://github.com/CGamesPlay/sqlite3_ext/blob/7e5b44dbe959fe05e5dd11f5d0901bb68bfe3b74/src/vtab/stubs.rs#L342-L370

gwenn:
> Note that the module macro is getting a bit cumbersome and should possibly be redesigned. Additionally the readonly_module, update_module, and update_module_with_tx methods simply aren't going to scale if other orthogonal virtual table features are added. It's not obvious to me what a better design would be but I'm happy to take suggestions.

Just for reference:
https://docs.rs/sqlite3_ext/latest/sqlite3_ext/attr.sqlite3_ext_vtab.html

PR #1703, opened 2025-05-29
Add `FromSqlError::other` convenience conversion

Kixunil:
When implementing `FromSql` it is common (as demonstrated by this commit diff) to call into a lower-level parsing function (such as `str::from_utf8`) and then convert the returned error to `FromSqlError::Other`. However this required manual boxing and unsizing which made the code a bit annoying. Thus this commit introduces an `other` generic method intentionally named after the `Other` variant. This method performs the boxing and unsizing so that callers can use it instead of writing a closure with `Box::new` (if `map_err` is used).

This commit also replaces all such occurrences in the `rusqlite`'s code which not only simplifies the code but also demonstrates the usefulness of the method. Incidentally, the change also simplifies manual `match`es to use `map_err` instead as well as replaces `FromStr` with the more idiomatic `parse` method in the those places.

### DIFF
--- a/src/types/chrono.rs
+++ b/src/types/chrono.rs
@@ -20,10 +20,7 @@ impl FromSql for NaiveDate {
     fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
         value
             .as_str()
-            .and_then(|s| match Self::parse_from_str(s, "%F") {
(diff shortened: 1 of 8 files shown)


gwenn:
rusqlite version 0.37.0 released

PR #1712, opened 2025-06-17
Fix for vtab::parameter parsing

### LINKED ISSUE #1711: vtab::parameter fails with extra "=" sign
Writing a VTab that has a more complicated schema generated.

```sql
CREATE VIRTUAL TABLE IF NOT EXISTS temp.ideas_status
USING markdowndb(
    schema='CREATE TABLE "ideas_status" ("inode" integer unsigned NOT NULL PRIMARY KEY CHECK ("inode" >= 0), "path" varchar(100) NOT NULL UNIQUE, "slug" varchar(50) NOT NULL UNIQUE, "content" text NOT NULL, "excerpt" text NOT NULL, "metadata" text NOT NULL CHECK ((JSON_VALID("metadata") OR "metadata" IS NULL)), "title" varchar(128) NOT NULL, "ordering" smallint NOT NULL)',
    path='.../content/ideas.Status'
);
```

When the schema is generated, there are some extra `=` signs in there, for example `CHECK ("inode" >= 0), `

https://github.com/rusqlite/rusqlite/blob/5c33953b75b59bf93e6ccf391ef884d11c5f4798/src/vtab/mod.rs#L1045-L1057

Currently does a simple `arg.split('=')` but perhaps migrating it to something like `arg.split_once('=')` might be more correct?

I am testing this locally though I can also try to prepare a patch if this seems reasonable (or if you have other suggestions)

```rust
/// `<param_name>=['"]?<param_value>['"]?` => `(<param_name>, <param_value>)`
pub fn parameter(c_slice: &[u8]) -> Result<(&str, &str)> {
    std::str::from_utf8(c_slice)?
        .trim()
        .split_once('=')
        .map(|parsed| {
            let param = parsed.0.trim();
            let value = vtab::dequote(parsed.1.trim());
            (param, value)
        })
        .ok_or_else(|| Error::ModuleError(format!("illegal argument: '{arg}'"))
}
```

kfdm:
vtab::parameter only needs to look for the first = to know where to split between the key and value as there may be values that also contain an '=' symbol.

This should fix #1711

I have only written some rust as a hobby, so please let me know if you would prefer it written a different way. I tried to copy the existing formats look.

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -1045,15 +1045,14 @@ pub fn parse_boolean(s: &str) -> Option<bool> {
 /// `<param_name>=['"]?<param_value>['"]?` => `(<param_name>, <param_value>)`
 pub fn parameter(c_slice: &[u8]) -> Result<(&str, &str)> {
     let arg = std::str::from_utf8(c_slice)?.trim();
-    let mut split = arg.split('=');
-    if let Some(key) = split.next() {
-        if let Some(value) = split.next() {
+    match arg.split_once('=') {
+        Some((key, value)) => {
             let param = key.trim();
             let value = dequote(value.trim());
-            return Ok((param, value));
+            Ok((param, value))
         }
+        _ => Err(Error::ModuleError(format!("illegal argument: '{arg}'"))),
     }
-    Err(Error::ModuleError(format!("illegal argument: '{arg}'")))
 }
 
 unsafe extern "C" fn rust_create<'vtab, T>(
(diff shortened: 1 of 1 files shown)


gwenn (approved):

kfdm:
I guess I neglected to run `cargo fmt`
Would you like me to do that and resubmit ?

gwenn:
> I guess I neglected to run `cargo fmt`
> Would you like me to do that and resubmit ?

Please do.

kfdm:
Ok ran `cargo fmt`. I also changed one of the test cases so it would be a bit shorter and fit on one line for aesthetic reasons. Hopefully should be in good shape now.

gwenn:
rusqlite version 0.37.0 released

PR #1733, opened 2025-08-30
Bump minimal SQLite version to 3.34.1

### LINKED ISSUE #1731: libsqlite3-sys: support 3.20 (or `sqlite3_prepare_v3()`) in pre-generated bindings
In SQLx, we're forced to use either the `bundled` or the `bindgen` features because we unconditionally depend on `sqlite3_prepare_v3()` which was not added until SQLite 3.20.0: https://www.sqlite.org/changes.html#version_3_20_0
 
We use the `SQLITE_PREPARE_PERSISTENT` flag as designed to avoid the depletion of lookaside memory because we cache prepared statements for potentially very long periods of time: https://www.sqlite.org/c3ref/c_prepare_dont_log.html#sqlitepreparepersistent

However, the `bundled` and `bindgen` features both have significant tradeoffs, so some users would probably like to have the option of using pre-generated bindings.

Would it be possible to add a `min_sqlite_version_3_20_0` feature that includes all APIs supported by 3.20? I'd be happy to open a PR if there's interest.

gwenn:
Fix #1731

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,10 +16,10 @@ install:
   - rustc -V
   - cargo -V
   # download SQLite dll (useful only when the `bundled` feature is not set)
-  - appveyor-retry appveyor DownloadFile https://sqlite.org/2018/sqlite-dll-win64-x64-3250200.zip -FileName sqlite-dll-win64-x64.zip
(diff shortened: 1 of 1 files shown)


abonander:
When I opened #1731, I assumed the minimum supported version of SQLite was so old because there's people out there who depend on that. I've encountered at least one such person: https://github.com/launchbadge/sqlx/issues/3905#issuecomment-3047929361

I'm not against this, but you might end up with some people crawling out of the woodwork to yell at you about it.

PR #1737, opened 2025-09-21
Use CARGO_CFG_TARGET_FEATURE for crt-static check

PaulWagener:
This change checks crt-static in the CARGO_CFG_TARGET_FEATURE environment variable instead of directly using the cfg!() macro. These can differ when cross-compiling from another platform.


PaulWagener:
AppVeyor CI is currently failing on this, but this is not caused by the change.

Here is a PR with essentially nothing changed which also fails: https://github.com/rusqlite/rusqlite/pull/1738

PR #1744, opened 2025-10-09
Remove shell scripts from the published package

weiznich:
This commit removes shell scripts from the published package. These were reported by cargo denys `bans.build.interpreted` setting and they seem to be not required for building the crate.


weiznich:
No worries, you are the maintainer here so if you prefer it a certain way that's fine. 
I'm happy as long as the scripts are removed

PR #1770, opened 2025-12-11
vtab: add traits and module builder for all vtab functions

enkore:
vtab doesn't seem to have many tests besides the smoke tests with the included modules, so I didn't add any.

The new const module builder is imho a much cleaner design than the combinatorial explosion that the fixed-function module builders had.

Note that building a loadable extension with a vtab implementing IntegrityVTab requires basically the following features:

```
features = ["vtab", "loadable_extension", "modern_sqlite", "buildtime_bindgen"]
```

Clippy notions that Module should have a `Default` impl, which I don't think makes any sense here, so I added an allowance instead of `Default`.

### DIFF
--- a/src/vtab/mod.rs
+++ b/src/vtab/mod.rs
@@ -816,6 +816,7 @@ pub trait FindFunctionVTab<'vtab>: VTab<'vtab> {
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 #[allow(missing_docs)]
 #[expect(non_camel_case_types)]
+#[non_exhaustive]
 pub enum IndexConstraintOp {
     /// Unknown/unsupported index constraint which should be ignored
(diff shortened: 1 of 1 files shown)


enkore:
Since it's now quite easy to add these traits, here are SavepointVTab (xSavepoint/xRelease/xRollbackTo) and FindFunctionVTab (xFindFunction). The former is very straightforward, the latter is a bit more involved. I haven't written any actual code using xFF, so I'm not entirely sure what the usage pattern is, _but_ I think the way that the lifetimes point is clear enough: you're probably not going to dynamically overload functions, they're part of the vtab, defined during connect/create and it is there where you close over the required integration points with the actual backend code of the vtab.

With this, all methods provided by sqlite are now wrapped.

PR #1811, opened 2026-02-27
Sqlcipher: support custom crypto provider

rubdos:
This PR introduces an interface to enable a custom crypto provider for Sqlcipher, and ships a simple custom crypto provider through the RustCrypto crates.

### Motivation
Sqlcipher uses OpenSSL as default crypto provider. Either it ships its own vendored OpenSSL (`bundled-sqlcipher-vendored-openssl`) or links to the system OpenSSL. Either way, it depends on OpenSSL, and may conflict with projects that use BoringSSL (exposing the same C symbols).

Second, this allows to further reduce reliance on C-based crypto implementations.

Third, custom crypto providers *technically* allow customized crypto on sqlcipher databases, although the sqlcipher interface really isn't designed to swap out primitives.

### Changes
- Added a new trait `SqlcipherCryptoProvider`, with some rough mapping to the `struct sqlcipher_provider` C-style vtable.
- Add `RustCryptoProvider`, an implementation of that trait, based on RustCrypto's AES/PBKDF2/HMAC/digest. This is compatible with the existing sqlcipher implementation.
- Hardcode the `RustCryptoProvider` when the `bundled-sqlcipher-custom-crypto` is enabled.
- Swap `sqlite3.h` out for `sqlcipher.h` in binding generation, to access the necessary interfaces.
- Backport https://github.com/sqlcipher/sqlcipher/commit/c89bab0b4 with a `.patch` in `upgrade_sqlcipher.sh`.

### Backward Compatibility
This change is fully backward-compatible. Existing code will continue to use the default provider unless explicitly overridden.

### Open Questions
- How do we test this, and keep this CI-tested? I have locally created an SQLcipher database, and tried to read it with our own interface. This works.
- We probably want to allow external implementations of the interface. SQLcipher has the infra to dynamically swap out the provider. However: it *kinda* wants a default fallback. Setting a default provider will probably also require some static RefCell. I didn't necessarily think a lot on this yet.
- Does the `rusqlite` repo actually want something like this? It's a maintenance overhead, and I don't necessarily want to burden you with it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ maintenance = { status = "actively-developed" }
 name = "rusqlite"
 
 [workspace]
-members = ["libsqlite3-sys"]
+members = ["libsqlite3-sys", "sqlcipher-crypto-provider"]
 
 [features]
 # if not SQLITE_OMIT_LOAD_EXTENSION
@@ -53,6 +53,7 @@ trace = []
 bundled = ["libsqlite3-sys/bundled", "modern_sqlite"]
 # Use SQLCipher instead of SQLite
 bundled-sqlcipher = ["libsqlite3-sys/bundled-sqlcipher", "bundled"]
+bundled-sqlcipher-custom-crypto = ["libsqlite3-sys/bundled-sqlcipher-custom-crypto", "bundled-sqlcipher"]
 bundled-sqlcipher-vendored-openssl = [
     "libsqlite3-sys/bundled-sqlcipher-vendored-openssl",
     "bundled-sqlcipher",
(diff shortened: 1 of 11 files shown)


marcprux:
Could this improvement also enable the libtomcrypt crypto provider for sqlcipher (which had been deprecated but then [restored](https://www.zetetic.net/blog/2026/03/17/sqlcipher-4.14.0-release/#libtomcrypt-provider-restored) in https://github.com/sqlcipher/sqlcipher/issues/564#issuecomment-4085025266)? This would be valuable for platforms that want a pure source build without any required dependencies (especially important for mobile platforms like iOS and Android), and for which a pure source build of OpenSSL might be infeasible or overkill.

PR #1831, opened 2026-04-16
Update link to pre-update hook

sondr3:
Noticed this now leads to 404 when clicking around


gwenn:
Some links are unstable (here, when `sqlite3_preupdate_blobwrite` was added).

PR #1836, opened 2026-04-20
feat: add BorrowingStatement for zero-copy SQLITE_STATIC binds

### LINKED ISSUE #164: Allow binding of text/blobs without SQLITE_TRANSIENT?
Currently, we always pass the `SQLITE_TRANSIENT` flag when binding a text or blob parameter, instructing SQLite to make its own copy of the data. It would be nice to be able to avoid that copy if we can come up with a safe interface that allows it.

Enduriel:
Would close #164 by adding 2 new statement variants that expose `raw_bind_parameter_ref` to bind parameters with a lifetime longer than the lifetime of the statements allowing for the use of SQLITE_STATIC lifetime when binding.

This also includes a bit of a refactor to prevent code duplication with `raw_bind_parameter`, and From<&'a X> impls for ValueRef<'a>.

I wanted to make this PR as I saw significant improvements (~25% faster) with large SQLITE insertions (~3 million rows) that were already optimized by using outer transactions and bulk insert statements, obviously this performance improvement will vary depending on situation, but this is a relatively simple change that will improve performance in large insertions.

In my opinion, this would be a good intermediary step to slowly integrate these lifetimes into the core `Statement` and `CachedStatement` structs, so that in the long term this performance could be offered for 'free', but at least for now it would be great to have it be available to people who want to and are able to use it at all.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -225,6 +225,10 @@ harness = false
 name = "exec"
 harness = false
 
+[[bench]]
+name = "borrowing_stmt"
+harness = false
+
 [[example]]
 name = "loadable_extension"
 crate-type = ["cdylib"]
--- a/benches/borrowing_stmt.rs
+++ b/benches/borrowing_stmt.rs
@@ -1,5 +1,5 @@
 use bencher::{benchmark_group, benchmark_main, Bencher};
-use rusqlite::Connection;
+use rusqlite::{params, Connection};
 use uuid::Uuid;
 
 struct BenchData {
@@ -130,6 +130,40 @@ fn bench_borrowing(b: &mut Bencher) {
     });
 }
 
+fn bench_borrowing_one_transaction(b: &mut Bencher) {
+    let mut db = init_db();
+
+    let data = create_data();
+
+    b.iter(|| {
+        let transaction = db.transaction().unwrap();
+        let mut stmt = transaction
+            .prepare_borrowing(&make_statement_sql())
+            .unwrap();
+        for insert_chunk in data.chunks(BATCH_VALUES) {
+            for (i, row) in insert_chunk.iter().enumerate() {
+                let i = i * COLS_PER_ROW + 1;
+                stmt.raw_bind_parameter_ref(i, row.name.as_str()).unwrap();
+                stmt.raw_bind_parameter_ref(i + 1, row.uuid.as_bytes())
+                    .unwrap();
+                stmt.raw_bind_parameter_ref(i + 2, row.parent.as_bytes())
+                    .unwrap();
+                stmt.raw_bind_parameter_ref(i + 3, row.some_u64 as i64)
+                    .unwrap();
+                stmt.raw_bind_parameter_ref(i + 4, row.string_a.as_str())
+                    .unwrap();
+                stmt.raw_bind_parameter_ref(i + 5, row.string_b.as_str())
+                    .unwrap();
+                stmt.raw_bind_parameter_ref(i + 6, row.string_c.as_str())
+                    .unwrap();
+            }
+            stmt.raw_execute().unwrap();
+        }
+        drop(stmt);
+        transaction.commit().unwrap();
+    });
+}
+
 fn bench_copy(b: &mut Bencher) {
     let db = init_db();
 
@@ -161,5 +195,82 @@ fn bench_copy(b: &mut Bencher) {
     });
 }
 
-benchmark_group!(borrowing_stmt_benches, bench_borrowing, bench_copy);
+fn bench_single_rows(b: &mut Bencher) {
+    let db = init_db();
+
+    let data = create_data();
+
+    b.iter(|| {
+        let mut stmt = db.prepare("INSERT INTO bench_data (name, uuid, parent, some_u64, string_a, string_b, string_c) VALUES (?, ?, ?, ?, ?, ?, ?)").unwrap();
+        for transaction_chunk in data.chunks(TRANSACTION_SIZE) {
+            let transaction = db.unchecked_transaction().unwrap();
+            for row in transaction_chunk {
+                      stmt.execute(params![&row.name, &row.uuid.as_bytes(), &row.parent.as_bytes(), row.some_u64 as i64, &row.string_a, &row.string_b, &row.string_c]).unwrap();       }
+            transaction.commit().unwrap();
+        }
+    });
+}
+
+fn bench_single_rows_one_transaction(b: &mut Bencher) {
+    let mut db = init_db();
+
+    let data = create_data();
+
+    b.iter(|| {
+		let transaction = db.transaction().unwrap();
+		{
+			let mut stmt = transaction
+				.prepare("INSERT INTO bench_data (name, uuid, parent, some_u64, string_a, string_b, string_c) VALUES (?, ?, ?, ?, ?, ?, ?)")
+				.unwrap();
+			for row in &data {
+				stmt.execute(params![
+					&row.name,
+					&row.uuid.as_bytes(),
+					&row.parent.as_bytes(),
+					row.some_u64 as i64,
+					&row.string_a,
+					&row.string_b,
+					&row.string_c
+				])
+				.unwrap();
+			}
+		}
+
+		transaction.commit().unwrap();
+	});
+}
+
+fn bench_single_rows_no_transaction(b: &mut Bencher) {
+    let db = init_db();
+
+    let data = create_data();
+
+    b.iter(|| {
+			let mut stmt = db
+				.prepare("INSERT INTO bench_data (name, uuid, parent, some_u64, string_a, string_b, string_c) VALUES (?, ?, ?, ?, ?, ?, ?)")
+				.unwrap();
+			for row in &data {
+				stmt.execute(params![
+					&row.name,
+					&row.uuid.as_bytes(),
+					&row.parent.as_bytes(),
+					row.some_u64 as i64,
+					&row.string_a,
+					&row.string_b,
+					&row.string_c
+				])
+				.unwrap();
+			}
+	});
+}
+
+benchmark_group!(
+    borrowing_stmt_benches,
+    bench_borrowing,
+    bench_borrowing_one_transaction,
+    bench_copy,
+    bench_single_rows,
+    bench_single_rows_one_transaction,
+    bench_single_rows_no_transaction
+);
 benchmark_main!(borrowing_stmt_benches);
--- a/benches/borrowing_stmt.rs
+++ b/benches/borrowing_stmt.rs
@@ -1,0 +1,165 @@
+use bencher::{benchmark_group, benchmark_main, Bencher};
+use rusqlite::Connection;
+use uuid::Uuid;
+
+struct BenchData {
+    name: String,
+    uuid: Uuid,
+    parent: Uuid,
+
+    some_u64: u64,
+    string_a: String,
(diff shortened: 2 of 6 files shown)


gwenn:
Are you sure that SQLite makes no copy of Rust strings which are not zero terminated ?
(because the [SQLITE_UTF8_ZT](https://sqlite.org/c3ref/c_any.html#sqliteutf8zt) suggests the contrary)

Enduriel:
Honestly, no, but I'm still seeing a significant improvement when benchmarking. Here are two excerpts from my flamegraph where the only change I have made between them is change the passed values in`bind_value_ref_static` between `ffi:SQLITE_TRANSIENT` and `ffi::SQLITE_STATIC`

image

image

You can see that the amount of time (and call stack) spent inside of `bind_value_ref_inner` is very different.

There might be some additional improvements to make here, and I'm definitely open to making this even better, but there are definitely real improvements here already. I also tested by only changing the text_dtor argument (my insert uses both blobs and text), and there was still a 12% decrease in time taken for my overall test, so this 100% does make an impact even for TEXT columns.

I am by no means an SQLite expert though (my C is not great) so the details of what exactly is still happening are hard for me to figure out.

I can just say that the lifetime management here respects the SQLite docs [here](https://sqlite.org/c3ref/bind_blob.html), specifically:
> (2) The special constant, [SQLITE_STATIC](https://sqlite.org/c3ref/c_static.html), may be passed to indicate that the application remains responsible for disposing of the object. In this case, the object and the provided pointer to it must remain valid until either the prepared statement is finalized or the same SQL parameter is bound to something else, whichever occurs sooner.

and my code runs faster.

gwenn:
With this quick and dirty patch (`SQLITE_STATIC` is forced when string length is 10):
````diff
diff --git a/Cargo.toml b/Cargo.toml
index 33b411a..1f75038 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -225,6 +225,10 @@ harness = false
 name = "exec"
 harness = false
 
+[[bench]]
+name = "param"
+harness = false
+
 [[example]]
 name = "loadable_extension"
 crate-type = ["cdylib"]
diff --git a/benches/param.rs b/benches/param.rs
new file mode 100644
index 0000000..7d8eb85
--- /dev/null
+++ b/benches/param.rs
@@ -0,0 +1,29 @@
+use bencher::{benchmark_group, benchmark_main, Bencher};
+use rusqlite::Connection;
+
+fn bench_dynamic(b: &mut Bencher) {
+    let db = Connection::open_in_memory().unwrap();
+    let mut stmt = db.prepare_cached("SELECT length(?)").unwrap();
+    b.iter(|| {
+        stmt.raw_bind_parameter(1, "123456789").unwrap();
+        let mut rows = stmt.raw_query();
+        while let Some(row) = rows.next().unwrap() {
+            assert_eq!(row.get::<_, i8>(0).unwrap(), 9i8);
+        }
+    });
+}
+
+fn bench_static(b: &mut Bencher) {
+    let db = Connection::open_in_memory().unwrap();
+    let mut stmt = db.prepare_cached("SELECT length(?)").unwrap();
+    b.iter(|| {
+        stmt.raw_bind_parameter(1, "1234567890").unwrap();
+        let mut rows = stmt.raw_query();
+        while let Some(row) = rows.next().unwrap() {
+            assert_eq!(row.get::<_, i8>(0).unwrap(), 10i8);
+        }
+    });
+}
+
+benchmark_group!(param_benches, bench_dynamic, bench_static);
+benchmark_main!(param_benches);
diff --git a/src/lib.rs b/src/lib.rs
index c69b89a..8802b31 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -327,7 +327,11 @@ fn str_for_sqlite(
 ) {
     let len = s.len();
     let (ptr, dtor_info) = if len != 0 {
-        (s.as_ptr().cast::<c_char>(), ffi::SQLITE_TRANSIENT())
+        if len != 10 {
+            (s.as_ptr().cast::<c_char>(), ffi::SQLITE_TRANSIENT())
+        } else {
+            (s.as_ptr().cast::<c_char>(), ffi::SQLITE_STATIC())
+        }
     } else {
         // Return a pointer guaranteed to live forever
         ("".as_ptr().cast::<c_char>(), ffi::SQLITE_STATIC())
````
````sh
$ cargo test --bench param
test bench_dynamic ... bench:         284 ns/iter (+/- 332)
test bench_static  ... bench:         269 ns/iter (+/- 293)
````
Too many noise, but not 25% faster...

**Comment by Enduriel** [comment]
That test seems incredibly noisy and like I outlined in the original message I saw this improvement when performing bulk inserts.

For some details I was inserting 3 million lines, 1 million each into 3 different tables (shared primary key), creating transactions of 10,000 inserts at a time per table and each statement was a batched insert of 200 rows at a time with 4-8 values per row depending on table.

In that scenario I see just under 25% improvement on my min spec M4 Mac Mini.

I could write a self contained benchmark to showcase this if that is necessary but even in the very basic and noisy test with very different characteristics we still see an improvement of 5% that is not achievable without this approach.

Would you like me to write a test that showcases the improvement here? I'm not sure what the concern is. I didn't think that would be necessary but I can do it if the improvement isn't obvious just from the code.

**Comment by gwenn** [comment]
> Are you sure that SQLite makes no copy of Rust strings which are not zero terminated ? (because the [SQLITE_UTF8_ZT](https://sqlite.org/c3ref/c_any.html#sqliteutf8zt) suggests the contrary)

`sqlite3_bind_text64` calls `bindText` which calls [`sqlite3VdbeMemSetText`](https://github.com/sqlite/sqlite/blob/d3abfd77f5cfe9f38b2810933fd8f720ad01256b/src/vdbemem.c#L1433).
And there is no copy made when `xDel` is `SQLITE_STATIC`.

**Comment by Enduriel** [comment]
That makes sense and explains the improvement.

**Comment by Enduriel** [comment]
@gwenn I added a more optimized benchmark to show the improvement. In this case I actually had a 50% reduction in time taken:

```
running 2 tests
test bench_borrowing ... bench: 123,446,433 ns/iter (+/- 1,584,886)
test bench_copy      ... bench: 254,493,600 ns/iter (+/- 2,455,877)
```

**Comment by Enduriel** [comment]
To show comparison to increasingly naive approaches, this is still all without re-preparing (even cached) statements.

```
running 6 tests
test bench_borrowing                   ... bench: 121,817,950 ns/iter (+/- 1,877,427)
test bench_borrowing_one_transaction   ... bench: 123,784,699 ns/iter (+/- 1,404,491)
test bench_copy                        ... bench: 252,688,424 ns/iter (+/- 2,665,737)
test bench_single_rows                 ... bench: 332,934,462 ns/iter (+/- 2,079,400)
test bench_single_rows_no_transaction  ... bench: 674,974,849 ns/iter (+/- 2,838,139)
test bench_single_rows_one_transaction ... bench: 335,583,933 ns/iter (+/- 3,209,982)
```

Enduriel:
Let me know if this is something you'd be interested in having merged in and/or if you'd like it integrated into the standard Statement API instead with a version bump or if you'd like to keep it this way (for now?)

gwenn:
Here are my concerns:
- code duplication: breaking changes might be OK
- `SQLITE_STATIC` with Rust string might be a problem when used with SQL functions / operators like `SELECT cast(? AS NUMBER)` that expect zero terminated C string.
- For Rust smart pointers (at least `Rc`, `Box`), we might be able to use their own destructors [free_rc](https://github.com/rusqlite/rusqlite/blob/63181d4cff680e2738b4618d9c422ed35eb23bd7/src/types/to_sql.rs#L45) / [free_boxed_value](https://github.com/rusqlite/rusqlite/blob/63181d4cff680e2738b4618d9c422ed35eb23bd7/src/util/mod.rs#L17).

Enduriel:
- Would you be okay with my reworking the normal statements structs to have the 'stmt lifetime and then adjusting all the appropriate functions to require either owned types or types with 'stmt lifetime? This would probably worsen ergonomics for people who in the past passed references with shorter lifetimes and require them to rework their calls, and unfortunately I don't think there's anything we can do to make that not be annoying without specialization over lifetimes.
- According to [the docs](https://sqlite.org/c3ref/bind_blob.html):
> The sixth argument (the E argument) to sqlite3_bind_text64(S,K,Z,N,D,E) must be one of SQLITE_UTF8, SQLITE_UTF8_ZT, SQLITE_UTF16, SQLITE_UTF16BE, or SQLITE_UTF16LE to specify the encoding of the text in the third parameter, Z. The special value SQLITE_UTF8_ZT means that the string argument is both UTF-8 encoded and is zero-terminated. In other words, SQLITE_UTF8_ZT means that the Z array is allocated to hold at least N+1 bytes and that the Z[N] byte is zero. If the E argument to sqlite3_bind_text64(S,K,Z,N,D,E) is not one of the allowed values shown above, or if the text encoding is different from the encoding specified by the sixth parameter, then the behavior is undefined.

Since SQLITE_UTF8_ZT specifically requires that the string be zero terminated, this implies that the other types don't need to be, and the section on lifetimes makes no reference to encoding type. I can test this just as a sanity check but I'm not sure a single test is a strong enough guarantee anyway, and I think trust the docs makes more sense.

With lifetimes specialization I can imagine a perfect world we would convert &str with a shorter lifetime into an owned Box<CStr> so that SQLite can make additional optimizations since we already have to pay for the allocation, but I don't think there's a way to take advantage of that encoding automatically in stable rust.

- this is actually a great suggestion, I agree that we should also do this. Would be willing to add that to this PR.

gwenn:
> For Rust smart pointers (at least Rc, Box), we might be able to use their own destructors [free_rc](https://github.com/rusqlite/rusqlite/blob/63181d4cff680e2738b4618d9c422ed35eb23bd7/src/types/to_sql.rs#L45) / [free_boxed_value](https://github.com/rusqlite/rusqlite/blob/63181d4cff680e2738b4618d9c422ed35eb23bd7/src/util/mod.rs#L17).

This doesn't seem feasible because with `Rc<str>`/`Box<str>`, `str` is not Sized so we cannot cast the raw pointer in the destructor (same for `[u8]`).

Enduriel:
> This doesn't seem feasible because with Rc<str>/Box<str>, str is not Sized so we cannot cast the raw pointer in the destructor (same for [u8]).

I'd have to dig into it to comment more on that. What are your thoughts regarding my other two points?

gwenn:
I know it would be suboptimal but there is already the `'conn` lifetime on `Statement`. So I guess we can avoid code duplication / breaking changes if the "static" values have a greater lifetime than `'conn`, no ?

Enduriel:
Unfortunately, while that is a much simpler approach, that's pretty much the exact inverse of the common case this PR is looking to fix: a long-lived (cached) prepared statement with transient borrowed rows bound and inserted in batches.

Since `'conn` is fixed with the connection, there's no way to reset it. The into_statement → BorrowingStatement::from round-trip works only because `'stmt` is different from `'conn`.

This means that in practice you'd have to hold all the values you want to insert in memory before even preparing a statement, so you couldn't for example generate values for each batch in rust, execute the statement, `clear_bindings` and send the values somewhere else or even drop them immediately.

It's an improvement vs the status quo, but I think going all the way here makes more sense, especially if we could also get SQLite to take ownership of and free owned values.

gwenn:
https://sqlite.org/forum/forumpost/282518aa1a
> The string need not be NUL terminated as long as "n" is greater than or equal to 0.

PR #1844, opened 2026-05-10
Add support to UtcDateTime

### LINKED ISSUE #1843: Please add support for converting `time::UtcDateTime` to/from SQL values under the `time` feature
Hi! I've been using `UtcDateTime`s to store dates in a database.

However, for all the `time` crate types that `rusqlite` supports under the `time` feature, `UtcDateTime` is not one of them - despite other time-like crates having support for their types that assume a timezone of UTC.

While it's not necessarily hard to write a newtype that allows me to store a `UtcDateTime` (in fact, I've already written one), it would be nice for `rusqlite` to support `UtcDateTime` without extra effort for me.

If this gets added to `rusqlite`, thank you in advance. :)

gwenn:
Fix #1843

### DIFF
--- a/src/types/time.rs
+++ b/src/types/time.rs
@@ -16,7 +16,7 @@ use crate::types::{FromSql, FromSqlError, FromSqlResult, ToSql, ToSqlOutput, Typ
 use crate::{Error, Result};
 use time::format_description::FormatItem;
 use time::macros::format_description;
-use time::{Date, OffsetDateTime, PrimitiveDateTime, Time};
+use time::{Date, OffsetDateTime, PrimitiveDateTime, Time, UtcDateTime};
 
 const OFFSET_DATE_TIME_ENCODING: &[FormatItem<'_>] = format_description!(
     version = 2,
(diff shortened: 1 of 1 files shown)


src/types/time.rs
```diff
@@ -158,14 +168,24 @@ impl FromSql for PrimitiveDateTime {
     }
 }
 
+impl FromSql for UtcDateTime {
+    #[inline]
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        value.as_str().and_then(|s| {
+            Self::parse(s, &PRIMITIVE_DATE_TIME_FORMAT)
```
LikeLakers2, line 175: Thank you!

A little note: I would've recommended using `UTC_DATE_TIME_FORMAT` here (in the `FromSql` impl), as the format supports a `Z` at the end - which for ISO8601 or RFC3339 timestamps means "this timestamp is explicitly in the UTC timezone without any offset".

gwenn: See #1859

PR #1850, opened 2026-05-24
Replace some cfg(not by cfg_select

gwenn:


djc:
This makes it impossible to use libsqlite3-sys on any version of Rust older than 1.95, which IMO is pretty hostile for what is a sort of low-level crate. Would you be open to a PR reverting this, committing to an explicitly maintained `rust-version`, and testing that that works in CI?

gwenn:
> This makes it impossible to use libsqlite3-sys on any version of Rust older than 1.95, which IMO is pretty hostile for what is a sort of low-level crate. Would you be open to a PR reverting this, committing to an explicitly maintained `rust-version`, and testing that that works in CI?

https://github.com/rusqlite/rusqlite#minimum-supported-rust-version-msrv
> Latest stable Rust version at the time of release.

xtqqczze:
@djc `cfg_select` can be polyfilled to support versions of Rust older than 1.95, if anyone was interested:

https://github.com/rust-lang/compiler-builtins/blob/9c7d3e8d064e331144dbe3426a0bf5311399a857/libm/src/math/support/macros.rs#L1-L23

@gwenn Would you be open to a PR with this workaround?

gwenn:
Recently, there is #1863 that also sets min version to 1.85.
And maybe others.
I might also bump the min SQLite version to 3.44 for #1791.

djc:
1.85 would be much better than 1.95 IMO.

xtqqczze:
Rust 1.88 would be the lowest achievable due to use of let-chains.

gwenn:
It seems that
```rust
#[rustversion::before(1.95)]
macro_rules! cfg_select {
... // copy paste from https://github.com/rust-lang/compiler-builtins/blob/9c7d3e8d064e331144dbe3426a0bf5311399a857/libm/src/math/support/macros.rs#L1-L23
}
```
doesn't work.
Which seems to imply that only
```rust
        #[rustversion::since(1.95)]
        cfg_select! {
...
        }
        #[rustversion::before(1.95)]
        // old code
```
would work.
Which is worse.

xtqqczze:
> ```rust
> #[rustversion::before(1.95)]
> macro_rules! cfg_select {
> ... // copy paste from https://github.com/rust-lang/compiler-builtins/blob/9c7d3e8d064e331144dbe3426a0bf5311399a857/libm/src/math/support/macros.rs#L1-L23
> }
> ```

Right, it looks like a limitation of this implementation is that branch bodies must be enclosed in braces. For example, the `foo` arm is supported, but the `bar` arm is not:

```rust
cfg_select! {
    foo => { m1() }
    bar => m2()
}
```

See also: https://github.com/rust-lang/hashbrown/issues/740#issuecomment-5137658766

gwenn:
I still have errors even after enclosing bodies in braces for example here:
https://github.com/rusqlite/rusqlite/blob/54f7d4d6079f969ca80dd37932fc3a9b25f5c70e/src/inner_connection.rs#L281-L284

xtqqczze:
@gwenn I look a look and was able to get this to compile: https://github.com/xtqqczze/rusqlite-rusqlite/commit/f4cde327fb5d9ab3eca55380fbb6cf8d5ec67be3

gwenn:
> @gwenn I look a look and was able to get this to compile: [xtqqczze@f4cde32](https://github.com/xtqqczze/rusqlite-rusqlite/commit/f4cde327fb5d9ab3eca55380fbb6cf8d5ec67be3)

Thanks.
See #1885
I've added braces in the macro body like you.
And I've reused the `error` module because of the current `#[macro_use]`.
And still no `rust-version` according to current MSRV policy.

xtqqczze:
It would be nice to include the `rust-version` metadata, since the Cargo resolver can then take it into account when picking dependencies. The MSRV policy could then simply state that the `rust-version` may be increased at any time.

gwenn:
> It would be nice to include the `rust-version` metadata, since the Cargo resolver can then take it into account when picking dependencies. The MSRV policy could then simply state that the `rust-version` may be increased at any time.

But this implies checking that we don't break the `rust-version`.
(currently, CI ensures that we don't break our MSRV policy by **simply** using the stable version)

xtqqczze:
Understood. That would amount to suggesting a change to the current policy, so I'll open a separate issue to discuss it.

gwenn:
And for your information, I am going to release version 0.40.2 (0.38.2 for libsqlite3-sys) with the patch.
Thanks @xtqqczze for your help.

xtqqczze:
I've crated a draft PR for discussion: https://github.com/rusqlite/rusqlite/pull/1886

PR #1854, opened 2026-06-04
Fix SQL injection when SAVEPOINT name is tainted

gwenn:


gwenn:
Not strictly related but we can check pragma with:
```sql
SELECT 1 FROM pragma_pragma_list WHERE name = 'schema_version';
```
https://sqlite.org/pragma.html#pragma_pragma_list

PR #1878, opened 2026-07-28
chore: remove dead branch

FranciscoTGouveia:
If I am not overlooking anything, I believe this branch is statically known to never be taken, thus being dead code.

For reference, this was found by an optimizer developed as a research project.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -558,9 +558,6 @@ impl Connection {
                 .prepare(self, sql, PrepFlags::default())?;
             if !stmt.stmt.is_null() && stmt.step()? {
                 // Some PRAGMA may return rows
-                if false {
-                    return Err(Error::ExecuteReturnedResults);
-                }
             }
             if tail == 0 || tail >= sql.len() {
                 break;


gwenn:
I would prefer to keep it untouched.
Because some `PRAGMA` used for updating the database return rows.

FranciscoTGouveia:
Apologies, but could you explain in what situations this branch would actually execute?
Even if `PRAGMA` returns rows, this branch would still be dead, no?

gwenn:
Currently, there is no way to know if an SQL command is changing the database (and eventually returning the changes).
So the branch is actually dead.
But maybe some day we will able to catch when someone does `execute_batch('SELECT 1;')`.

FranciscoTGouveia:
> But maybe some day we will able to catch when someone does `execute_batch('SELECT 1;')`.

But it wouldn't be caught with the current branch; it would need to be an entirely different condition, right?

PS: I'm not just trying to get the PR merged; I'm genuinely interested in understanding the reasoning behind this.

gwenn:
https://github.com/rusqlite/rusqlite/commit/0de7020603782e3e5279dad809c9df0fbd1fb9e4

FranciscoTGouveia:
Ok, I think I have got it now. Thanks for linking the commit.

I believe we can either close the PR or delete the dead branch and leave a comment explaining the situation.
What would be your preference?

gwenn:
https://github.com/rusqlite/rusqlite/pull/1878#issuecomment-5105856515

PR #1882, opened 2026-08-05
Expose sqlite3_file_control

### LINKED ISSUE #1881: feature request: interface for setting SQLITE_FCNTL_PERSIST_WAL file control option
currently, i have to do something like this to prevent the -wal and -shm files from being deleted when the last connection to the database closes:
```rust
unsafe {
    let mut value = 1_i32;
    let rc = sqlite3::ffi::sqlite3_file_control(
        db.handle(),
        c"main".as_ptr(),
        sqlite3::ffi::SQLITE_FCNTL_PERSIST_WAL,
        (&raw mut value).cast(),
    );
    if rc != 0 {
        panic!("{:?}", sqlite3::ffi::Error::new(rc));
    }
}
```

this is necessary to allow read-only connections to the database file in WAL mode when there are no simultaneous connections with write access. from https://www.sqlite.org/wal.html:
> Usually, the WAL file is deleted automatically when the last connection to the database closes. However, if the last process to have the database open exits without cleanly shutting down the database connection, or if the [SQLITE_FCNTL_PERSIST_WAL](https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpersistwal) [file control](https://www.sqlite.org/c3ref/file_control.html) is used, then the WAL file might be retained on disk after all connections to the database have been closed. [...]
> 
> [...]
>
> On newer versions of SQLite, a WAL-mode database on read-only media, or a WAL-mode database that lacks write permission, can still be read as long as one or more of the following conditions are met:
> - The -shm and -wal files already exist and are readable. 
> - [...]

gwenn:
Fix #1881
- [X] Check which ops are relevant in https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html
- [X] Tests all ops
- [X] exclude wasm-...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@
 //! intended.
 //!
 //! ```rust
-//! use rusqlite::{params, Connection, Result};
+//! use rusqlite::{Connection, Result, params};
 //!
 //! #[derive(Debug)]
 //! struct Person {
@@ -72,7 +72,7 @@ pub use sqlite_wasm_rs as ffi;
 
 use std::cell::RefCell;
 use std::default::Default;
-use std::ffi::{CStr, CString, c_char, c_int, c_uint};
+use std::ffi::{CStr, CString, c_char, c_int, c_uint, c_void};
 use std::fmt;
 
 use std::path::Path;
@@ -354,6 +354,8 @@ fn path_to_cstring(p: &Path) -> Result<CString> {
 
 /// Shorthand for `Main` database.
 pub const MAIN_DB: &CStr = c"main";
+/// Shorthand for default name.
+pub const DEFAULT_NAME: Option<&CStr> = None;
 /// Shorthand for `Temp` database.
 pub const TEMP_DB: &CStr = c"temp";
 
@@ -1110,8 +1112,44 @@ impl Connection {
 
     /// Set error code and message
     #[cfg(feature = "modern_sqlite")] // 3.51.0
-    pub fn set_errmsg(&self, code: c_int, msg: Option<&std::ffi::CStr>) -> Result<()> {
+    pub fn set_errmsg(&self, code: c_int, msg: Option<&CStr>) -> Result<()> {
         unsafe { error::set_errmsg(self.handle(), code, msg) }
+    }
+
+    /// Low-level control of database files
+    ///
+    /// See `https://sqlite.org/c3ref/file_control.html` for details.
+    pub fn file_control<N: Name>(&self, db_name: Option<N>, op_and_arg: FileControl) -> Result<()> {
+        let (op, arg) = match op_and_arg {
+            FileControl::SizeHint(p) => (ffi::SQLITE_FCNTL_SIZE_HINT, p as *mut _ as *mut c_void),
+            FileControl::SizeLimit(p) => (ffi::SQLITE_FCNTL_SIZE_LIMIT, p as *mut _ as *mut c_void),
+            FileControl::ChunkSize(p) => (ffi::SQLITE_FCNTL_CHUNK_SIZE, p as *mut _ as *mut c_void),
+            FileControl::PersistWal(p) => {
+                (ffi::SQLITE_FCNTL_PERSIST_WAL, p as *mut _ as *mut c_void)
+            }
+            FileControl::PowerSafeOverwrite(p) => (
+                ffi::SQLITE_FCNTL_POWERSAFE_OVERWRITE,
+                p as *mut _ as *mut c_void,
+            ),
+            FileControl::MMapSize(p) => (ffi::SQLITE_FCNTL_MMAP_SIZE, p as *mut _ as *mut c_void),
+            #[cfg(unix)]
+            FileControl::HasMoved(p) => (ffi::SQLITE_FCNTL_HAS_MOVED, p as *mut _ as *mut c_void),
+            FileControl::LockTimeout(p) => {
+                (ffi::SQLITE_FCNTL_LOCK_TIMEOUT, p as *mut _ as *mut c_void)
+            }
+            #[cfg(feature = "modern_sqlite")]
+            FileControl::BlockOnConnect(p) => (
+                ffi::SQLITE_FCNTL_BLOCK_ON_CONNECT,
+                p as *mut _ as *mut c_void,
+            ),
+            FileControl::DataVersion(p) => {
+                (ffi::SQLITE_FCNTL_DATA_VERSION, p as *mut _ as *mut c_void)
+            }
+            FileControl::ReserveBytes(p) => {
+                (ffi::SQLITE_FCNTL_RESERVE_BYTES, p as *mut _ as *mut c_void)
+            }
+        };
+        unsafe { self.db.borrow().file_control(db_name, op, arg) }
     }
 }
 
@@ -1161,8 +1199,8 @@ impl<'conn, 'sql> Batch<'conn, 'sql> {
     }
 }
 impl<'conn> fallible_iterator::FallibleIterator for Batch<'conn, '_> {
-    type Item = Statement<'conn>;
     type Error = Error;
+    type Item = Statement<'conn>;
 
     /// Iterates on each batch statements.
     ///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1110,9 +1111,28 @@ impl Connection {
...
+            FileControl::ChunkSize(p) => (ffi::SQLITE_FCNTL_CHUNK_SIZE, p.as_ptr().cast()),
+            FileControl::PersistWal(p) => (ffi::SQLITE_FCNTL_PERSIST_WAL, p.as_ptr().cast()),
+            FileControl::PowerSafeOverwrite(p) => {
+                (ffi::SQLITE_FCNTL_POWERSAFE_OVERWRITE, p.as_ptr().cast())
+            }
+            FileControl::HasMoved(p) => (ffi::SQLITE_FCNTL_HAS_MOVED, p.as_ptr().cast()),
+            FileControl::LockTimeout(p) => (ffi::SQLITE_FCNTL_LOCK_TIMEOUT, p.as_ptr().cast()),
+            FileControl::DataVersion(p) => (ffi::SQLITE_FCNTL_DATA_VERSION, p.as_ptr().cast()),
+        };
+        unsafe { self.db.borrow().file_control(db_name, op, arg) }
```
z-erica, line 1152: this is unsound; it's possible to crash the process doing

```rust
db.file_control(
    None::<&str>,
    sqlite3::FileControl::PersistWal(NonNull::dangling()),
)
.unwrap();

```

z-erica:
it works! but it's clunky in a couple of ways
```rust
db.file_control(
    None::<&str>,
    sqlite3::FileControl::PersistWal(NonNull::from_mut(&mut 1)),
)
.unwrap();
```
- first of all, the `::<&str>` turbofish, which is required when `None` is passed as the database name argument, is confusing. the signature feels overly generic
- more importantly, the requirement to pass a `NonNull<i32>` is strange; even if it matches the c api exactly, i believe it's more correct to expose this specific file control operation as a boolean flag

src/lib.rs
```diff
@@ -1444,18 +1491,16 @@ mod test {
...
+        assert_matches!(
+            Connection::open_with_flags(filename, OpenFlags::SQLITE_OPEN_READ_ONLY),
+            Err(Error::SqliteFailure(
+                ffi::Error {
+                    code: ErrorCode::CannotOpen,
+                    extended_code: ffi::SQLITE_CANTOPEN,
+                },
+                Some(msg)
+            )) if msg.contains(filename)
+        );
```
z-erica, line 1529: should these changes be part of this commit? they look unrelated

gwenn: I have fixed all the places where we do assertions on error to be like https://github.com/rusqlite/rusqlite/pull/1882/changes#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R2432.

src/lib.rs
gwenn: I don't know how to prevent that (except by making the function `unsafe`.

gwenn:
> * first of all, the `::<&str>` turbofish, which is required when `None` is passed as the database name argument, is confusing. the signature feels overly generic

This is already used pervasively:
https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.load_extension
https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.transaction_state
https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.column_exists
https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.column_metadata

But we can introduce:
```diff
/// Shorthand for `Main` database.
pub const MAIN_DB: &CStr = c"main";
+/// Shorthand for default database.
+pub const DEFAULT_DB: Option<&CStr> = None;
/// Shorthand for `Temp` database.
pub const TEMP_DB: &CStr = c"temp";
```

src/lib.rs
nekopsykose: `PersistWal(NonNull<c_int>)` could be `PersistWal(bool)` instead, and then casted to 1/0 and passed as a pointer in `file_control`.

gwenn:
> * more importantly, the requirement to pass a `NonNull<i32>` is strange; even if it matches the c api exactly, i believe it's more correct to expose this specific file control operation as a boolean flag

But SQLite requires an integer:
https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpersistwal
> The fourth parameter to [sqlite3_file_control()](https://sqlite.org/c3ref/file_control.html) for this opcode should be a pointer to an integer.

src/lib.rs
nekopsykose: ah, yeah according to the docs -1 gives you a value back so that wouldn't work

z-erica:
> But SQLite requires an integer: https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpersistwal
> 
> > The fourth parameter to [sqlite3_file_control()](https://sqlite.org/c3ref/file_control.html) for this opcode should be a pointer to an integer.

it's possible for that parameter to safely be a pointer to the stack, for example (code is untested):
```rust
    pub fn file_control<N: Name>(&self, db_name: Option<N>, op_and_arg: FileControl) -> Result<()> {
        match op_and_arg {
            FileControl::SetPersistWal(flag) => {
                let value: i32 = flag as i32; // will be 0 for false and 1 for true
                unsafe { self.db.borrow().file_control(db_name, ffi::SQLITE_FCNTL_PERSIST_WAL, (&raw const value).cast()) };
            }
            _ => todo!("irrelevant to this discussion"),
        }
    }
```
the remaining case is -1, for querying the current value of the flag, which would require an entirely different signature (since the pointer becomes an out parameter)

overall i don't think it's practical to make a single method for all possible (supported) file control operations like this

z-erica:
> overall i don't think it's practical to make a single method for all possible (supported) file control operations like this

hence, i suggest making a unique method for each. say (again, untested):
```rust
    pub fn get_persist_wal<N: Name>(&self, db_name: Option<N>) -> Result<bool> {
        let mut value: i32 = -1;
        unsafe { self.db.borrow().file_control(db_name, ffi::SQLITE_FCNTL_PERSIST_WAL, (&raw mut value).cast()) };
        [...]
        Ok(value != 0)
    }

    pub fn set_persist_wal<N: Name>(&self, db_name: Option<N>, persist_wal: bool) -> Result<()> {
        let value: i32 = flag as i32; // will be 0 for false and 1 for true
        unsafe { self.db.borrow().file_control(db_name, ffi::SQLITE_FCNTL_PERSIST_WAL, (&raw const value).cast()) };
        [...]
    }
```

gwenn:
This doesn't scale well.
For example,
https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntllocktimeout
> The parameter is a pointer to a 32-bit signed integer that contains the value that M is to be set to. Before returning, the 32-bit signed integer is overwritten with the previous value of M.

So in this case, the pointer is both used to set and retrieve the previous value...

z-erica:
> This doesn't scale well. For example, https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntllocktimeout
> 
> > The parameter is a pointer to a 32-bit signed integer that contains the value that M is to be set to. Before returning, the 32-bit signed integer is overwritten with the previous value of M.
> 
> So in this case, the pointer is both used to set and retrieve the previous value...

sure! but that's another file control operation entirely. as i said, i think it would warrant getting its own method, separate from `get_`/`set_persist_wal`

gwenn:
> > This doesn't scale well. For example, https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntllocktimeout
> > > The parameter is a pointer to a 32-bit signed integer that contains the value that M is to be set to. Before returning, the 32-bit signed integer is overwritten with the previous value of M.
> > 
> > 
> > So in this case, the pointer is both used to set and retrieve the previous value...
> 
> sure! but that's another file control operation entirely. as i said, i think it would warrant getting its own method, separate from `get_`/`set_persist_wal`

https://sqlite.org/cli.html
> .filectrl CMD ...        Run various sqlite3_file_control() operations

```sh
sqlite> .filectrl
Available file-controls:
  .filectrl chunk_size SIZE
  .filectrl data_version
  .filectrl has_moved
  .filectrl lock_timeout MILLISEC
  .filectrl persist_wal [BOOLEAN]
  .filectrl psow [BOOLEAN]
  .filectrl reserve_bytes [N]
  .filectrl size_limit [LIMIT]
  .filectrl tempfilename
```
=> single command / method

src/lib.rs
gwenn: Fixed by https://github.com/rusqlite/rusqlite/compare/fe8adb7c0e79c98dddc014acad99b16c9896ff8e..9b502fed7fc7299410a84c8a623e0f456ef25efb

src/lib.rs
```diff
@@ -1110,9 +1112,45 @@ impl Connection {
...
     #[cfg(feature = "modern_sqlite")] // 3.51.0
-    pub fn set_errmsg(&self, code: c_int, msg: Option<&std::ffi::CStr>) -> Result<()> {
+    pub fn set_errmsg(&self, code: c_int, msg: Option<&CStr>) -> Result<()> {
         unsafe { error::set_errmsg(self.handle(), code, msg) }
     }
+
+    /// Low-level control of database files
+    ///
+    /// See `https://sqlite.org/c3ref/file_control.html` for details.
+    pub fn file_control<N: Name>(&self, db_name: Option<N>, op_and_arg: FileControl) -> Result<()> {
```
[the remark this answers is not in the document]
gwenn: See https://doc.rust-lang.org/reference/lifetime-elision.html

PR #1886, opened 2026-08-08
Clarify MSRV policy

### LINKED ISSUE #566: Consider supporting an older rustc? ("MSRV")

### LINKED ISSUE #1625: Accidental MSRV of 1.82
#1620 and #1613 before that, kind of accidentally bumped the MSRV to 1.82 because of the use of `unsafe extern "C"` blocks in newer versions of bindgen.
rusqlite should have an explicit MSRV and/or set rust_target() on the bindgen builder.

### LINKED ISSUE #1631: could not compile `libsqlite3-sys` (lib) due to 242 previous errors [rusqlite 0.33.0]
After change `rusqlite` version to latest `0.33.0` in `Cargo.toml` I have following errors on application build:

```
error: extern block cannot be declared unsafe
    --> /target/debug/build/libsqlite3-sys-dab7a37efd566e68/out/bindgen.rs:2559:1
     |
2559 | unsafe extern "C" {
     | ^^^^^^
     |
     = note: see issue #123743 <https://github.com/rust-lang/rust/issues/123743> for more information

error: could not compile `libsqlite3-sys` (lib) due to 242 previous errors
warning: build failed, waiting for other jobs to finish...
```

With `0.32.1` everything works. Is that error related to deprecated system dependencies?
It's Fedora 41

### LINKED ISSUE #1880: MSRV should be 1.95 and in Cargo.toml
Compiling rusqlite requires `cfg_select`, which was stabilised in 1.95. This should be added to the `Cargo.toml` `rust-version`. Currently the minimum supported rust version is hinted at in the README.md, which may not be seen by downstream users (especially if it's a transitive dependency).

I have a downstream package which can optionally depend on rusqlite or turso. If rusqlite specified its own MSRV, I and my users can rely on that to avoid unexpected build failures; as it is, I need to set my MSRV to 1.95 even though only the rusqlite backend imposes that constraint.

Also noted in https://github.com/rusqlite/rusqlite/issues/566#issuecomment-2245391186 , and would have avoided https://github.com/rusqlite/rusqlite/issues/1625

xtqqczze:
Closes https://github.com/rusqlite/rusqlite/issues/566
Closes https://github.com/rusqlite/rusqlite/issues/1625
Closes https://github.com/rusqlite/rusqlite/issues/1631
Closes https://github.com/rusqlite/rusqlite/issues/1880

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,8 +3,8 @@ name = "rusqlite"
 # Note: Update version in README.md when you change this.
 version = "0.40.1"
 authors = ["The rusqlite developers"]
-edition = "2024"
-rust-version = "1.88.0"
+edition.workspace = true
+rust-version.workspace = true
 description = "Ergonomic wrapper for SQLite"
 repository = "https://github.com/rusqlite/rusqlite"
 documentation = "https://docs.rs/rusqlite/"
@@ -33,6 +33,10 @@ name = "rusqlite"
 
 [workspace]
 members = ["libsqlite3-sys"]
+
+[workspace.package]
+edition = "2024"
+rust-version = "1.88.0"
 
 [features]
 # if not SQLITE_OMIT_LOAD_EXTENSION
--- a/README.md
+++ b/README.md
@@ -257,6 +257,10 @@ Depending on the set of enabled cargo `features`, rusqlite and libsqlite3-sys wi
 
 Both of these are quite permissive, have no bearing on the license of the code in `rusqlite` or `libsqlite3-sys` themselves, and can be entirely ignored if you do not use the feature in question.
 
-## Minimum supported Rust version (MSRV)
+## Minimum Supported Rust Version (MSRV)
 
-Latest stable Rust version at the time of release. It might compile with older versions.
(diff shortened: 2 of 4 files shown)


gwenn:
Check MSRV:
https://github.com/rusqlite/rusqlite/pull/1890/changes#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3

PR #1892, opened 2026-08-21
fix: changed `restore` fn `progress` arg to a concrete type

Darckfast:
Before, in order to use the `restore` fn with a `None` `progress` argument, a concrete type should be specified with turbofish. e.g.

```rust

conn.restore::<&CStr, PathBuf, fn(backup::Progress)>(MAIN_DB, path, None).unwrap();

```

This changes the `progress` argument to be a concrete type, just like in the `backup` fn, making less verbose to call `restore`

```rust

conn.restore(MAIN_DB, path, None).unwrap();

```

### DIFF
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -99,11 +99,11 @@ impl Connection {
     ///
     /// Will return `Err` if the destination path cannot be opened
     /// or if the restore fails.
-    pub fn restore<N: Name, P: AsRef<Path>, F: Fn(Progress)>(
+    pub fn restore<N: Name, P: AsRef<Path>>(
         &mut self,
         name: N,
         src_path: P,
-        progress: Option<F>,
+        progress: Option<fn(Progress)>,
     ) -> Result<()> {
         use self::StepResult::{Busy, Done, Locked, More};
         let src = Self::open(src_path)?;


gwenn:
I am pretty sure that `fn` and `Fn` are not the same thing in Rust.

Darckfast:
> I am pretty sure that `fn` and `Fn` are not the same thing in Rust.

yes, they are different

`fn` is function pointer, and `Fn` is a trait that describe a calling function, if implemented it allow us to call it using the standard function syntax (e.g. `f()`). [A more detailed explanation](https://users.rust-lang.org/t/difference-between-fn-and-box-dyn-fn/39493)

given the scope of the `restore` fn, i made so it used the same arguments as the `backup` fn

https://github.com/rusqlite/rusqlite/blob/bc267609a4547e0c7bc156f424662ebaf05a06a7/src/backup.rs#L64-L69

but i can change it to a object trait instead, so the `restore` function keeps using the `Fn` trait

gwenn:
Indeed, there is a discrepancy between `backup` and `restore`.
But I think we should fix `backup` instead.
Because with `Fn`, you can still pass an `fn` but not the opposite way.
See https://github.com/rusqlite/rusqlite/pull/1685 for a semi related PR.

gwenn:
Could you please have look to #1893 ?

PR #1899, opened 2026-09-03
use-after-free from safe code in trace_v2 / busy_handler / ...

gwenn:
Fix issue reported by @marius-momeu :
> While auditing rusqlite for memory-safety issues reachable from safe code, we
found a use-after-free introduced on `master` by PR #1791
("sqlite3_set_clientdata / sqlite3_get_clientdata"), commits 36f3b84 (trace) and
9900bd6 (busy). Released 0.40.1 is not affected: there those callbacks were
still plain `fn` pointers. We found more similar issues, this is just the first one
that we're reporting. Let me know if you'd like them reported the same way.
>
> Issue
> -----
> `Connection::trace_v2` (src/trace.rs:211) and `Connection::busy_handler`
(src/busy.rs:75) box the user closure and store it with
`sqlite3_set_clientdata`. That C function unconditionally runs the previous
entry's destructor (sqlite3.c, `if( p->xDestructor ) p->xDestructor(p->pData);`)
— it has no "in use" guard, unlike `sqlite3_create_function*` /
`sqlite3_create_collation*`, which return SQLITE_BUSY while `db->nVdbeActive`.
> 
> Both callbacks fire inside `sqlite3_step()`, where rusqlite holds no RefCell
borrow on the connection. So if the callback re-registers itself — the natural
"trace N events then stop", or "retry twice then uninstall the busy handler" —
rusqlite drops the `Box<F>` while `trace_callback::<F>` is still executing
`(*trace_fn)(...)` inside it. Everything the closure captured is freed and the
`&mut F` goes dangling. No unsafe, no panic, API returns Ok(()).
```rust
    conn.trace_v2(TraceEventCodes::SQLITE_TRACE_ROW, Some(move |_e| {
        budget[0] -= 1;                       // Vec<u8> captured by the closure
        if budget[0] == 0 {
            conn.trace_v2(TraceEventCodes::empty(), None::<fn(TraceEvent<'_>)>).unwrap();
        }
        println!("{}", budget[0]);            // <-- read of freed memory
    }))?;
```
> 
> AddressSanitizer (nightly, `-Zsanitizer=address`, bundled SQLite 3.53.4):
```
    ERROR: AddressSanitizer: heap-use-after-free ... READ of size 8
      #7  <closure> src/main.rs:39
      #8  rusqlite::Connection::trace_v2::trace_callback  src/trace.rs:193
      #14 sqlite3VdbeExec / sqlite3_step
    freed by thread T0 here:
      #8  rusqlite::util::free_boxed_value  src/util/mod.rs:12
      #9  sqlite3_set_clientdata            sqlite3.c:191298
      #11 rusqlite::Connection::trace_v2    src/trace.rs:211
```
> 
> Without a sanitizer it segfaults. An equivalent reproducer exists for
`busy_handler` (also `busy_timeout`, src/busy.rs:86).
>
> Scope
> -----
> The same shape exists for every `set_clientdata`-backed callback:
`hooks/mod.rs` (commit/rollback/update/wal hooks, progress_handler, authorizer)
and `hooks/preupdate_hook.rs`. We reproduced only trace and busy.
`create_scalar_function` / `create_collation` are safe — SQLite's own
SQLITE_BUSY guard blocks the re-entrant replacement.
> 
> Fix direction: do not let SQLite free the old box eagerly (keep it owned by the
connection / defer the drop), or reject re-entrant registration.
>
> Reproducers available on request. We are only disclosing; no patch is attached.
> Found by [Crustify](https://github.com/crustify-rs/crustify-audit), an
experimental UB/soundness auditing agent developed at UC Berkeley and running on
claude-opus-5, then manually reviewed and independently reproduced.

### DIFF
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -337,13 +337,25 @@ impl InnerConnection {
         crate::error::check(unsafe { ffi::sqlite3_file_control(self.db, cn, op, arg) })
     }
 
-    pub fn set_clientdata<T: Send + 'static, N: Name>(
+    pub fn set_clientdata<
+        T: Send + 'static,
+        N: Name,
+        F: Fn(*mut ffi::sqlite3, *mut c_void) -> c_int,
+    >(
         &mut self,
         name: N,
         data: Option<T>,
+        preset: F,
     ) -> Result<*mut T> {
         let name = name.as_cstr()?;
         let ptr = data.map_or(ptr::null_mut(), |d| Box::into_raw(Box::new(d)));
+        let res = self.decode_result(preset(self.db, ptr.cast()));
+        if res.is_err() {
+            if !ptr.is_null() {
+                unsafe { crate::util::free_boxed_value::<T>(ptr.cast()) };
+            }
+            res?;
+        }
         self.decode_result(unsafe {
             ffi::sqlite3_set_clientdata(
                 self.db,
--- a/src/busy.rs
+++ b/src/busy.rs
@@ -52,7 +52,7 @@ impl Connection {
     /// Newly created connections default to a
     /// [`busy_timeout()`](Connection::busy_timeout) handler with a timeout
     /// of 5000ms, although this is subject to change.
-    pub fn busy_handler<F>(&self, callback: Option<F>) -> Result<()>
+    pub fn busy_handler<F>(&mut self, callback: Option<F>) -> Result<()>
     where
         F: FnMut(i32) -> bool + Send + 'static,
     {
@@ -72,8 +72,10 @@ impl Connection {
         }
         let x = callback.as_ref().map(|_| busy_handler_callback::<F> as _);
         let mut c = self.db.borrow_mut();
-        let bh = c.set_clientdata(c"sqlite3_busy_handler", callback)?;
-        c.decode_result(unsafe { ffi::sqlite3_busy_handler(c.db(), x, bh.cast()) })
+        c.set_clientdata(c"sqlite3_busy_handler", callback, |db, bh| unsafe {
+            ffi::sqlite3_busy_handler(db, x, bh)
+        })?;
+        Ok(())
     }
 }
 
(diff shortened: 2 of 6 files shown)


gwenn:
- [ ] Add `compile_fail` examples

thomcc:
It's a little unfortunate to require `&mut` here, but honestly it seems like the best approach. Requiring `&mut` access to configure hooks and other things isn't so bad (you can still make queries with just `&Connection`, after all).

gwenn:
> It's a little unfortunate to require `&mut` here, but honestly it seems like the best approach. Requiring `&mut` access to configure hooks and other things isn't so bad (you can still make queries with just `&Connection`, after all).

For busy handler, It seems to prevent another bug:
https://sqlite.org/c3ref/busy_handler.html
> The busy callback should not take any actions which modify the database connection that invoked the busy handler. In other words, the busy handler is not **reentrant**. Any such actions result in undefined behavior.

Idem for commit / rollback hooks:
https://sqlite.org/c3ref/commit_hook.html
> The commit and rollback hook callbacks are not reentrant. The callback implementation must not do anything that will modify the database connection that invoked the callback. Any actions to modify the database connection must be deferred until after the completion of the [sqlite3_step()](https://sqlite.org/c3ref/step.html) call that triggered the commit or rollback hook in the first place. Note that running any other SQL statements, including SELECT statements, or merely calling [sqlite3_prepare_v2()](https://sqlite.org/c3ref/prepare.html) and [sqlite3_step()](https://sqlite.org/c3ref/step.html) will modify the database connections for the meaning of "modify" in this paragraph.

But I am afraid that you can still invoke `sqlite3_step`.

src/inner_connection.rs
```diff
@@ -337,13 +337,23 @@ impl InnerConnection {
...
+    >(
         &mut self,
         name: N,
         data: Option<T>,
+        preset: F,
     ) -> Result<*mut T> {
         let name = name.as_cstr()?;
         let ptr = data.map_or(ptr::null_mut(), |d| Box::into_raw(Box::new(d)));
+        let res = self.decode_result(preset(self.db, ptr.cast()));
+        if res.is_err() && !ptr.is_null() {
```
gwenn: FIXME, should be:
```rust
if res.is_err() {
  if !ptr.is_null() {
    unsafe { crate::util::free_boxed_value::<T>(ptr.cast()) };
  }
  res?;
}
```

PR #1900, opened 2026-09-03
Don't enable SQLITE_ENABLE_MEMORY_MANAGEMENT in the bundled build

lthiery:
Updated to take @gwenn's suggestion: no feature flag, just drop the define from the bundled build.

The flag has been hardcoded since the flag list was first put together in 2016 (#197), where it was copied over from other SQLite bindings. It has a real runtime cost (memory accounting overhead, and it forces the unified page cache), and since #1608 nothing in rusqlite itself needs it: `Connection::release_memory` uses `sqlite3_db_release_memory`, which works either way, and `sqlite3_release_memory` is only reachable through raw ffi.

Anyone who still wants it can pass `LIBSQLITE3_FLAGS="-DSQLITE_ENABLE_MEMORY_MANAGEMENT"`.

### DIFF
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -163,7 +163,6 @@ mod build_bundled {
             .flag("-DSQLITE_ENABLE_FTS5")
             .flag("-DSQLITE_ENABLE_JSON1")
             .flag("-DSQLITE_ENABLE_LOAD_EXTENSION=1")
-            .flag("-DSQLITE_ENABLE_MEMORY_MANAGEMENT")
             .flag("-DSQLITE_ENABLE_RTREE")
             .flag("-DSQLITE_ENABLE_STAT4")
             .flag("-DSQLITE_SOUNDEX")


lthiery:
Oops I guess you can do LIBSQLITE3_FLAGS="-USQLITE_ENABLE_MEMORY_MANAGEMENT" to get rid of it. Leaving this up in case you want it but feel free to close - I don't need this necessarily but I do think it's more ergonomic than the env var FWIW

gwenn:
There are too many SQLite compile-time options.
We cannot have a feature flag for each of them.
But I am ok to just remove the default activation:
```diff

            .flag("-DSQLITE_ENABLE_JSON1")
            .flag("-DSQLITE_ENABLE_LOAD_EXTENSION=1")
-            .flag("-DSQLITE_ENABLE_MEMORY_MANAGEMENT")
            .flag("-DSQLITE_ENABLE_RTREE")
            .flag("-DSQLITE_ENABLE_STAT4")
            .flag("-DSQLITE_SOUNDEX")
```

lthiery:
That works for me!

PR #1907, opened 2026-09-22
feat: add support for set_last_insert_rowid

### LINKED ISSUE #1757: Please expose sqlite3_set_last_insert_rowid
Hi.  Thanks for this library.

In my application, I would like to explicitly set this to zero so that I can more reliably detect certain programming errors.

There's a function for this in the C library but AFAICT it isn't exposed in libsqlite3-sys nor in rusqlite.

If you like this idea, I can prepare an MR.

anatawa12:
Fixes #1757

In the thread of #1757, #1770 was mentioned for set_last_insert_rowid, but the PR was not updated for several months, and it adds `set_last_insert_rowid` for vtab.

This PR instead adds `set_last_insert_rowid` to inner_connection and connection. I did not touch vtab part because as far as I searched `last_insert_rowid` was not used in vtab module. I'm sorry if I missed something.


anatawa12:
~~why is this marked as invalid?~~

EDIT: I looked for contributing.md but not look for contributing section in README. I fixed lint error

gwenn:
PR is marked as invalid because AFAIK `sqlite3_set_last_insert_rowid` can only be used by virtual table (if you search its usages in SQLite source code, only FTS5 is using it).
Which was implicitly stated here:
https://github.com/rusqlite/rusqlite/issues/1757#issuecomment-3579477684
> I am fine if it is exposed only on [VTabConnection](https://docs.rs/rusqlite/latest/rusqlite/vtab/struct.VTabConnection.html).

anatawa12:
I don't think `sqlite3_set_last_insert_rowid()` is restricted to virtual tables implementation.

The API documentation explicitly says that "allows the application to set the value returned by calling sqlite3_last_insert_rowid(D) to R", without any restriction. The `sqlite3_last_insert_rowid()` documentation also says that its value may be set explicitly by `sqlite3_set_last_insert_rowid()`.

https://sqlite.org/c3ref/set_last_insert_rowid.html
https://sqlite.org/c3ref/last_insert_rowid.html

I also don't see any restriction in the implementation of `sqlite3_set_last_insert_rowid()` that would limit it to virtual table implementations.



I understand that SQLite itself primarily uses this API from virtual table implementations, but I don't see that as a restriction on applications using the public API.



Could you clarify what makes you think `sqlite3_last_insert_rowid()` invalid for non-virtual tables implementation?

gwenn:
SQLite is well known for its test coverage.
So if only FTS5 uses `sqlite3_set_last_insert_rowid`, I assume that only virtual table can use it.

anatawa12:
I asked about application use of `sqlite3_set_last_insert_rowid` on the SQLite forum, and Dan Kennedy, one of the SQLite developers, confirmed that there is no reason application use of `sqlite3_set_last_insert_rowid` would not be supported (at least for my use case).
He also mentioned that there are tests for it in SQLite's non-public test suite.

SQLite's commonly cited high test coverage includes the proprietary test suite,
so the lack of a corresponding test in the open-source test suite might not be strong evidence that this is an unsupported use case.

For reference: https://sqlite.org/forum/forumpost/d61ffb0aebbcfa8278592f866594041db0a59b98871aedfdc5f35a6246e2644e

gwenn:
But for such an unusual case, you can already use: [conn.handle()](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.handle) and something like:
```rust
use rusqlite::ffi;

unsafe { ffi::sqlite3_set_last_insert_rowid(db, ...)  }
```

anatawa12:
But now that I've confirmed that `sqlite3_set_last_insert_rowid` is supported for application use, I don't think there is a reason to require users to use `unsafe` for this.
Since the underlying SQLite API supports this use case, I think it would be reasonable for rusqlite to provide a safe wrapper for it.

The implementation of `sqlite3_set_last_insert_rowid` is also quite simple just like `sqlite3_last_insert_rowid`.
Is there any concern about providing a safe wrapper for this API, such as that `sqlite3_set_last_insert_rowid()` could become more complex or introduce safety concerns for other use cases in the future?

gwenn:
> My use case is to distinguish the INSERT and UPDATE paths of an UPSERT.

Maybe you should use https://sqlite.org/syntax/returning-clause.html instead ?

gwenn:
Maybe something like https://github.com/rusqlite/rusqlite/discussions/1349

anatawa12:
I don't think `RETURNING` alone can distinguish the two paths, since it will be executed for both the `INSERT` and `UPDATE` paths.

For example, `RETURNING rowid` would return a row in either case. We don't care rowid of the updated row, we just want to know whether the row was inserted or updated.
Comparing `RETURNING rowid` with `sqlite3_last_insert_rowid()` may produce false positives when the updated rowid matches with the `sqlite3_last_insert_rowid()` before the UPSERT.

I googled and I found one way to distinguish the two paths with `RETURNING` like `INSERT (is_insert, ...) VALUES (1, ...) ON CONFLICT (...) DO UPDATE SET is_insert = 0 RETURNING is_insert`, but this requires additional schema change, which is not desirable.

This is why I use `sqlite3_set_last_insert_rowid()` as a way to distinguish the paths without adding anything to the schema.

As far as I understand, #1349 does not need to distinguish between the INSERT and UPDATE paths, so it is not related to my intention of this PR.

